PR #83, opened 2022-10-27
I18N: Read translation files from game archive

WatchTheFort:


Jazcash:
Rejecting, as per https://discord.com/channels/549281623154229250/944691667955482624/1035242376517660872

tl;dr: would prefer to fetch i18n files as part of a compilation script, and have them bundled into the lobby, rather than fetching them at runtime

PR #85, opened 2022-11-09
Map library view, map detail view, map selector modal in battle view

thehobojoe:
### Work done:

- Map grid in library/map page
- Search by name, sort by name/size
- Map details upon clicking
- Cleanup map preview (less redraw)
- Map picker modal in battle view (also with search/sort)
- Sort simple map picker dropdown by map name

### Known issues
- Map sort dropdown goes out of place if search bar is focused/unfocused while its open
- Map search field is focused when opening map modal, but not when opening map grid in library
- Start box rendering is a little flaky. I'm not sure how much of this is A) hardcoded boxes for local hosting B) Failed server response for online hosting or C) bugs in my rendering/update code
- Map details view is obviously unfinished. I intend for it to be a placeholder until I can figure out implementation for 3D view etc and for certain map API's to become available. I believe the current state is acceptable, since current capability here is a lot more than chobby (besides the extra hardcoded info chobby has available)

### Notes

Please be very picky in the review! This is my first PR to this codebase and my first time working with vue, so I would like any mistakes to get corrected and to fit within the overall code style as much as I reasonably can.


src/api/session.ts
```diff
@@ -22,6 +23,7 @@ export class SessionAPI {
     public readonly outgoingFriendRequests: ComputedRef<User[]>;
     public readonly incomingFriendRequests: ComputedRef<User[]>;
     public readonly friends: ComputedRef<User[]>;
+    public readonly libraryPage: Ref<MapData | null> = shallowRef(null);
```
thehobojoe, line 26: I was unsure the best way to implement this - this is meant to be for navigation data for loading up specific library pages. I named it this way with the intent of it being usable for the eventual unit pages, but there is probably a better implementation.

src/components/battle/BattleComponent.vue
```diff
@@ -18,7 +18,7 @@
             <BattleChat v-if="!isOfflineBattle" :battle="battle" />
         </div>
         <div class="right-col flex-col gap-md">
-            <MapPreview
+            <BattleMapPreview
```
thehobojoe, line 21: I renamed this component because I was initially planning to make some kind of `SimpleMapPreview` component that had less canvas/startpos behavior. ATM I'm not sure if that's really necessary, so I'm happy to revert this name change.

src/components/maps/MapListComponent.vue
```diff
@@ -0,0 +1,149 @@
...
+        case SortMethod.Size:
+            maps.sort((a, b) => {
+                return a.width * a.height - b.width * b.height;
+            })
+    }
+    return maps;
+});
+
+function mapSelected(map: MapData) {
+    emit("mapSelected", map);
```
thehobojoe, line 75: This emit name gets reused a lot. Should I make some kind of top-level event symbol instead of this hardcoded name in a few places?

src/components/maps/MapOverviewCard.vue
```diff
@@ -0,0 +1,138 @@
...
+          <div class="thumbnail">
+            <BattleMapPreview
+                :map="map.scriptName"
+            />
+          </div>
+          <div class="details">
+            <div class="detail-text"><b>Size:</b> {{map.width}} x {{map.height}}</div>
+            <div class="detail-text"><b>Wind:</b> {{map.minWind}} - {{map.maxWind}}</div>
+            <div class="detail-text"><b>Description:</b> {{map.description}}</div>
+            <div class="detail-text"><b>Author:</b> {{map.mapInfo.author}}</div>
```
thehobojoe, line 22: Some minor duplication between this and the map details page. Not sure if it's worth making a dedicated component.

src/api/session.ts
Jazcash: The best way to do this is by passing query/data to the route, i.e. https://router.vuejs.org/guide/essentials/passing-props.html#object-mode

_However_, this currently isn't compatible with the automatically generated routes by the `vue-auto-routing` package that this project is currently using, so actually the way you've done it here is actually the only way I can think to do it, but it is indeed quite hacky. I'm currently in the process of switching the project over to Vite and this newer auto routing package: https://github.com/posva/unplugin-vue-router, which will allow us to use route params properly. I'll remember to fix this once that's done

src/components/battle/BattleComponent.vue
Jazcash: I think I'd rather keep the component flexible for now, and as you've used it in non-battle pages, I'll revert the name

src/components/maps/MapOverviewCard.vue
```diff
@@ -0,0 +1,125 @@
...
```
Jazcash, line 22: This breaks rendering when the user has old maps installed that don't have `mapInfo.lua` files (which means `map.mapInfo` here is null)

src/components/misc/Error.vue
```diff
@@ -21,14 +21,20 @@ const promiseError: Ref<PromiseRejectionEvent | null> = ref(null);
...
 });
 
 window.addEventListener("error", (event) => {
-    console.error(event);
-    error.value = event;
-    isVisible.value = true;
+  error.value = event;
+  if(event.message === "ResizeObserver loop limit exceeded") {
+    return;
+  }
```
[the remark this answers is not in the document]
Jazcash: I'm not sure about this, going to try fixing the root cause first

src/components/maps/MapListComponent.vue
```diff
@@ -0,0 +1,96 @@
...
+import { MapData } from "@/model/map-data";
+
+enum SortMethod {
+    Name = "Name",
+    Size = "Size",
+}
+
+const sortMethods = ref([SortMethod.Name, SortMethod.Size]);
+const sortMethod = ref(SortMethod.Name.valueOf());
+const searchVal = ref("");
```
Jazcash, line 40: I'm leaning more towards banning enums entirely for this project in eslint, for reasons https://www.youtube.com/watch?v=0fTdCSH_QEU

Jazcash:
I think it might be better to utilise [PrimeVue's DataView component](https://primefaces.org/primevue/dataview) for the map list

src/components/maps/MapListComponent.vue
Jazcash: Replaced with union type

src/components/battle/BattleComponent.vue
```diff
@@ -64,9 +64,10 @@
                     :placeholder="battle.battleOptions.map"
                     @update:model-value="onMapSelected"
                 />
-                <Button>
-                    <Icon :icon="cog" height="23" />
+                <Button @click="openMapList">
+                    <Icon :icon="listIcon" height="23" />
```
Jazcash, line 68: The previous cog button here was for the map options modal, although I don't think it was working which probably made it confusing. I'll leave this for now and add the cog button back in the future when we've got the map options modal working

src/components/maps/MapListComponent.vue
Jazcash: At the least, we can remove the `mapSelected` event on the `MapOverviewCard` element itself and instead add the `@click` handler in the parent component, done it here https://github.com/beyond-all-reason/bar-lobby/pull/85/commits/3a675db3c30d77a11a7d5779853d110286cbe368

src/components/battle/MapListModal.vue
```diff
@@ -0,0 +1,33 @@
...
+</template>
+
+<script lang="ts" setup>
+import { Ref, ref } from "vue";
+
+import Modal from "@/components/common/Modal.vue";
+import MapListComponent from "@/components/maps/MapListComponent.vue";
+import { MapData } from "@/model/map-data";
+
+const modal: Ref<null | InstanceType<typeof Modal>> = ref(null);
```
Jazcash, line 16: This is only if you need to access the modal element in code, which isn't happening here, so can be safely removed

Jazcash:
I think this is good enough to merge now. I can think of some nice improvements, and would ideally like to use PrimeVue's DataView component here like I said, but no reason we can't do that later. Thx for the PR 👍

PR #92, opened 2022-11-28
AI UI

thehobojoe:
### Overview
- Added non-engine AI support
- Broke up the playerlist component for better readability and easier customization
- Fixed and improved drag-and-drop between team behavior (shows invalid moves)
- Configuration on right-click will now only show up on bots that have configuration options (only engine AI's)

### Testing notes
Engine AI's will not work on engine version `105.1.1-1375-gd2541c7 BAR105`, I don't know why. I've reported it and it seems out of scope with this PR, so make sure you're on an older version to test.

This PR is offline only. Online support will require a lot of API knowledge that I do not have yet. Looking forward to learning though 🙂 

### Bot setup
This is copied almost 1:1 from chobby - I'm quite certain this is not the best way to do things (a bunch of hardcoded strings in a TS file) and it definitely does not have localization support in its current state. I'm not sure how important that is at this stage. In particular the merging of the engine AI's and the game AI's in `ai-content.ts` could use some validation. 

Another thing that I tried to do but ultimately abandoned was setting up some stronger types for the game AI's. Ultimately I gave up because it ended up being even uglier than what we have, and it was hard to marry with the engine AI's - if you have a good solution for this that would be great.

Lastly, the caching of engine bots seems broken. I'm not sure how this was supposed to work so I just forced it to reprocess. Maybe not ideal, but I don't imagine there's a performance problem here so it's probably ok.

### Opinionated changes
Breaking up the views is not required, but it made the changes I wanted easier and it should enable easier adjustment in the future (i.e. showing teams as columns instead of rows is a change I'd like to make in the future).

For handling the UI, I've designated specs with a `teamId` of `-1` for simplicity of handling them the same way as other teams so all the UI code can be shared. The scope of this is completely limited to `TeamComponent` and `Playerlist`, so it won't mess with anything else.

I set up prettier line-lengths for `.vue` files. This is just a readability change, I got tired of prettier putting all the view component parameters on the same line, it made things very annoying to read and adjust. I know you have a pre-existing setup that I'm sure you prefer so this can be rolled back if needed, but I really would like a way to keep those longer component declarations in vue html as multi-line if possible.

### Todo
Preventing scav/raptors from getting one of the pre-selected AI names. I know there are vague plans for a better UI for raptor/scav specifically so I did not bother implementing that in this PR. Once in-game, the game takes over and just calls them by the AI name, not the one assigned by the lobby, so it's not that big of a deal.

### DIFF
--- a/prettier.config.js
+++ b/prettier.config.js
@@ -1,3 +1,11 @@
 module.exports = {
     ...require("./node_modules/jaz-ts-utils/prettier.config"),
+    overrides: [
+        {
+            "files": "src/**/*.vue",
+            "options": {
+                "printWidth": 140
+            }
+        }
+    ]
 };
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ routes.js
 todo.txt
 /replays.db
 /test.json
+out
--- a/package-lock.json
+++ b/package-lock.json
@@ -5348,6 +5348,27 @@
             "strip-ansi": "^6.0.0"
           }
         },
+        "@vue/vue-loader-v15": {
+          "version": "npm:vue-loader@15.10.1",
+          "resolved": "https://registry.npmjs.org/vue-loader/-/vue-loader-15.10.1.tgz",
+          "integrity": "sha512-SaPHK1A01VrNthlix6h1hq4uJu7S/z0kdLUb6klubo738NeQoLbS6V9/d8Pv19tU0XdQKju3D1HSKuI8wJ5wMA==",
+          "dev": true,
+          "requires": {
+            "@vue/component-compiler-utils": "^3.1.0",
+            "hash-sum": "^1.0.2",
+            "loader-utils": "^1.1.0",
+            "vue-hot-reload-api": "^2.3.0",
+            "vue-style-loader": "^4.1.0"
+          },
+          "dependencies": {
+            "hash-sum": {
+              "version": "1.0.2",
+              "resolved": "https://registry.npmjs.org/hash-sum/-/hash-sum-1.0.2.tgz",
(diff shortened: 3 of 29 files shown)


thehobojoe:
If you're not a fan of some of the UI changes (particularly unifying teams/spectators into a shared component) or anything else, feel free to let me know and I'm happy to adjust things, don't feel like you have to make changes on my work like that yourself. 🙂

prettier.config.js
```diff
@@ -1,3 +1,11 @@
...
-};
\ No newline at end of file
+    overrides: [
+        {
+            "files": "src/**/*.vue",
+            "options": {
+                "printWidth": 140
+            }
+        }
+    ]
```
Jazcash, line 10: ha, I couldn't settle on a good printWidth to use when I first setup prettier. I messed around with different values but it always looked weird in some places and fine in others. I'm not really that fussed tbh, so happy to go with this

Jazcash:
Have skimmed through the code and tested a bit and it looks solid. Not been super thorough, but we can sort any lingering issues afterwards because we're still in a kind of alpha phase. In regards to some of your points:

The current method the lobby is using to grab the latest engine is simply grabbing the latest entry from https://github.com/beyond-all-reason/spring/releases. However, I believe these releases are only tested locally by the engine devs, and are effectively completely experimental for full production use with BAR every time. As discussed [here](https://discord.com/channels/549281623154229250/927564746104905728/1046747232063471647), it wouldn't even be enough for engine devs to mark releases as "experimental" or "stable", because it's on BAR people to make the decision of which engine versions are stable with the latest game releases. This is only really problematic for offline battles, because in online battles the lobby doesn't make the decision, it just has to grab the version used by the autohost - but for offline, we ideally want to always use the latest game release and latest engine release, so it's probable that we'll need to maintain hardcoded engine/game version strings in the lobby for the versions to use for offline content.

As for hardcoded AI definitions, totally fine for now. Eventually, it would be great if we can find a reliable way of grabbing all those definitions from the engine and game, along with all the localised descriptions, settings etc - but we can worry about that later.

Thanks for the PR!

Issue #160, opened 2023-03-20
Feature request: Search lobbies by title


Teifion:
I want to be able to type into a text box and filter down the list of lobbies by title and/or user membership.

Issue #158, opened 2023-03-20
Battle List Monotony/Filters


Descyphal:
The battle list could be more informative. Although clicking on a lobby to see info is great... The map preview should be downloaded when doing so, if possible. Maybe have icons and/or colour-coding for game/lobby states... It is currently bland and lacking identifiers so as to differentiate games at a glance. 

The filters reset after every action and need to be consistent. More over the filters aren't organizing properly with the categories, at times:

(sorted by players)

![image](https://user-images.githubusercontent.com/128336457/226235883-da8c7fdd-70cb-4062-b8ea-84ae07b8457f.png)

Issue #150, opened 2023-03-18
Lobby list filters don't stick


Teifion:
If I remove locked battles, next time I go to list the multiplayer lobbies it no longer filters out locked lobbies.

PR #172, opened 2023-03-21
Search battles by map and player names, save battle filter settings

thehobojoe:
Resolves https://github.com/beyond-all-reason/bar-lobby/issues/160
Resolves https://github.com/beyond-all-reason/bar-lobby/issues/150
Partially resolves https://github.com/beyond-all-reason/bar-lobby/issues/158

- Search now searches by battle title, map title, player name, spectator name
- Added filter by running (if the game is active)
- All filters now save with other lobby settings

There is no debounce on the search currently, the actual filter time is miniscule on my machine which should be a good indication that it won't be problematic on older machines - however re-rendering the DOM is another matter, I'll make a note to work on a search debounce later.

electron_SLfxIdCed3
electron_mjMpHhsL01

### DIFF
--- a/src/common/model/settings.ts
+++ b/src/common/model/settings.ts
@@ -8,4 +8,8 @@ export const settingsSchema = Type.Object({
     musicVolume: Type.Number({ default: 5, minimum: 0, maximum: 100 }),
     loginAutomatically: Type.Boolean({ default: true }),
     devMode: Type.Boolean({ default: false }),
+    battlesHideInProgress: Type.Boolean({ default: false }),
+    battlesHidePvE: Type.Boolean({ default: false }),
+    battlesHideLocked: Type.Boolean({ default: false }),
+    battlesHideEmpty: Type.Boolean({ default: true }),
 });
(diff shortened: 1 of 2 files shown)


Jazcash:
Have you seen https://primevue.org/datatable/#basic_filter?

thehobojoe:
> Have you seen https://primevue.org/datatable/#basic_filter?

I hadn't - how were you thinking of using it in this case, like having multiple search boxes? I feel like it should remain pretty simple to avoid making the UI more cluttered.

Jazcash:
> > Have you seen https://primevue.org/datatable/#basic_filter?
> 
> I hadn't - how were you thinking of using it in this case, like having multiple search boxes? I feel like it should remain pretty simple to avoid making the UI more cluttered.

I think 1 search box is fine, exactly as you placed it - but if possible I'd prefer if we used the built-in PrimeVue filter for behind the scenes, to keep our code simplier, and in case PrimeVue's filter is more optimised for their DataTable component. If it looks too awkward to hook up to then don't worry about it, but would be great if you could try it

Issue #142, opened 2023-03-08
Add friend by userId goes below 0


Jazcash:
![image](https://user-images.githubusercontent.com/1434248/223876918-2341b1ba-3a9d-48a4-9d46-5ee496b8c015.png)

PR #177, opened 2023-03-29
Switch to InputNumber for friendID for validation

IsaacWarren:
Addresses #142 
Tested that adding friends still works

https://user-images.githubusercontent.com/42949629/228409402-96addfbd-efa5-42a9-8e83-3a56fa55a9ff.mp4

### DIFF
--- a/src/renderer/components/navbar/Friends.vue
+++ b/src/renderer/components/navbar/Friends.vue
@@ -104,7 +104,7 @@ const emits = defineEmits<{
 const activeIndex = ref(0);
 const accordianActiveIndexes = ref([0, 1, 2]);
 const friendId = ref();
-let addFriendDisabled = ref(true);
+const addFriendDisabled = ref(true);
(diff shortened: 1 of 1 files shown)


src/renderer/components/navbar/Friends.vue
```diff
@@ -92,7 +103,8 @@ const emits = defineEmits<{
 
 const activeIndex = ref(0);
 const accordianActiveIndexes = ref([0, 1, 2]);
-const friendIdStr = ref("");
+const friendId = ref();
+let addFriendDisabled = ref(true);
```
Jazcash, line 107: try to use `const` when possible, especially for refs, as it's almost always a mistake to reassign refs, as their `.value` property is usually what people intend to change. helps future devs avoid mistakes

IsaacWarren: Makes sense, I'm new to Vue so appreciate the feedback, fixed.

Issue #157, opened 2023-03-20
New Lobby: !help and $help commands


Descyphal:
Very simply, the commands from !help and $help have always needed more intuitive and practical presentation/availability.

perhaps these commands should be listed here:

![commands](https://user-images.githubusercontent.com/128336457/226235149-11b72acb-85e2-4bda-9002-e35bd8b92032.png)

dusekj4:
will look to it

Jazcash:
fixed in https://github.com/beyond-all-reason/bar-lobby/pull/187

PR #184, opened 2023-04-05
New Lobby: !help and $help commands

dusekj4:
**Issue**
#157 

**Test**
1. go to >>Library >> Commands

### DIFF
--- a/src/renderer/views/library/commands.vue
+++ b/src/renderer/views/library/commands.vue
@@ -1,22 +1,143 @@
 <route lang="json5">
-{ meta: { title: "Commands", order: 4, transition: { name: "slide-left" } } }
+    { meta: { title: "Commands", order: 4, transition: { name: "slide-left" } } }
 </route>
 
+
 <template>
-    <div>
-        <h1>{{ route.meta.title }}</h1>
-        <Markdown
-            source="
-- Same as online commands page
-"
-        />
+  <div class="flex-col gap-lg flex-grow fullheight">
+    <h1>{{ route.meta.title }}</h1>
+    <div class="flex-row gap-md">
+      <SearchBox v-model="searchVal" />
+      <Select v-model="sortMethod" :options="sortMethods" label="Sort By" />
     </div>
+
+    <div class="flex-col flex-grow fullheight">
+      <div class="scroll-container">
+        <div v-for="command in filteredCommands" :key="command.cmd" class="command">
+          <div class="cmd">{{ command.cmd }}</div>
+          <div class="cmdDescription">{{ command.cmdDescription }}</div>
+        </div>
+      </div>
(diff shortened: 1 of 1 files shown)


Jazcash:
As for the general approach, I think this is the best we can do for now. Sending direct messages to bots and parsing the response obviously feels hacky, but there's not much else we can do about it unless server/SPADS provide a proper commend reference API. Parsing from their github docs is also an option, but I think I prefer this way for now because there's no guarantee their docs will be kept up to date.

I'm not sure the dropdown for filtering by name _or_ description is necessary, because I don't think sorting by description adds much value.

No need to worry about styling stuff too much, what you've done looks fine for now. I'll probably be iterating on style/design over time anyways.

I have some ideas for improvement, but they're more like additional feature requests so I'll open them as new issues after this is merged.

src/renderer/views/library/commands.vue
```diff
@@ -3,20 +3,126 @@
...
+    api.comms.request("c.communication.send_direct_message", {
+        recipient_id: 3137,
+        message: `$help`,
+    });
+});
+
+// Unsubscribe from the response listener when the component is not visible anymore
+onUnmounted(() => {
+    /**
+  I don't know how to unsubscribe from the response listener
```
Jazcash, line 94: `.onResponse` returns a `SignalBinding` instance that you can later call `.destroy()` on to solve this. See `HostBattle.vue` for an example

src/renderer/views/library/commands.vue
```diff
@@ -3,20 +3,126 @@
...
+
+        cmdDescription && !cmdDescription.includes("*") && commands.push({ cmd, cmdDescription });
+    });
+    // Send a message to the server to get all the commands
+    api.comms.request("c.communication.send_direct_message", {
+        recipient_id: 3137,
+        message: `!helpall`,
+    });
+    api.comms.request("c.communication.send_direct_message", {
+        recipient_id: 3137,
```
Jazcash, line 86: Does this work? I thought `$help` only worked in public chat or in private message with `Coordinator`?

Jazcash: Yeah that's what I meant. I know `!help` works but I would've been surprised if `$help` works. `!`are SPADS commands for the autohosts, and `$` are server commands.

Default SPADS commands: http://planetspads.free.fr/spads/doc/spadsDoc.html (https://github.com/Yaribz/SPADS/blob/master/doc/spadsDoc_Commands.html)
Server commands: https://github.com/beyond-all-reason/teiserver/blob/9fa121b0a7e011957f90aad153c77badca593e11/lib/teiserver/coordinator/coordinator_lib.ex

Basically just change this `recipient_id` for `$help` to `3082` and it should work.

Issue #163, opened 2023-03-21
Rank icon


Descyphal:
Ranks are nowhere to be found in the New Lobby. In the playerlist, on the playercard, etc.

PR #190, opened 2023-04-07
Rank Icon in Multiplayer Lobby

Baegll:
Issue #163

Worked with multiple users, and added in all the rank icons from Chobby, including moderator icons.


Baegll:
We can also delay this for adding the check for moderator users

Jazcash:
I've been looking at this some more, and I've figured how to fix the image for local and prod builds, but I'm still on the fence about the actual rank icon functionality itself.

The fundamental issue here is that this rank icon is now based on rating rather than play time, yet the server still sends it as part of the user response rather than the client response.

In my eyes, I think user and client responses should be the same object, but as it's implemented right now on the server, user responses hold more permanent information like `username` and client responses hold battle-specific information like `isReady`. Because this rank icon is now based on rating, and rating is different depending on the mode (duel, team, ffa), I think this rank icon data should be part of the client response, so it can update whenever the mode changes. I don't believe it currently has any way to change.

I'm not actually sure what mode the rank icon sent in the user response even relates to, and so I don't think we should implement it at this time. We can revisit this once the server response for rank and ratings is more sane.

Issue #174, opened 2023-03-21
Leaving a game room still shows user


ArturasNoit:
After joining and leaving a game room it keeps showing user in that room. 
Tried like 5 rooms, it kept showing my username in all of them.

**Note:** Need to test if this issue shows up for other users too or just for user who repeatedly joins different rooms

dusekj4:
well even if you go to other room without click on leave, you will stay there, i guess if you join to other room it will not remove you from the room you are in, actually you cannot get back to that aditional room

dusekj4:
will look to it

PR #191, opened 2023-04-07
#174 bug fix

dusekj4:
# Description
- Fixed problem that enabled user to be in mupltiple lobbys
- Add loader when joining lobby, while testing i found out that if servers are overloaded you just watch nothing happening on join button click, not sure if all fail events are handled, this needs check

# Test

- bug was found when you join the lobby and then switch to custom multiplayer and picked other lobby
- try to recreate beeing in multiple rooms


src/renderer/views/multiplayer/custom.vue
```diff
@@ -226,6 +240,10 @@ async function attemptJoinBattle(battle: SpadsBattle) {
     if (battle.battleOptions.passworded) {
         passwordPromptOpen.value = true;
     } else {
+        loading.value = true;
+        // if user is in a other lobby, leave it
+        api.session.onlineBattle.value && (await api.comms.request("c.lobby.leave"));
```
Jazcash, line 245: I'm not a huge fan of inline conditional expressions like this, would prefer:
```typescript
if (api.session.onlineBattle.value) {
    await api.comms.request("c.lobby.leave");
}
```

src/renderer/views/multiplayer/custom.vue
```diff
@@ -179,7 +182,18 @@ let active = true;
 
 const abortController = new AbortController();
 
+const watchForLobbyJoin = api.comms.onResponse("s.lobby.updated_client_battlestatus").add(() => {
+    loading.value = false;
+});
+
+const watchForLobbyJoinFailure = api.comms.onResponse("s.lobby.join").add((data) => {
+    data.result === "failure" || data.reason === "Battle locked" ? (loading.value = false) : null;
```
Jazcash, line 190: Same as other comment, would prefer this to be an `if` block as it's more readable:
```typescript
if (data.result === "failure" || data.reason === "Battle locked") {
    loading.value = false;
}
```

src/renderer/views/multiplayer/custom.vue
```diff
@@ -4,7 +4,8 @@
 
 <template>
     <div class="flex-row flex-grow gap-md hide-overflow">
-        <div class="flex-col flex-grow gap-md">
+        <Loader v-if="loading"></Loader>
+        <div v-if="!loading" class="flex-col flex-grow gap-md">
```
Jazcash, line 8: You can replace `v-if="!loading"` here with `v-else`

Jazcash (changes requested):

src/renderer/views/multiplayer/custom.vue
dusekj4: changed

src/renderer/views/multiplayer/custom.vue
dusekj4: changed

src/renderer/views/multiplayer/custom.vue
dusekj4: changed

Jazcash (approved):

PR #198, opened 2023-04-10
Getting started on the casual and ranked matchmaking system

knoxy5467:
I am assuming that there is going to be a standardized schema for matchmaking endpoints in tachyon. 

I am developing the MultiplayerAPI under that assumption

created a component for multiplayer views that can be shared between Ranked and Casual. and used to create new Queue Views.

### DIFF
--- a/src/renderer/api/comms.ts
+++ b/src/renderer/api/comms.ts
@@ -1,3 +1,18 @@
+/**
+ * @fileoverview Comms API
+ *
+ * This documentation comment created on 12APR2023 by @knoxy5467
+ *
+ * last updated on 12APR2023 by @knoxy5467
+ *
+ * @class CommsAPI
+ * @classdesc Comms API
+ *
+ * This class is a extension of the TachyonClient class. It is used to communicate with the server.
+ * all communication with the server should be done through this class.
+ *
+ * this includes matchmaking, chat, direct messages, and other lobby related functions.
+ */
 import { Static, TSchema } from "@sinclair/typebox";
 import { arrayToMap, assign } from "jaz-ts-utils";
 import { battleSchema, myUserSchema, TachyonClient } from "tachyon-client";


src/common/model/settings.ts
```diff
@@ -12,4 +12,16 @@ export const settingsSchema = Type.Object({
...
+    queue3v3: Type.Boolean({ default: true }),
+    queue4v4: Type.Boolean({ default: true }),
+    queue5v5: Type.Boolean({ default: true }),
+    queue6v6: Type.Boolean({ default: true }),
+    queue7v7: Type.Boolean({ default: true }),
+    queue8v8: Type.Boolean({ default: true }),
+    queueSmallFFA: Type.Boolean({ default: true }),
+    queueMediumFFA: Type.Boolean({ default: true }),
+    queueLargeFFA: Type.Boolean({ default: true }),
+    queueTeamFFA: Type.Boolean({ default: true }),
```
Jazcash, line 26: rather than having a separate setting for each queue, I think it would be better to store the enabled queues as an array of the queue ids, like `enabledQueues: Type.Array(Type.String(), { default: [] })`

src/renderer/views/multiplayer/queue.vue
```diff
@@ -0,0 +1,64 @@
...
+            <Checkbox v-model="settings.queue6v6" label="6v6" />
+            <Checkbox v-model="settings.queue7v7" label="7v7" />
+            <Checkbox v-model="settings.queue8v8" label="8v8" />
+        </div>
+        <div class="flex-row flex-grow gap-md">
+            <span>FFA</span>
+            <Checkbox v-model="settings.queueSmallFFA" label="Small FFA" />
+            <Checkbox v-model="settings.queueMediumFFA" label="Medium FFA" />
+            <Checkbox v-model="settings.queueLargeFFA" label="Large FFA" />
+            <Checkbox v-model="settings.queueTeamFFA" label="Team FFA" />
```
Jazcash, line 40: Rather than hardcoding all the queues here, we'll want to be fetching them from the server and displaying them dynamically. I know that's not possible to do until that's implemented in the server, but I think it's still possible to write the frontend dynamically and just use some fake mock data for now to show the intention

Jazcash (changes requested):
Make sure to run `npm run fix` if you didn't already.

src/renderer/api/multiplayer.ts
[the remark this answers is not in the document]
knoxy5467: I was under the impression that the comms API was specific to messages within lobbies and direct messaging. if the intent is to have all communications between the server and the client be located within the comms API; would you mind if  I added some commenting / documentation signaling that?

src/renderer/views/multiplayer/queue.vue
knoxy5467: No problem, is the "queue" view component an idea you want to continue with?

src/renderer/api/multiplayer.ts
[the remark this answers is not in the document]
Jazcash: Yeah, go for it

src/renderer/views/multiplayer/queue.vue
Jazcash: I was thinking more individual components for each queue, but typically I don't mind doing that later when it's more obvious which things should be split into their own components, so this is fine for now

PR #199, opened 2023-04-12
added documentation header to comms.ts

knoxy5467:
Added some documentation as lined in my last pull request #198

### DIFF
--- a/src/renderer/api/comms.ts
+++ b/src/renderer/api/comms.ts
@@ -1,13 +1,6 @@
 /**
  * @fileoverview Comms API
  *
- * This documentation comment created on 12APR2023 by @knoxy5467
- *
- * last updated on 12APR2023 by @knoxy5467
- *
- * @class CommsAPI
- * @classdesc Comms API
- *
  * This class is a extension of the TachyonClient class. It is used to communicate with the server.
  * all communication with the server should be done through this class.
  *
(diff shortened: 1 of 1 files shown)


src/renderer/api/comms.ts
```diff
@@ -1,3 +1,18 @@
+/**
+ * @fileoverview Comms API
+ *
+ * This documentation comment created on 12APR2023 by @knoxy5467
+ *
+ * last updated on 12APR2023 by @knoxy5467
+ *
+ * @class CommsAPI
+ * @classdesc Comms API
```
Jazcash, line 9: Can you remove this fluff? I'm keen on adding more overview documentation to classes, but I'd like to keep it to only the bare necessitities. If we ever get into auto-generating frontend documentation based on the code comments, a lot should be able to be inferred from Typescript.

Looking as JSDoc, I think this should probably be:
```
@fileoverview This class is a extension of the TachyonClient class. It is used to communicate with the server. All communication with the server should be done through this class, includeing matchmaking, chat, direct messages, and other lobby related functions.
```

Jazcash (changes requested):

knoxy5467:
Ok will do.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

knoxy5467:
Done removed the fluff

PR #208, opened 2023-05-02
Update LoginForm.vue to include case sensitivity note

Penguin2600:
I had trouble logging in and opened a  ticket and wasted some poor fool's time because I didn't expect Email to be a case sensitive field.

This change updates the email box to have a note about case sensitivity, which will hopefully reduce the support workload for staff.

### DIFF
--- a/src/renderer/components/login/LoginForm.vue
+++ b/src/renderer/components/login/LoginForm.vue
@@ -5,7 +5,7 @@
             <p v-if="loginError" class="error">
                 {{ loginError }}
             </p>
-            <Textbox v-model="email" type="email" label="Email" required validate class="fullwidth" />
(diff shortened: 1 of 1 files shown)


Penguin2600:
> Why is email case sensitive at all? @Jazcash will anything break if it's made case insensitive?

I agree making email case-insensitive would be more intuitive IMO. Happy to try and submit a PR for that if we decide that is a better direction.

Jazcash:
> Why is email case sensitive at all? @Jazcash will anything break if it's made case insensitive?

Not sure, but it's a server problem, not the lobby's.

Penguin2600:
Turns out this is the wrong codebase. Thanks everyone for Pointing me in the right direction, I'll Open a new PR in the correct place.

PR #217, opened 2023-05-22
Button text gets hidden on overflow

### LINKED ISSUE #194: Button text gets hidden on overflow
image

Is an issue for smaller resolution/window size when elements get cramped and start to split lines. This can be fixed by removing the `max-height: 35px;` line in `Control.vue`, but that line was written as a hacky way to make all controls be the same vertical height, for pretty alignment. Instead it might be better for controls to fill their vertical space, but will need to test the whole app to make sure this doesn't break stuff, e.g:

image

The other alternative is forcing button text to always be on same line, then making sure that doesn't break sibling elements from getting pushed off horizontally.

1-alex98:
Fixes #194

### DIFF
--- a/src/renderer/components/controls/Control.vue
+++ b/src/renderer/components/controls/Control.vue
@@ -33,10 +33,8 @@ function onMouseEnter() {
 <style lang="scss" scoped>
 .control {
     display: flex;
-    min-height: 35px;
-    max-height: 35px;
+    align-self: stretch;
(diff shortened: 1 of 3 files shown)


1-alex98:
image

1-alex98:
Clicked through the client could not find any side effects

1-alex98:
Only this 
image
But it was there before

PR #220, opened 2023-05-22
Error on SinglePlayer page

### LINKED ISSUE #219: Error on SinglePlayer page
Problem UNO is no "ais" for engine version in cache.db.
If Problem UNO is fixed Problem DOS is that serialization plugin is not correctly initialized on first start.

1-alex98:
Fixes #219

### DIFF
--- a/src/renderer/api/cache-db.ts
+++ b/src/renderer/api/cache-db.ts
@@ -44,8 +44,6 @@ export class CacheDbAPI extends Kysely<CacheDatabase> {
     }
 
     public async init() {
-        await this.serializePlugin.setSchema(this);
-
         await this.schema
             .createTable("map")
             .ifNotExists()
@@ -122,7 +120,11 @@ export class CacheDbAPI extends Kysely<CacheDatabase> {
             .addColumn("fileName", "varchar", (col) => col.primaryKey())
             .execute();
 
+        await this.serializePlugin.setSchema(this); //might be needed already for migration
+
         await this.migrateToLatest();
+
+        await this.serializePlugin.setSchema(this); //again in case migration changed the database
 
         return this;
     }
--- a/src/renderer/api/content/engine-content.ts
+++ b/src/renderer/api/content/engine-content.ts
@@ -214,7 +214,7 @@ export class EngineContentAPI extends AbstractContentAPI<EngineVersion> {
     }
 
     protected async parseAi(aiPath: string): Promise<EngineAI> {
-        const aiDefinitions = await glob.promise(`${aiPath}/**/{AIInfo.lua,AIOptions.lua}`);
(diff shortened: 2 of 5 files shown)


src/renderer/api/cache-db.ts
```diff
@@ -44,8 +44,6 @@ export class CacheDbAPI extends Kysely<CacheDatabase> {
     }
 
     public async init() {
-        await this.serializePlugin.setSchema(this);
```
1-alex98, line 47: makes no sense to do this before the schema even exists. That works fine unless the db is in creation

src/renderer/api/content/engine-content.ts
```diff
@@ -214,7 +214,7 @@ export class EngineContentAPI extends AbstractContentAPI<EngineVersion> {
     }
 
     protected async parseAi(aiPath: string): Promise<EngineAI> {
-        const aiDefinitions = await glob.promise(`${aiPath}/**/{AIInfo.lua,AIOptions.lua}`);
+        const aiDefinitions = await glob.promise(`${aiPath}/**/{AIInfo.lua,AIOptions.lua}`, { windowsPathsNoEscape: true });
```
1-alex98, line 217: Note Glob patterns should always use / as a path separator, even on Windows systems, as \ is used to escape glob characters. If you wish to use \ as a path separator instead of using it as an escape character on Windows platforms, you may set windowsPathsNoEscape:true in the options. In this mode, special glob characters cannot be escaped, making it impossible to match a literal * ? and so on in filenames.

1-alex98: Let's see if there are any users with ? or * in user name XD

src/renderer/utils/serialize-plugin.ts
```diff
@@ -17,6 +17,7 @@ export class SerializePlugin implements KyselyPlugin {
     protected colTypes = new Map<string, "json" | "datetime" | "boolean">();
 
     public async setSchema(db: Kysely<unknown>) {
+        this.colTypes.clear();
```
1-alex98, line 20: better like this since it is called twice now

src/renderer/config/default-battle.ts
```diff
@@ -31,6 +31,6 @@ export function defaultBattle(mapScriptName?: string) {
             restrictions: [],
         },
         users: [me],
-        bots: [{ playerId: 1, teamId: 1, ownerUserId: me.userId, name: barb.name, aiShortName: barb.shortName, aiOptions: {} }],
+        bots: barb ? [{ playerId: 1, teamId: 1, ownerUserId: me.userId, name: barb.name, aiShortName: barb.shortName, aiOptions: {} }] : [],
```
1-alex98, line 34: will not throw on missing ai but be empty

PR #221, opened 2023-05-22
License

### LINKED ISSUE #214: Please add license file
There are even now contributions from multiple people, and license of those contributions is also not clear (and you should follow up with everybody to confirm they are ok with proposed one). The earlier it's added, the better.

1-alex98:
Fixes #214

Whatever Beyond All Reason is legally

### DIFF
--- a/license.txt
+++ b/license.txt
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023 Beyond All Reason
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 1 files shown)


1-alex98:
@Jazcash @thehobojoe @dusekj4 @p2004a @JasonReed-2 @knoxy5467 @WatchTheFort @Jreed40 @eNTi

1-alex98:
Please comment if MIT License would be ok with you (could be adjusted in the future)

JasonReed-2:
MIT license is ok by me. I don't have a strong opinion here. I am also Jreed40.

knoxy5467:
I agree no strong preference here

Jazcash:
I've actually gone back and forth several times on adding a license file, then having second thoughts and removing it. My initial concern was with other people potentially taking the project and simply swapping out the branding for their own game's, which I believe may cheapen BAR's own experience if people have seen the lobby from other games before coming to BAR. The same way people who played ZK come to BAR and see the same lobby claiming "they stole it!". The other reason was I poured a lot of creative effort into this project, and the idea of someone taking it and building a profitable game on top of it didn't feel good for me.

That being said, I think those are fairly weak reasons for restricting the license in the scheme of things, so I think I'm leaning towards something fairly open again. Gonna merge this

1-alex98:
@Jazcash on the other hand if u have no License any cotributor can sue you for using their copyrighted material or stop you from further using it. I guess that is way worse. Further if anybody can use the client and rebrand it, good for them maybe even good for us if they make valuable improvements. If you are afraid of ppl stealing your project(without contributing) you can also take a liscence that forces anybody to publish their changes to the material. (Something like GNU General Public License https://choosealicense.com/licenses/gpl-3.0/)

1-alex98:
For the future of BAR it is also an operational risk that some contributor drops out gets angry and then his code can not be used further because no license was choosen. Belive me FAF has a lot of fun with mods that were uploaded ages ago and nobody knows the authors anymore and hence legally no modifications can be made to those mods.

PR #228, opened 2023-06-09
CI Fixes And Refactor

wereii:
- I have split the checks (format, lint, future tests) from build so that it runs once and only on linux as windows in CI is painful. As long as the npm steps and CI is alright it should :tm: work even if windows is not covered, it also speeds it up because GA Windows is some slow ~~2016~~ server.
  Like this it should also be possible to run checks for pull requests without building.
- the ~setup-node~ action installs very old npm by default (for node 14, using `nvm use` on Linux I get v9.6.7 by default) so I have added a step that pulls the latest npm, this fixes errors like `cipm can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync ...` or `Cannot read property '@iconify-icons/mdi' of undefined` which happens when newer npm generates the package lock
- Fixed prettier runs by quoting the paths - (linux) shellnanigans - depending on the running shell it could expand the paths beforehand and skip some files that prettier itself would not skip


wereii:
Please try running fix, dev and build on your machine before accepting, I am developing on Linux and I am not sure if the changes to prettier will work on windows

wereii:
- the new `release` script in package.json is important as it tells electron-builder to absolutely definitely do a github release
  https://www.electron.build/configuration/publish#how-to-publish
- tests have been extended with npm build to check that vite can build it (for electron)

**The main differences to current CI** 
- tests (type check, format, lint) are run only on linux host, for speed ups
- tests and build run on different events - build will run only for tags, the idea here is that the final building shouldn't be affected by code changes, for testing electron versions and configuration changes I ~~could~~ should add manual trigger
- instead of using the unmaintained `samuelmeuli/action-electron-builder` action I opted to use `electron-builder` for publishing directly

package.json
```diff
@@ -14,6 +14,7 @@
     "debug": "electron-vite --remote-debugging-port=9223",
     "build": "electron-vite build",
     "compile": "electron-builder --config .electron-builder.config.js",
+    "release": "electron-builder --config .electron-builder.config.js",
```
Jazcash, line 17: what's the reason for this script compared to the existing compile script?

wereii: https://www.electron.build/configuration/publish#how-to-publish 

> Add to scripts in the development package.json:
>
> "release": "electron-builder"
>
>and if you run yarn release, a release will be drafted (if doesn’t already exist) and artifacts published.

wereii: If I understand the docs right it's either this or calling `electron-builder --config .electron-builder.config.js --publish onTag` directly, also it might be just enough to run `npm compile` because it autodetects it's running in CI and defaults to `--publish onTag`

.github/workflows/build.yml
```diff
@@ -1,41 +1,50 @@
...
-                  github_token: ${{ secrets.github_token }}
-                  args: "--config .electron-builder.config.js"
-                  release: ${{ startsWith(github.ref, 'refs/tags/v') }}
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [windows-latest, ubuntu-latest]
+        node-version: [14]
```
Jazcash, line 16: iirc, the reason I had to revert back to node 14 when attempting to upgrade this project to node 18 some time ago was because the previous electron-builder action plugin broke with it. as you've removed the reliance on that plugin, I'm gonna try node 18 again here and see how it goes

Jazcash: actually I'll do that separately, gonna merge this

PR #229, opened 2023-06-10
Rename game

### LINKED ISSUE #201: UI control for changing battle title
An often asked question is how to rename a battle. Should implement perhaps some edit button next to battle title that prompts the user to enter a new one.

1-alex98:
Fixes #201

### DIFF
--- a/src/renderer/components/battle/BattleComponent.vue
+++ b/src/renderer/components/battle/BattleComponent.vue
@@ -1,12 +1,7 @@
 <template>
     <div :class="['battle-container', { singleplayer: isOfflineBattle(battle) }]">
         <div class="header flex-col gap-md">
-            <div class="flex-row flex-center-items">
-                <h1 class="title">{{ battle.battleOptions.title }}</h1>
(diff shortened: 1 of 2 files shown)


1-alex98:
Please rerun CI

Jazcash:
Am I right in thinking that `$rename` only works for the lobby boss? Someone told me the other day (sorry if it was you) that it's possible to find out the lobby boss with `$rename`:

image

I think we should add `boss` as a property to `SpadsBattle` and then use that to only show this rename button if the current user is the boss. I guess this would mean BAR Lobby would have to send `$status` on joining and then parse the DM output, and then also keep track of the boss with a handler for the `* Boss mode enabled for Jazcash` message.

Let me know if you want to give that a shot, otherwise I'll have a go at it.

1-alex98:
I did not implement the parsing of $status because I assume that this be obsolete soon since the new protocol will offer new ways of doingt this

1-alex98:
@Jazcash should I implement the $status thing? If yes I can do so. But if there soon will be another way to do that I would rather wait and do it later

PR #232, opened 2023-06-13
Markdown allow link toggle

JasonReed-2:
Adds the ability to enable links in Markdown. This does not enable links for chats / DMs. Adds the option for campaigns / future use-cases that don't allow end-users to submit markdown.


src/renderer/components/misc/MarkDown.vue
```diff
@@ -20,14 +23,29 @@ marked.use({
...
+const allowedAttributes: string[] = [];
+
+if (props.allowLinks) {
+    allowedTags.push("a");
+
+    allowedAttributes.push("href");
+    allowedAttributes.push("target");
+
+    markdownRenderer.link = (href, title, text) => {
+        return `<a href="${href}" target="_blank">${text}</a>`;
```
1-alex98, line 37: Check for XSS please!!! Both directly in the text, the href and with the javascript protocol

1-alex98: https://brutelogic.com.br/blog/alternative-javascript-pseudo-protocol/

1-alex98: more on XSS in general https://owasp.org/www-community/attacks/xss/

1-alex98: Sadly a correct filter for it is not easy

JasonReed-2: That part of the code just controls how the link tag is rendered. I added so that "target=_blank" attribute is added. The html generated by marked is later put through DOMPurify.sanitize. Also: allowLinks should not be exposed to end-user submitted entry. This pull does not enable link entry for DMs or lobby chat or anywhere else that end-users can submit markdown.

Issue #223, opened 2023-06-03
Multiplayer Custom Battles Runtime should not show seconds


kozak-codes:
Since all the timestamps update at a different time, it appears very busy when the runtime is updating. In my opinion, it would clean this up if we just use HH:mm. This is very useful information to determine if a lobby will be restarting soon but seconds are not pertinent to that information.

wereii:
Hmm, is it just me or now it looks like it could be confused with minutes:seconds that are stuck. 

![image](https://github.com/beyond-all-reason/bar-lobby/assets/8504482/4aa35284-aa85-46a5-9042-ab05e645f981)

wereii:
This seems a bit too verbose but does the trick
![image](https://github.com/beyond-all-reason/bar-lobby/assets/8504482/d524ea0c-13ad-4db8-a031-7af5248e420a)

PR #238, opened 2023-06-18
Sort custom lobbies based on "best lobby" instead of by players

### LINKED ISSUE #237: Custom battle match scoring & sorting
The current custom battle sorting is a little simplistic, and doesn't take into account all the factors that I manually use when selecting a battle. Rather than sorting custom battles based on players, we should sort them based on how much of a match that battle is. For example, an ideal match that I look for when selecting a lobby is as follows:

1. Not locked or passworded
2. Has at least 1 player
3. Is almost full
4. Is not full (if it is full, has a small join queue)
5. is within my skill range
6. The median skill is close to my own

Other potential factors for sorting could include:
- Whether the current user is in that lobby already (eg, they need to rejoin)
- A friend is in this lobby
- Whether the match is ending soon. If there are no good matches, there may be a match that will end within the next few minutes. "Ending soon" could be calculated based on the median match time for that game type. More advanced versions of this could be calculated by the server (Eg, a significant resource/energy different for an individual team)
- Someone I've blocked is in the lobby
- A map that I've rated highly is being played
- The map that is being played has been downloaded
- The number of spectators
- Is it a noob lobby, and have I played less than 10 games?

What I propose is a scoring system that, based on all the above factors, assigns a score to each battle and then sorts the battles based on that score. That score can be shown on the right side in the form of a summary of why it's a good match and color coded to visually distinguish that matches potential to jump into a battle. This way, we can still sort by individual headers such as title, map, runtime, or you can just sort by what the game thinks is the best match for you.

The way it would work is simple. Each battle would have the following "factor" checks done to them. If they pass the check, then "points" are added to that battle. We then sort custom battles by these "points" in descending order. The most notable factor is then indicated in the "match" section which visually distinguishes the sort.

Here is an example factor chart, that would  need to be tweaked as we gain more data on the potential factors:

| Factor                                           | Points                                                       | Description                                                          |
|--------------------------------------------------|--------------------------------------------------------------|----------------------------------------------------------------------|
| Locked                                           | -5                                                           |                                                                      |
| Passworded                                       | -5                                                           |                                                                      |
| Has no players                                   | -10                                                          | 0/x players                                                          |
| Is almost full                                   | (-5 / half of required players) * remaining players + 5      | Only applied when lobby is at least half full, not applied when full |
| Full                                             | -5 - queueSize * 5                                           |                                                                      |
| Within Skill Range                               | +10                                                          |                                                                      |
| Median Skill Close to own                        | 1 / ( standard deviations away from median + 1/5 )           |                                                                      |
| Already in lobby                                 | +50                                                          |                                                                      |
| A friend is in this lobby                        | +5 * friends in lobby                                        |                                                                      |
| Match ending soon                                | Standard deviation from this match type median duration * 10 | Only applied if duration > median duration                           |
| Someone I've blocked is in this lobby            | -5 * blocked in lobby                                        |                                                                      |
| A map that I've rated highly is being played     | +map rating out of 5                                         |                                                                      |
| The map that is being played has been downloaded | +3                                                           |                                                                      |
| The number of spectators                         | log(x) * 5                                                   |                                                                      |
| Is it a noob lobby, and am I a noob?             | +15                                                          |                                                                      |

The item that is highest in this list is noted to the user. That is what is sorted, perhaps visually shown on the right hand side of the table. We can choose to visually distinguish (eg, set tag color) by the _top_ items or only items that match a certain score.

Gonna work on a quick little demo PR to show how this can improve selecting a match significantly.

kozak-codes:
Closes #237 
Fix https://github.com/beyond-all-reason/bar-lobby/issues/223

The first lobby should be the lobby that gives you the best chance of jumping into a game as fast as possible

Adds a "scoreBattle" function to the end of the battles list that will score values based on provided values in the issue list. This also re-orders the columns of the custom battles to place the battle "best status" on the far left side. Scores are adjusted so that running battles, or battles ending soon, will display above locked battles but still show locked battles on the first page, if you have them enabled.

Let me know what you think and if you'd like me to polish further.

Example battle list result:
image

The battle that you want to join will always be placed at the top. One thing to note is the "ending soon" battles that are full and potentially will be finishing up shortly. Here, we also have a lobby full of spectators and it's placed higher because it might be a bit more renown

### DIFF
--- a/src/renderer/views/multiplayer/custom.vue
+++ b/src/renderer/views/multiplayer/custom.vue
@@ -28,7 +28,7 @@
                     class="p-datatable-sm"
                     selectionMode="single"
                     :sortOrder="-1"
-                    sortField="playerCount.value"
+                    sortField="score"
                     paginator
                     :rows="16"
                     :pageLinkSize="20"
@@ -63,6 +63,13 @@
                                 <div v-if="data.bots.length > 0" class="flex-row flex-center-items gap-xs" style="gap: 4px">
                                     <Icon :icon="robot" height="17" />{{ data.bots.length }}
                                 </div>
+                            </div>
+                        </template>
+                    </Column>
+                    <Column header="Best Battle" sortable sortField="score">
+                        <template #body="{ data }">
+                            <div class="flex-row flex-center-items gap-md">
+                                {{ data.primaryFactor }}, {{ Math.round(data.score * 100) / 100 }}
                             </div>
                         </template>
                     </Column>
@@ -116,6 +123,13 @@ const hostBattleOpen = ref(false);
 const searchVal = ref("");
 const selectedBattle: Ref<SpadsBattle | null> = shallowRef(null);
 const settings = api.settings.model;
+
+interface ScoredSpadsBattle extends SpadsBattle {
+    score: number;
+    factors: { [factorName: string]: number };
+    primaryFactor: string;
+}
+
 const battles = computed(() => {
     let battles = Array.from(api.session.battles.values());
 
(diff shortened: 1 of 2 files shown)


src/renderer/views/multiplayer/custom.vue
```diff
@@ -28,24 +28,18 @@
...
-                        </template>
-                    </Column>
-                    <Column header="Runtime" sortable sortField="runtimeMs.value">
+                    <Column header="Best Battle" sortable sortField="score">
                         <template #body="{ data }">
-                            <div v-if="data.runtimeMs.value >= 1">
-                                {{ getFriendlyDuration(data.runtimeMs.value) }}
+                            <div class="flex-row flex-center-items gap-md">
+                                {{ data.primaryFactor }}
+                                <!-- {{ Math.round(data.score * 100) / 100 }} -->
```
Jazcash, line 42: remove this

src/renderer/views/multiplayer/custom.vue
```diff
@@ -159,15 +175,101 @@ const battles = computed(() => {
...
-    return battles;
+    return scoredBattles;
 });
 
+function scoreBattle(battle: SpadsBattle) {
+    let score = 0;
+    let factors = {};
+    let primaryFactor = "";
+
+    const inBattle = battle.players.value.find((p) => p.username === "kozan");
```
Jazcash, line 193: think you left this in 😉

src/renderer/views/multiplayer/custom.vue
```diff
@@ -159,15 +175,101 @@ const battles = computed(() => {
...
+        selectedBattle.value = scoredBattles[0];
     }
 
-    return battles;
+    return scoredBattles;
 });
 
+function scoreBattle(battle: SpadsBattle) {
+    let score = 0;
+    let factors = {};
```
Jazcash, line 190: Could you try typing this object? I believe this will probably fail the type-check anyway

src/renderer/views/multiplayer/custom.vue
```diff
@@ -191,6 +293,23 @@ if (active.value) {
...
+${Object.entries(data.factors)
+    .map(([factorName, factorScore]) => {
+        return `${factorName.padEnd(20, " ")}: ${factorScore.toFixed(2)}`;
+    })
+    .join("\n")}
+`;
+
+    console.log(scoreExplanation, event.data);
+}
+
```
Jazcash, line 312: I get that this is for debugging, but I think it should be removed before PR is merged

Jazcash (changes requested):
This is an interesting idea, and I think this would be more useful to players. The scoring is subjective and arbitrary, and can make the order seem kind of weird, but I think players would appreciate it regardless.

With this I think it'd make sense to scrap the existing `Runtime` column and merge that functionality into this new column so it says something like `Running for 40:15`. I'd also rename the column header to `Status`, and rename `Passworded` to `Private`.

Would be curious to know what others think of this idea.

src/renderer/views/multiplayer/custom.vue
Jazcash: Oh it didn't fail the type check 🤔 Either way more explicit typing would be appreciated here

kozak-codes:
> This is an interesting idea, and I think this would be more useful to players. The scoring is subjective and arbitrary, and can make the order seem kind of weird, but I think players would appreciate it regardless.
> 
> With this I think it'd make sense to scrap the existing `Runtime` column and merge that functionality into this new column so it says something like `Running for 40:15`. I'd also rename the column header to `Status`, and rename `Passworded` to `Private`.
> 
> Would be curious to know what others think of this idea.

Thanks for the feedback! The scoring is definitely much more arbitrary, but it's meant to be something that can be fine tuned over time with feedback. I'll implement all your suggestions, however the "Best Match" versus "Status" was meant to indicate that the sort is based on best match, and not based on the status of the game. IE you would expect "status" to sort the status alphabetically in that case. In this case, "Best Match" simply indicates _why_ something is a good or bad match, which can double as a status. What's cool about using "Best Match" is the sort indicator points to the first match, so players reading left to right see "Best Match" with an arrow towards the first match, so the sort indicator serves two purposes in that way.

src/renderer/views/multiplayer/custom.vue
kozak-codes: Not sure it's great as a tooltip, or maybe I just formatted it weirdly
image

kozak-codes:
I stole the fixed runtime function from https://github.com/beyond-all-reason/bar-lobby/pull/226

src/renderer/views/multiplayer/custom.vue
Jazcash: Eh, changed my mind. I was thinking just display the final score, but even that's not really relevant to end users. This function is useful to devs, but maybe just comment out the `console.log` for now and add a note saying what the function has been left in for

Jazcash:
> the "Best Match" versus "Status" was meant to indicate that the sort is based on best match, and not based on the status of the game. IE you would expect "status" to sort the status alphabetically in that case. In this case, "Best Match" simply indicates _why_ something is a good or bad match, which can double as a status. What's cool about using "Best Match" is the sort indicator points to the first match, so players reading left to right see "Best Match" with an arrow towards the first match, so the sort indicator serves two purposes in that way.

True, although tbh I'm kind of sceptical about the usefulness of having this column be sortable at all. I can't imagine there being a use case really where users would want to see the battles with the lowest scores first. Tbh that goes for the title, map and locked columns too. Even for the playercount and runtime columns, I think it's more likely that users would go to the last page to find the opposite of what they're looking for. So I'm leaning towards changing the behaviour of clicking a column heading to sort it by the most likely desired order, rather than flipping between ascending and descending. I guess that's not exactly relevant to this PR though.

The hope for the future is good matchmaking implementation will reduce the amount of custom battles, so I'm hesistant to go overboard with the amount of functionality, but overall I'm happy enough with this to merge. Will wait a bit in case anyone else weighs in.

Jazcash (approved):

PR #246, opened 2023-09-05
Refactor Scored Spads Battle Data

### LINKED ISSUE #245: BattlePreview has fatal error due to not recognizing scored battles as abstract battles.
**Issue**
When clicking on "Multiplayer" tab, I receive this error:
image

**Impact**
Users can't open the custom battle list browser. Also: letting this error sit with the electron app open will cause a massive slowdown as it eats up all the computers memory :(

**Investigation**
Looking at this code:
image

When debugged: props.battle.contenders is actually ComputedRefImpl object. To access the value, it should be "props.battle.contenders.value".

So the code is recognizing the battle as a Replay when it actually should recognized as an AbstractBattle and enter the if block where the ComputedRefImpl objects are handled appropriately.

So why is the battle being treated as a replay and not a battle?

Looking in `type-checkers.ts`:
image

So we only recognize a battle as an abstract battle if `instanceof` abstract battle. Explanation of `instanceof` from stackoverflow: https://stackoverflow.com/questions/2449254/what-is-the-instanceof-operator-in-javascript

tldr: `instanceof` checks that the object has the constructor for the given class.

So that means that an object must be instantiated using the constructor of the class [or subclass] to be recognized by `instanceof`.

Looking at `custom.vue` in `scoreBattle` function:
image

So while typescript recognizes this as a `scoredSpadsBattle`: `instanceof` will not because the object wasn't created used the constructor of a subclass of `AbstractBattle`.

**Solution**
Fold the scored battles data into the `SpadsBattle` class. Have the `custom.vue` component interact with the scored battles data that is stored within the `SpadsBattle`.

JasonReed-2:
Fixes #245

### DIFF
--- a/src/renderer/model/battle/abstract-battle.ts
+++ b/src/renderer/model/battle/abstract-battle.ts
@@ -15,6 +15,7 @@ export interface BattleConfig<T extends BattleOptions = BattleOptions> {
 }
 
 export abstract class AbstractBattle<T extends BattleOptions = BattleOptions> {
+    public readonly discriminator: "AbstractBattle";
     public readonly battleOptions: T;
     public readonly bots: Bot[];
     public readonly users: User[];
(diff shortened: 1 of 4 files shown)


JasonReed-2:
After discussing: the better approach to fixing #245 is to have the battles in the list remain as SpadsBattle objects that contain the methods of the parent classes.

PR #254, opened 2023-10-08
Fix autosuggestion keyboard selection & navigation

### LINKED ISSUE #253: Keyboard selecting an auto-suggestion automatically submits the command
**Issue**
When a user uses the keyboard to select an autosuggestion: pressing enter automatically submits the auto-suggestion instead of populating it into the text-box then letting the user submit it themselves.

**Impact**
This is a problem for commands that require the user to enter arguments after the command itself. Example: the "!clusterstatus" command requires you to enter the <clusterName> as an argument after the command.

JasonReed-2:
Fixes #253 & an unrelated issue with navigating up and down when there is no autosuggestion entries.


p2004a:
Underlying issue is obsolete, sorry, there was plenty of refactoring and changes since then (both in code and direction)

Issue #161, opened 2023-03-21
No way to tell is someone is AFK.


Descyphal:
AFK indicator doesn't seem to exist.

JasonReed-2:
This is implemented here in chobby:
https://github.com/beyond-all-reason/BYAR-Chobby/blob/7341db50c38953455965d1ef04835b4be7cf9419/LuaMenu/widgets/gui_away_tracker.lua#L46

Hectate:
Portions of the client will require non-public development to determine requirements, scope, visual design, and user experience. Because public contributions can not be accepted until that is complete, this issue is being closed to ensure volunteers do not spend time on pull requests that cannot be merged. When and where possible, new issues will be created with adequate information for volunteers to contribute to the final requirements.

PR #257, opened 2024-01-13
Client Tracking AFK

JasonReed-2:
#161 [does not resolve it]. 

Implements client AFK tracking. To be used when the tachyon endpoints for AFK are ready.


p2004a:
The underlying implementation of custom battles is gone entirely, so unfortunatelly this is obsolete by now

PR #266, opened 2024-08-27
wsl single player flow - click around errors

tetrisface:
I had some errors pop up when trying singleplayer in wsl.


p2004a:
You need to provide more information about each of the changes, the changes look very random to me.
- why tool version?
- why eslint change?
- why switch from Jaz ts utils to lodash?
- why adding esbuild?

tetrisface:
I wouldn't put too much effort into this pr. It was kinda provided as is if someone had use for it. i just wanted to do something but after seeing the oauth branch i realize there are bigger fish to fry.
> * why tool version?
this is a file for the asdf runtime version manager. it is recommended in the teiserver readme and since that should be ran in tandem with this repo locally in a dev situation I thought it would be reasonable to set that up here too. It's also a great tool that I think can simplify a lot of developers' environments. perhaps above all it aids in communication as well. for example without it it would be impossible to know what node version i used making this pr.
> * why eslint change?
my environment complained that the boolean is actually a string. so a type fix.
> * why switch from Jaz ts utils to lodash?
i had runtime errors from jaz utils but not from lodash.
> * why adding esbuild?
for me it was needed to build the project

tetrisface:
changing to draft. not sure if this is needed until people with same environment tests it.

PR #268, opened 2024-09-22
Fixes #264

### LINKED ISSUE #264: AI fast copy config
currently in chobby when you right click to add a duplicate ai, it does not duplicate the ai in full
only the type of ai is copied.
it would be best for the last ai added to the room or client to be fully copied
which includes things like bonus

ill redo this issue later
this much more about having a note instead of a fully fillled out issue at this time

Mark104:
Implemented as a dropdown option on a bot participant as shown below, requires validation by @drivver44 as if he feels this is enough for now (perhaps last added AI right click could be an additional issue built from this?).

image

Two watchouts:
  - Needs testing on an online lobby, this is currently only tested in custom singleplayer (only just started contributing so not fully setup yet!).
  - Has been implemented with a json stringify/parse which should handle nested objects correctly, if these objects are just data models (no methods).

Please feel free to let me know if any of this is incorrect or if I can do anything better, this is my first contribution for this project and felt I could add a very little bit of value with a simple issue.

### DIFF
--- a/src/renderer/components/battle/BotParticipant.vue
+++ b/src/renderer/components/battle/BotParticipant.vue
@@ -20,7 +20,7 @@
 import { Icon } from "@iconify/vue";
 import robot from "@iconify-icons/mdi/robot";
 import { MenuItem } from "primevue/menuitem";
-import { Ref, ref } from "vue";
+import { Ref, ref, toRaw } from "vue";
 
 import LuaOptionsModal from "@/components/battle/LuaOptionsModal.vue";
 import TeamParticipant from "@/components/battle/TeamParticipant.vue";
@@ -80,8 +80,7 @@ async function configureBot() {
 }
 
 function setBotOptions(options: Record<string, unknown>) {
-    console.log("Set bot options");
-    props.battle.setBotOptions(props.bot.name, options);
+    props.battle.setBotOptions(props.bot.playerId, options);
 }
 </script>
 
--- a/src/renderer/components/battle/BotParticipant.vue
+++ b/src/renderer/components/battle/BotParticipant.vue
@@ -3,7 +3,7 @@
         <div class="flex-row flex-center">
             <Icon :icon="robot" :height="16" />
         </div>
-        <div>{{ bot.name }}</div>
+        <div>{{ bot.name }} - {{ botNameAffix }}</div>
     </TeamParticipant>
     <LuaOptionsModal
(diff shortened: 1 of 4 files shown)


src/renderer/components/battle/BotParticipant.vue
```diff
@@ -59,6 +63,13 @@ function kickBot() {
     props.battle.removeBot(props.bot);
 }
 
+// Duplicates this bot and its settings and gives it a new player id.
+function duplicateBot() {
+    let duplicatedBot = JSON.parse(JSON.stringify(props.bot));
```
p2004a, line 68: [`structuredClone`](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) should be cleaner in what you are trying to do here vs pasing object via JSON serialization. It also should be better for typing.

Also change `let` to `const`.

Mark104: Thanks! Adjustments made in a877be / 959a9fc, alongside a fix for options not being set by some sort of Id.

src/renderer/model/battle/abstract-battle.ts
```diff
@@ -90,6 +90,19 @@ export abstract class AbstractBattle<T extends BattleOptions = BattleOptions> {
...
+            const isBot = !("userId" in participant);
+            if (isBot) {
+                if (participant.playerId === playerId) {
+                    return true;
+                }
+            }
+            return false;
+        });
+    }
+
```
Jazcash, line 105: I think this should `.find` on `this.bots` instead of `this.participants`, then you can remove `User` from the return type

Jazcash (approved):

src/renderer/model/battle/abstract-battle.ts
Mark104: Thanks for the suggestion!

Addressed in 7bcc76.

src/renderer/model/battle/abstract-battle.ts
```diff
@@ -90,6 +90,16 @@ export abstract class AbstractBattle<T extends BattleOptions = BattleOptions> {
         });
     }
 
+    // Returns a bot by it's PlayerId. Returns undefined if not found..
+    public getBotByPlayerId(playerId: number): Bot | undefined {
+        return this.bots.find((bot) => {
```
p2004a, line 95: This could have been just `(bot) => bot.playerId === playerId` but whatever.

PR #270, opened 2024-09-24
Feat/new tachon implementation changes

Mark104:
Proposed changes to fit the new tachyon specification with development server "server5". These include:

- Disabling of Steam Auth. I'm unsure if this is live in the latest tachyon / telserver but was throwing errors currently.
- Config changes to direct client to server5.
- Login.vue login state changes. Broadly kept the same but unless login automatically is set in the settings, users will be currently presented with two login options, BAR / Offline.
- Added some offline checks in commands view to handle offline state without errors.
- Commented out the opening of the battle for singleplayer/custom, as this looks like its not implemented currently (presumably new schema for tachyon).

This should allow the client to run in both offline / online, with online functionality very much limited till implementation of tachyon is complete.

I believe for now this should be kept on this "new-tachyon-implementation" branch. Pending a review of the custom offline battle logic, as currently compared to main, we lose the ability to start custom offline battles. Other offline uses such as starting scenarios appears to work correctly.


Mark104:
Added another commit to fit hardcoded issues raised by Mongy.

Mark104:
Well this is weird, this post install hasent been changed. Perhaps a failure of the machine?

Issue #248, opened 2023-09-30
Changing tabs causes styling issues


JRugen:
When changing tabs, the content of the new tab loads underneath the content of the old tab. This quickly adjusts, but shouldn't be shown to the user when changing tabs.

Jazcash:
I think this was intentional but intended to be temporary: https://github.com/beyond-all-reason/bar-lobby/commit/8a9b592700f393b61d986c55cd99d1b044642ca2

The transition mode `out-in` is desired but was causing a much more critical bug with the whole Vue renderer crashing: https://github.com/beyond-all-reason/bar-lobby/issues/61

I believe there's still open PRs to vue-core that I and several others have submitted to fix this and some similar issues but they still haven't been merged. Best solution for us to fix the issue right now (if it's bothering us too much) is to fix the issue in our local vue package using [patch-package](https://www.npmjs.com/package/patch-package), which we're already doing for one of the vue bugs (see patches folder in root).

Issue #179, opened 2023-04-01
Vue crash related to Suspense + Transition + RouterView + Teleport


Jazcash:
This affects quite a lot of our pages. Have done some digging and found the route cause to be related to using Suspense, Transition, RouterView and Teleport together, along with a top level await that resolves quickly: https://github.com/vuejs/core/issues/7966

Repro steps for current master in BAR Lobby:
1. Visit a page that uses Teleport in some way (e.g. singleplayer battle)
2. Visit a page that redirects using `await api.router.push(blahblah` (e.g. player profile)

Result:
![image](https://user-images.githubusercontent.com/1434248/229288448-981a6957-b0ed-46e1-862e-e6c75b939c96.png)

```
runtime-dom.esm-bundler.js:36 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'nextSibling')
    at nextSibling (runtime-dom.esm-bundler.js:36:31)
    at removeFragment (runtime-core.esm-bundler.js:6208:20)
    at remove2 (runtime-core.esm-bundler.js:6173:17)
    at unmount (runtime-core.esm-bundler.js:6142:17)
    at Object.remove (runtime-core.esm-bundler.js:6529:21)
    at unmount (runtime-core.esm-bundler.js:6126:28)
    at unmountComponent (runtime-core.esm-bundler.js:6230:13)
    at unmount (runtime-core.esm-bundler.js:6115:13)
    at unmountChildren (runtime-core.esm-bundler.js:6259:13)
    at unmount (runtime-core.esm-bundler.js:6133:17)
```

Jazcash:
ooo that's a cool package, go for it

Issue #61, opened 2022-09-03
Swapping between routes very quickly crashes the Vue renderer


Jazcash:
```
Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.
```

Seems easier to repro on multiplayer custom battle list page, possibly related to suspense api

Jazcash:
https://lightrun.com/answers/nuxt-framework-error-when-fast-navigating-failed-to-execute-insertbefore-on-node

Jazcash:
Probably related to https://github.com/vuejs/core/issues/7966#issuecomment-1488801586

1-alex98:
https://github.com/vuejs/core/issues/8105

1-alex98:
Removing transition removes the issue 
https://github.com/beyond-all-reason/bar-lobby/blob/b2f3b1ec71e34ac89482904ea8160b14ea885b04/src/renderer/App.vue#L32

1-alex98:
Maybe we should do that until issue is fixed in vue or we have a better solution

1-alex98:
Also removing out-in works. Does not look as good tho

PR #271, opened 2024-09-25
Feat/page transition fix

Mark104:
Vue update to latest version and a fix for the page transition issue the lobby has been experiencing for a while. This issue has a bit of history to it by the looks of it from #61 #179 #248 and the intent is for this PR to solve this.

**This requires a merge of #270 PR first as I've had to base this branch off those changes to allow for testing.**

Issue:

- Changing tabs would show a styling issue (where the content of the previous page was shown for a short period and the new content shifted to the bottom).
- This was caused by a workaround to fix a bigger issue, whereby switching pages very fast would sometimes throw a critical issue and break the client.
- A part of this issue was due to a Vue internal issue. This was highlighted to the Vue team by @Jazcash and the workaround used till the team implemented a fix.

Proposed fix:

- Updating to the latest Vue version that should include fixes to this issue.
- Testing the site to ensure that upgrading the vue version did not cause any build issues. This appears to be fine.
- This exposed that actually #179 was a different issue, related to using an await in a component setup. Therefore I've moved this code to a onMounted function which appears to solve the issue.

Results in:

- No styling issues when transitioning between pages.
- No observed critical errors when changing lots of tabs really quickly.


Mark104:
Closing this PR for. We've got alot of things to fix to get this passing the checks again (such as the seperate tachyon-client). Will get this and other PRS merged into a single branch and then start to look to fix the branch to pass these checks.

Sorry for any time wasted, but just looking to get this merged in when its stable and passing.

PR #274, opened 2024-09-29
Node version documentation

ngraham20:
- with Node.js version
- to indicate that Volar has been moved to Vue - Official


README.md
```diff
@@ -27,15 +27,35 @@ BAR Lobby is a new WIP lobby for the RTS game [Beyond All Reason](https://www.be
...
 
 ### Requirements
 
--   [Node.js](https://nodejs.org/en/download/)
+-   [Node.js 18](https://nodejs.org/en/download/)
 
-### Project setup:
+#### Installing Node 18 on Bash
+```bash
+curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.0/install.sh | bash
```
p2004a, line 38: I would just document what you need to have node 18, and for *unix* base systems we recommend nvm, and link to that docs.

I would definitely not copy nvm install instruction, pinned to specific version.

.gitignore
```diff
@@ -15,6 +15,7 @@ pnpm-debug.log*
 
 # Editor directories and files
 .idea
+.vscode
```
p2004a, line 18: Why remove all of vscode recommended settings and so on?

README.md
```diff
@@ -27,15 +27,35 @@ BAR Lobby is a new WIP lobby for the RTS game [Beyond All Reason](https://www.be
...
+
+nvm install 18
+
+node -v # should print `v18.20.4`
+```
+
+#### Installing Node 18 on Fish
```
p2004a, line 45: IMHO, fish is way too obscure to have dedicated section about it in readme, also, another reason why just pointing at nvm docs will be better.

.gitignore
ngraham20: didn't realize there were specific settings tbh, I'm just used to gitignoring the IDE settings files. I'll update the PR and take that back out :+1:

README.md
ngraham20: in my particular case, the nvm docs only *had* bash instructions, so it took a hot sec to figure it out at all. Totally fair if we don't want to include it though

ngraham20:
@p2004a let me know if you'd prefer I do a squash commit btw. I didn't see any contributing documentation, so I'm not sure if you want the PR's commit history in here :joy:

p2004a (approved):
We can squash as part of accepting the PR, and by default we will be squashing ;)

PR #276, opened 2024-10-03
Fix crash when opening the game options

JasonReed-2:
This fixes a crash when opening the game options. The issue was due to luaObjectObj.desc sometimes being null.


p2004a:
Thank you!

PR #277, opened 2024-10-20
migration to electron-forge, secure app

bcdrme:
# Refactoring

## Initial goal

Update and secure the new lobby, be idiomatic to the electron way
- One renderer process for rendering (aka the VueJS app), 
- one main process to handle "backend stuff" (aka interacting with the host filesystem, calling external services etc...)

It meant removing all node integration stuff, re-enabling all security features including sandbox mode and context isolation

--> refactoring is needed to move node api calls to the main process
--> ended up having to cleanup some logic when rendering concerns and "backend" concerns were intertwined

I quickly found out that it wouldn't be possible without a big update of deps, of electron, of node.

Some dependencies were problematic :
- sqlite3 -> native lib, very difficult to package into the final executable (I couldn't do it)
- jazcashutils -> transitive dependencies leaking into our project, can't import this lib in the renderer because of node api dep
- map-parser -> transitive deps
- sdfz-demo-parser -> transitiv deps

## Changes

Disclaimer, some things are going to be a bit controversial.

### Adding preload.ts to conform to the context isolation security feature:
- preload.ts is the step to bind ipcRenderer to ipcMain 
- *.service.ts, it's the entrypoint of the main process' api

### Removed sqlite3 and cache-db.ts, replaced by IndexedDB (with dexiejs) in the renderer process.
sqlite3, I really tried to make it work but I hit a wall when it came to generating a working executable.
The concerns were to have a place where we can easily look through what's installed, it's particularly usefull for maps because we can query against our db.
Theses concerns can all be adressed in IndexedDB (the db that ships with chrome). 

### Rewrites
rewrite of main/main window to be more idiomatic with electron's way of doing things
rewrite of abstract battle 
rewrite of the initialization step of content apis to scan the filesystem for packages, maps, replays. Using chokidar to detect automatically new replays in the replay folder

introduction of the store pattern in Vue so that different components can react to changes.


src/main/main-window.ts
```diff
@@ -1,97 +1,115 @@
...
+        title: "Beyond All Reason",
+        fullscreen: settings.fullscreen,
+        frame: true,
+        show: true,
+        minWidth: 1440,
+        minHeight: 900,
+        paintWhenInitiallyHidden: true,
+        webPreferences: {
+            preload: path.join(__dirname, "../build/preload.js"),
+            zoomFactor: zoomFactor(primaryDisplay),
```
p2004a, line 27: This seem to mess up scale on my HiDPI display when electron runs under x11 (`ELECTRON_OZONE_PLATFORM_HINT=x11`), when running under native wayland `ELECTRON_OZONE_PLATFORM_HINT=wayland` the scale is ok. I have to press ctrl+0 to reset zoom level after each restart.

Why is touching zoom level needed?

forge.config.ts
```diff
@@ -0,0 +1,75 @@
...
+import { VitePlugin } from "@electron-forge/plugin-vite";
+
+const config: ForgeConfig = {
+    packagerConfig: {
+        asar: false, // Cannot enable asar, it breaks map parsing
+    },
+    rebuildConfig: {},
+    makers: [
+        new MakerSquirrel({}),
+        new MakerZIP({}, ["darwin"]),
```
p2004a, line 17: MacOs not supported

forge.config.ts
```diff
@@ -0,0 +1,75 @@
...
+                icon: "src/renderer/assets/images/icon.png",
+                homepage: "https://www.beyondallreason.info/",
+                license: "MIT",
+                name: "bar-lobby",
+                productName: "BAR Lobby",
+            },
+        }),
+        new MakerDeb({}),
+        new MakerFlatpak({}),
+        new MakerSnap({}),
```
p2004a, line 31: I would start only with appimage and windows installer.

prettier.config.mjs
```diff
@@ -0,0 +1,16 @@
+export default {
+  useTabs: false,
+  trailingComma: "es5",
+  tabWidth: 4,
+  semi: true,
+  singleQuote: false,
+  printWidth: 200,
```
p2004a, line 7: 140 is already massive and IMHO too large, 200 is quite brutal. I would reduce it all to max 120. I'm raising it as prettier output is not clear at the moment anyway so reformat is needed.

src/common/jaz-ts-utils/delay.ts
p2004a, line 1: I would move the jaz-ts-utils, map-parse, and sdfz-demo-parser to some clearly named folder like "vendored" or "third_party" etc.

src/main/main-window.ts
```diff
@@ -1,97 +1,115 @@
...
+        mainWindow.show();
+        mainWindow.focus();
+    });
+
+    mainWindow.webContents.on("render-process-gone", (event, details) => {
+        console.error(details);
+    });
+
+    mainWindow.webContents.setWindowOpenHandler(({ url }) => {
+        shell.openExternal(url);
```
p2004a, line 45: I wouldn't do that https://benjamin-altpeter.de/shell-openexternal-dangers/

p2004a (changes requested):
`npm run lint` is all red. `npx prettier --check .` is not clean either.

For now I've just tried to run locally and what I noticed via high level skimming

src/main/main-window.ts
bcdrme: I tried to keep aspect ratio and the overall size of the UI elements consistent across resolutions and dpi. But that's not it.
I think the way to go is to stop using px as a css unit, and to use rem. Need some more investigating.

I'm going to remove this for now.

src/main/main-window.ts
bcdrme: What was I thinking...

Would be grateful if you double check this change :
https://github.com/beyond-all-reason/bar-lobby/pull/277/commits/e6d2a0ffd1515a378746c77af2d5d3176f89997a

src/main/main-window.ts
p2004a: I don't think that's expected, e.g. I never want to have exactly the same size of elements as measured by real world mm across devices. I want things to be smaller on my 13" laptop, and larger on my 27" screen.

When I want to scale something, there are OS settings for that purpose.

rem/em won't be useful as they are still just size relative to some other element on site sized with absolute sizes px/pt/in, and those are automatically scaled by browser based on OS scaling, so that px isn't really a real pixel.

bcdrme: While that may be true for a desktop app, for a game lobby though ... I think we want to stay consistent across display sizes. 

But I'm not 100% set on this, I'm no design expert. I have an ultra widescreen though and it looks dumb in fullscreen in the current implementation.

p2004a: I realized maybe I misunderstood, which option do you want:
- Absolute size in real world units like `cm` consistent across devices?
- The relative size consistent, so e.g. the height of the "SINGLEPLAYER" button be always let's say 5% of total window height?

bcdrme: I think we want the same content displayed on any screen size. If I take SC2 as a reference I don't think it's "responsive" like a classic webapp. Electron by default behaves like a responsive webapp, and it looks dumb on my ultrawide, lots and lots of empty space. A discussion need to happen with the designers anyway, let's maybe move this discussion on Discord

https://www.youtube.com/watch?v=byBIOsDDZiE

forge.config.ts
p2004a: We need a format for Linux that is not tied to any distributions. I see in that thread that there are some third-party packagers for appimage.

I also see that electron forge points at electron updater which points at https://www.electronjs.org/docs/latest/api/auto-updater#platform-notices for autoupdates

> There is no built-in support for auto-updater on Linux

so, using electron forge might be dead on arrival

src/main/main-window.ts
p2004a: Hmm, I believe you still need to have `mainWindow.webContents.setWindowOpenHandler` with `{ action: 'deny' }` as without that it will open new window, that we likely ~never want.

I do it in similar way in launcher where I know it's 100% single page application: https://github.com/beyond-all-reason/spring-launcher/blob/2c0a6e17b50ab04ea592c53963c571e5ccb071ba/src/update_dialog.js#L49-L68

forge.config.ts
bcdrme: Need some thinking.

But even if we decide to go back to electron-builder it shouldn't be a big deal and it can be decided later.

src/main/main-window.ts
bcdrme: I had a bad understanding of this. I simply deny everything now.

prettier.config.mjs
bcdrme: it's this project's default. I'm used to a lot less myself, I tried reducing it but it quickly becomes unreadable in vue templates. 

Overall prettier and eslint configuration could be improved a lot. Could be done outside of this PR

bcdrme:
fixed a bunch of eslint errors, the rest of them are  in heavily wip parts of the app.
prettier --write . ran with current configuration

p2004a (changes requested):
I've pushed a few more fixes: source maps loading and fix test CI workflow.

---

Okey, I've went roughly through the whole thing, but wasn't able to dive into every detail.

I believe it's move in correct direction for the main process vs renderer separation, updating and cleanup of dependencies was also very very important, but, there are multiple regressions in this PR:
1. Electron forge is breaking Linux packaging and releases, current github actions workflow won't work either: maybe that's acceptable as we are not close to release, but we might need to revert to electron builder simply because it offers the functionality we need.
2. Dependencies were vendored even though we own them: we can't manage multiple forks of common libraries, the map parser and demo parses fixes need to be upstreamed and here cleaned up. I would not want to block this PR on this, but I would like this cleanup to follow soon so that implementations don't drift apart.
3. Some functionality like loading of local game checkout (`sdd`) was dropped, so it kinda looks like in-the-middle refactoring
4. Linting still fails
5. TypeScript type checking fails

All points are problematic, but because it's WIP project, and we don't have to provide test builds yet points 1-3 are acceptable for me, but, the points 4-5 are blocking.

Especially the type checking should just work. It's the basic reason why this repo is in TypeScript and not pure JavaScript and is in my opinion important for any contributors to help navigate the code base and validate things are correctly tied together. When I clone main branch of any repo, this is the very basic thing, next to tests (sad we don't have any here), and "does it start" that I just expect to work if I'm to be able to work on the project.

bcdrme:
> I've pushed a few more fixes: source maps loading and fix test CI workflow.
> 
> Okey, I've went roughly through the whole thing, but wasn't able to dive into every detail.
> 
> I believe it's move in correct direction for the main process vs renderer separation, updating and cleanup of dependencies was also very very important, but, there are multiple regressions in this PR:
> 
> 1. Electron forge is breaking Linux packaging and releases, current github actions workflow won't work either: maybe that's acceptable as we are not close to release, but we might need to revert to electron builder simply because it offers the functionality we need.
> 2. Dependencies were vendored even though we own them: we can't manage multiple forks of common libraries, the map parser and demo parses fixes need to be upstreamed and here cleaned up. I would not want to block this PR on this, but I would like this cleanup to follow soon so that implementations don't drift apart.
> 3. Some functionality like loading of local game checkout (`sdd`) was dropped, so it kinda looks like in-the-middle refactoring
> 4. Linting still fails
> 5. TypeScript type checking fails
> 
> All points are problematic, but because it's WIP project, and we don't have to provide test builds yet points 1-3 are acceptable for me, but, the points 4-5 are blocking.
> 
> Especially the type checking should just work. It's the basic reason why this repo is in TypeScript and not pure JavaScript and is in my opinion important for any contributors to help navigate the code base and validate things are correctly tied together. When I clone main branch of any repo, this is the very basic thing, next to tests (sad we don't have any here), and "does it start" that I just expect to work if I'm to be able to work on the project.

Thank you for your thorough review. It was a lot of work. I'm just gonna comment on 4.  and 5. since they're blocking.

In order for these checks to pass, I would, in a perfect world, have to fix these functionalities. Or implement them even. It's online related stuff and as we both know, it's heavily wip. 

So I think we can either delete thoses files (not ideal as it's a good blueprint for actually finishing the implementation) or comment out everything. 

Wdyt @p2004a ?

p2004a:
Makes sense, nice way would be to fix rendering using some stub backend model, e.g. return fake list of battles etc. I expect all the communication layer needs to be rewritten, but the UI elements that exists here already _might_ be helpful in the future, but also definitely not all.

I've pushed this branch here as https://github.com/beyond-all-reason/bar-lobby/tree/pr-277-refactoring-with-broken-multiplayer so it's not lost and clearly marked, and here I would personally vote for deleting the code.

p2004a (changes requested):
I see you updated package-json etc, I see CI doesn't pass failing installing dependencies. I've run into this specific thing already myself when updating deps, downgrading to electron 33.0.1 resolved it for me.

Given that you updated package-lock.json, I will need to again retest if it still works under windows too.

bcdrme:
It's so annoying, I had the same issue locally. I had to remove package-lock.json to be able to `npm install`.
In the mean time `bun install` works flawlessly, no issue at all. Weird bug.

p2004a:
Any hard issues here?

bcdrme:
Can't find why `npm install` is failing on linux. 
```
Node 20.18
npm 10.9.0
```
Downgrading electron doesn't work. Tried on node 22.9.0, it's not working either.
Current workaround is to use `bun`, `bun install` works no problemo ... 

I guess it's an issue with npm on linux specifically.

p2004a:
Downgrading to electron 33.0.1 works as I've said, and visible in CI.

Let me check windows again.

p2004a (changes requested):
It did not work on Windows. I've fixed it again with just reinstall. I also see that bun, is dropping all integrity hashes.

I will just then recommend let's not use bun to modify package-lock.json, and always use npm.

---

After doing another test, I see all passes in CI, but map parsing is broken, both on Windows and Linux:

```
[00:03:09.942] ERROR: map-content.ts - Error parsing map: quicksilver_remake_1.24.sd7
[00:03:09.942] ERROR: map-content.ts - Cannot find module '/home/p2004a/Workspace/BAR/bar-lobby/.vite/build/parse-map-worker.js'
    err: {
      "type": "Error",
      "message": "Cannot find module '/home/p2004a/Workspace/BAR/bar-lobby/.vite/build/parse-map-worker.js'",
      "stack":
          Error: Cannot find module '/home/p2004a/Workspace/BAR/bar-lobby/.vite/build/parse-map-worker.js'
              at Module._resolveFilename (node:internal/modules/cjs/loader:1232:15)
              at Module._load (node:internal/modules/cjs/loader:1058:27)
              at c._load (node:electron/js2c/node_init:2:16955)
              at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:189:12)
              at MessagePort.<anonymous> (node:internal/main/worker_thread:186:26)
              at [nodejs.internal.kHybridDispatch] (node:internal/event_target:820:20)
              at MessagePort.<anonymous> (node:internal/per_context/messageport:23:28)
      "code": "MODULE_NOT_FOUND",
      "requireStack": []
    }
```

bcdrme:
> Downgrading to electron 33.0.1 works as I've said, and visible in CI.
> 
> Let me check windows again.

Maybe an issue with my setup then 🤔
Fyi, bun doesn't use nor generate package -lock.json, it uses it's own lock file format (bun.lockb iirc).

Will check why map parsing broke.

bcdrme:
Ah indeed it does work. 
I was still trying to `npm install` against `33.0.2` due to `^` in `"electron": "^33.0.1",`. Forcing to 33.0.1 works. 

I opened an issue https://github.com/electron/electron/issues/44529

p2004a (approved):

PR #280, opened 2024-11-13
App icons

CollinsSpencer:
This PR adds custom app icons as detailed here: https://www.electronforge.io/guides/create-and-add-icons

I have verified that the icon gets set in installer in Windows 11 and the game app in Windows 11 and WSL Ubuntu (with the deb installer).

---

This is the default gif that the Windows installer opens while the game installs. [source](https://github.com/electron/windows-installer/blob/main/resources/install-spinner.gif)

installer_gif

I replaced it with one of the promokit images, but it could benefit from a dedicated, BAR-designed loading gif.


p2004a:
FYI: As I was commenting on other recent PRs, we aren't likely to keep Electron Forge as it doesn't support AppImage and automatic updates via it on Linux and we need to revert to the electron builder.

PR #284, opened 2024-12-23
Create distributables for Windows + Linux

CollinsSpencer:
## Problem
* This app needs to be packaged for both Windows + Linux and be set up for auto updating.
* Context: https://discord.com/channels/549281623154229250/1308862441450639431

## Solution
* Use `electron-builder` to create these distributables, since `electron-builder` supports AppImage + auto-update on Linux.
* Still use Electron Forge for the Vite development.
* This PR creates the NSIS `.exe` for Windows and the `.AppImage` for Linux.
* This PR does not fully implement auto-updates.


bcdrme (changes requested):
These changes are looking good but I cannot run the generated AppImage on my machine (Fedora 41). It shows a white screen with the following error message in the terminal 

`(node:16549) electron: Failed to load URL: http://localhost:5173/ with error: ERR_CONNECTION_REFUSED`


Screenshot From 2025-01-02 11-45-58

bcdrme:
I've got it fixed :+1:  
see https://github.com/beyond-all-reason/bar-lobby/tree/forge-with-electron-builder

PR #290, opened 2025-01-19
Authentication with teiserver using oauth2

bcdrme:


src/main/config/server.ts
```diff
@@ -1,4 +1,3 @@
-export const serverConfig = {
-    host: "server4.beyondallreason.info",
-    port: 8202,
-};
+export const OAUTH_AUTHORIZATION_SERVER_URL = `https://server5.beyondallreason.info`;
```
geekingfrog, line 1: Later, we’ll want to make that configurable, similar to how chobby can target prod, staging, or any url. But that's going to be another piece of work altogether.

src/main/json/model/account.ts
```diff
@@ -1,6 +1,6 @@
 import { Type } from "@sinclair/typebox";
 
 export const accountSchema = Type.Object({
-    email: Type.String({ default: "" }),
     token: Type.String({ default: "" }),
+    refreshToken: Type.String({ default: "" }),
```
geekingfrog, line 5: nitpick: nullable string instead of defaulting to empty string? I'm not a big fan of sentinel values like that.

src/main/oauth2/redirect-handler.ts
```diff
@@ -10,38 +10,31 @@
...
 
-export class RedirectHandler {
+export default class RedirectHandler {
     private path: string;
     private server: http.Server;
     private error?: Error;
     private callbackUrl?: URL;
 
-    constructor(signal?: AbortSignal) {
+    constructor() {
```
geekingfrog, line 19: what happened to the abort signal? If the user kill the client while it’s waiting for the callback, what happens to the local server?

src/main/oauth2/utils.ts
```diff
@@ -0,0 +1,148 @@
...
+    const response = await fetch(WELL_KNOWN_OAUTH_AUTHORIZATION_SERVER_URL);
+    if (response.status !== 200) {
+        const error = `Failed to fetch OAuth2 authorization server metadata: ${response.status} ${response.statusText}`;
+        log.error(error);
+        throw new Error(error);
+    }
+    const { authorization_endpoint, token_endpoint } = await response.json();
+    if (!authorization_endpoint || !token_endpoint) {
+        const error = "Invalid OAuth2 authorization server metadata";
+        log.error(error);
```
geekingfrog, line 29: would be nice to also dump the response there. If you only have the logs it’s a bit sparse to debug.

src/main/oauth2/utils.ts
```diff
@@ -0,0 +1,148 @@
...
+        throw new Error(error);
+    }
+    // TODO: Remove this hack once the server is fixed
```
geekingfrog, line 32: Add the link: https://github.com/beyond-all-reason/teiserver/pull/555 it is, at time of writing, merged, but not yet deployed.

src/main/oauth2/utils.ts
```diff
@@ -0,0 +1,148 @@
...
+
+// Careful with shell.openExternal. https://benjamin-altpeter.de/shell-openexternal-dangers/
+function openInBrowser(url: string) {
+    if (!["https:", "http:"].includes(new URL(url).protocol)) return;
+    // Additional checks to prevent opening arbitrary URLs
+    if (!url.startsWith(OAUTH_AUTHORIZATION_SERVER_URL)) return;
+    shell.openExternal(url);
+}
+
+function createUrlWithQuerystring(baseUrl: string, params: Record<string, string | number | boolean>): string {
```
geekingfrog, line 46: I am not convinced of the value of this function tbh. It's a tiny wrapper around `querystring.stringify` and seems to only add a layer of indirection. The type signature is a mouthful as well.

src/main/oauth2/utils.ts
```diff
@@ -0,0 +1,148 @@
...
+            client_id: OAUTH_CLIENT_ID,
+            code,
+            code_verifier,
+            redirect_uri,
+        });
+        const tokenResponse = await fetch(tokenUrl, {
+            method: "POST",
+        });
+        if (tokenResponse.status !== 200) {
+            const error = `Failed to fetch OAuth2 token: ${tokenResponse.status} ${tokenResponse.statusText}`;
```
geekingfrog, line 80: I think it would also be valuable to grab the response's body `await tokenResponse.text()` and log that as well. Because many things can get you a 400, and the body will tell you more, like "code verifier invalid length" or whatnot

src/main/oauth2/utils.ts
```diff
@@ -0,0 +1,148 @@
...
+            log.error(error);
+            throw new Error(error);
+        }
+        const { access_token, refresh_token, expires_in } = await tokenResponse.json();
+        if (!access_token || !refresh_token) {
+            const error = "Invalid OAuth2 token response";
```
geekingfrog, line 86: Same, let's dump the response there, useful for debugging.

src/main/services/auth.service.ts
```diff
@@ -0,0 +1,31 @@
...
+            await accountService.saveRefreshToken(refreshToken);
+            startTokenRenewer((expiresIn / 2) * 1000);
+        } catch (error) {
+            log.error("Error during login");
+            accountService.wipe();
+            throw error;
+        }
+    });
+    ipcMain.handle("auth:logout", async () => {
+        // TODO maybe revoke refresh token as well here
```
geekingfrog, line 23: no need, when you renew the access token, you get a new pair of "token/refresh token" and the previous pair is invalidated automatically

src/renderer/store/me.store.ts
```diff
@@ -17,4 +17,19 @@ export const _me = reactive({
     status: "offline",
 } as CurrentUser);
 
+async function login() {
+    await window.auth.login();
+    _me.isOnline = true;
```
geekingfrog, line 22: not quite true, because you still need to actually login and open the websocket connection. That can fail for a number of reasons, you're banned for example.
You can leave that in but keep in mind getting an auth token is not the same as actually logging in

src/renderer/views/login.vue
```diff
@@ -105,4 +82,53 @@ connect();
...
+.login-button:hover::before {
+    box-shadow: 0 0 15px rgba(34, 197, 94, 0.4);
+}
+
+.buttons-container {
+    display: flex;
+    gap: 20px;
+    flex-direction: column;
+    align-items: center;
+}
```
geekingfrog, line 133: any chance to have a screenshot with the PR to have an idea how that looks?

geekingfrog (commented):
some comments but nothing major.

src/main/oauth2/utils.ts
```diff
@@ -0,0 +1,148 @@
...
+        });
+        const tokenResponse = await fetch(tokenUrl, {
+            method: "POST",
+        });
+        if (tokenResponse.status !== 200) {
+            const error = `Failed to fetch OAuth2 token: ${tokenResponse.status} ${tokenResponse.statusText}`;
+            log.error(error);
+            throw new Error(error);
+        }
+        const { access_token, refresh_token, expires_in } = await tokenResponse.json();
```
p2004a, line 84: in principle, refresh token is optional, not required to be returned by the server, at least add a TODO or comment that this client requires it.

src/main/oauth2/utils.ts
```diff
@@ -0,0 +1,148 @@
...
+    const tokenResponse = await fetch(tokenUrl, {
+        method: "POST",
+    });
+    if (tokenResponse.status !== 200) {
+        const error = `Failed to renew token: ${tokenResponse.status} ${tokenResponse.statusText}`;
+        log.error(error);
+        throw new Error(error);
+    }
+    const { access_token, refresh_token, expires_in } = await tokenResponse.json();
+    if (!access_token || !refresh_token) {
```
p2004a, line 126: Same as above, refresh token is optional:
> The authorization server MAY issue a new refresh token, in which case
> the client MUST discard the old refresh token and replace it with the
> new refresh token.

at least add a TODO or just a comment that this client requires it.

src/main/oauth2/pkce.ts
```diff
@@ -0,0 +1,21 @@
...
+    const buf = Buffer.alloc(len);
+    for (let i = 0; i < buf.length; i++) {
+        const idx = randomInt(0, charSpace.length);
+        buf.write(charSpace.charAt(idx), i);
+    }
+    const hash = createHash("sha256");
+    hash.update(buf);
+    const challenge = hash.digest("base64url");
+    return [buf.toString(), challenge];
+}
```
p2004a, line 21: Or, up to you, feel free to just resolve:

```js
const charSpace = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-._~";
const len = 47; // must be between 43 and 128
const verifier = Array(len).fill(0).map(() => charSpace[randomInt(0, charSpace.length)]).join('');
const challange = createHash("sha256").update(verifier).digest("base64url");
return {verifier, challange};
```

src/main/config/server.ts
bcdrme: Sure thing
Let's talk about it in Discord

src/main/oauth2/redirect-handler.ts
bcdrme: If the user kills the client, not an issue. Everything shuts down. 
But we can potentially leak a server if you hit f5 on the app during this process. I'll implement a timeout of some sort to avoid this corner case

src/main/json/model/account.ts
bcdrme: Can't set it to nullable because of the implementation of file-store.ts
It would need a refactor, not worth it imho

src/main/config/server.ts
p2004a: I will want to connect it to this https://discord.com/channels/549281623154229250/927564746104905728/1328446417336930306, I've started working on it.

src/main/oauth2/redirect-handler.ts
```diff
@@ -10,38 +10,31 @@
 import http from "node:http";
 import { AddressInfo } from "node:net";
 
-export class RedirectHandler {
+export default class RedirectHandler {
```
CollinsSpencer, line 13: Auth + redirecting are all in `tachyon-client` package already. Any reason not to use that?

src/main/oauth2/utils.ts
```diff
@@ -0,0 +1,148 @@
...
+    authorizationEndpoint: string;
+    tokenEndpoint: string;
+}> {
+    const response = await fetch(WELL_KNOWN_OAUTH_AUTHORIZATION_SERVER_URL);
+    if (response.status !== 200) {
+        const error = `Failed to fetch OAuth2 authorization server metadata: ${response.status} ${response.statusText}`;
+        log.error(error);
+        throw new Error(error);
+    }
+    const { authorization_endpoint, token_endpoint } = await response.json();
```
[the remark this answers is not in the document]
bcdrme: it's returning `any` yeah, and it's a fake typed language :hankey: 
it's not validating anything.
so destructuring will result in two `any` variables, I could check more on the next line instead of just testing if it's defined
not necessarily worth it

src/main/oauth2/redirect-handler.ts
p2004a: FYI: the redirector here is already a copy of what was implemented in tachyon-client, you can compare: https://github.com/beyond-all-reason/tachyon-client/commit/a923447faea5db786309908a85f2d0199ad4c51e, so it's not from scratch.

Overall we decided that vendoring the tachyon client code into bar lobby at the moment is a very good idea because:

1. There will be quite rapid iteration on the protocol, and needing to change tachyon-protocol package, then tachyon-client package, and only then bar-lobby seems a bit much indirection.
2. There aren't any other tachyon-client consumers planned yet

src/main/oauth2/utils.ts
bcdrme: To guarantee a good UX we're gonna go with refresh tokens are not optional. 
I wouldn't go for long lived token

PR #297, opened 2025-01-30
Add support for favoriting maps

### LINKED ISSUE #250: Player Map Favorites
[Link to Discord suggestion.](https://discord.com/channels/549281623154229250/1116768416267771975)

The user should be able to easily find & select their favorite maps within the map selection list through marking them as a favorite.

lbulej3:
Adds support for being able to favorite maps and filter according to that.

Current Add/Remove button is _placeholder_ as I will redo the whole Map details page once this is merged - as such it's purely functional.

I have toyed with the idea of having the heart on the map card be clickable as well, but decided against it because it would have to remain visible on hover while everything else fades away - can be done though if desired (to me it feels better UX-wise).

Resolves #250


src/main/content/maps/map-data.ts
```diff
@@ -6,4 +6,5 @@ export type MapData = MapMetadata & {
     };
     isInstalled?: boolean;
     isDownloading?: boolean;
+    favorite?: boolean;
```
bcdrme, line 9: I would have named that `isFavorite` for consistency

src/renderer/components/maps/MapOverviewCard.vue
```diff
@@ -92,6 +100,13 @@ const imageUrl = computed(() =>
     }
     transition: background-image 0.1s ease-in-out;
 }
+.favorite {
```
bcdrme, line 103: maybe add a backdrop / background of some sort to improve readability on light maps  (snow maps especially)

bcdrme (changes requested):

src/renderer/components/maps/MapOverviewCard.vue
lbulej3: Added the same backdrop that is used for attributes to keep it consistent

bcdrme (approved):

PR #319, opened 2025-02-28
typescript refactor : proposed changes for a more strictly typed usage of typescript

ehsan-ahmadi-g:
typescript refactor : 
- proposed changes for a more strictly typed usage of typescript
- fixes several bugs and crashes
- added multiple `type.ts` files for folders and sub-folders to be able to avoid some of the type duplications

### DIFF
--- a/src/main/content/engine/engine-content.ts
+++ b/src/main/content/engine/engine-content.ts
@@ -13,6 +13,7 @@ import { extract7z } from "@main/utils/extract-7z";
 import { contentSources } from "@main/config/content-sources";
 import { AbstractContentAPI } from "@main/content/abstract-content";
 import { CONTENT_PATH } from "@main/config/app";
+import { DownloadEngine } from "@main/content/game/type";
 
 const log = logger("engine-content.ts");
 
@@ -84,7 +85,9 @@ export class EngineContentAPI extends AbstractContentAPI<string, EngineVersion> 
             });
     }
 
-    public async downloadEngine(engineVersion: string) {
+    public downloadEngine: DownloadEngine = async (engineVersion) => {
+        if (!engineVersion) return;
+
         try {
             if (this.isVersionInstalled(engineVersion)) {
                 return;
@@ -118,7 +121,7 @@ export class EngineContentAPI extends AbstractContentAPI<string, EngineVersion> 
                 headers: { "Content-Type": "application/7z" },
                 onDownloadProgress: (progress) => {
                     downloadInfo.currentBytes = progress.loaded;
-                    downloadInfo.totalBytes = progress.total;
+                    downloadInfo.totalBytes = progress.total || 0;
                     this.downloadProgress(downloadInfo);
                 },
             });
@@ -138,7 +141,7 @@ export class EngineContentAPI extends AbstractContentAPI<string, EngineVersion> 
         } catch (err) {
             log.error(err);
         }
-    }
+    };
 
     public async uninstallVersion(version: EngineVersion | string) {
         if (typeof version === "object") {
--- a/src/main/content/game/game-content.ts
+++ b/src/main/content/game/game-content.ts
@@ -17,6 +17,7 @@ import { Scenario } from "@main/content/game/scenario";
 import { SdpFileMeta, SdpFile } from "@main/content/game/sdp";
 import { PrDownloaderAPI } from "@main/content/pr-downloader";
 import { CONTENT_PATH, GAME_VERSIONS_GZ_PATH } from "@main/config/app";
+import { GetScenarios } from "@main/content/game/type";
 
 const log = logger("game-content.ts");
 const gunzip = util.promisify(zlib.gunzip);
@@ -138,7 +139,7 @@ export class GameContentAPI extends PrDownloaderAPI<string, GameVersion> {
         return this.parseAis(luaai);
     }
 
-    public async getScenarios(gameVersion: string): Promise<Scenario[]> {
+    public getScenarios: GetScenarios = async (gameVersion) => {
         try {
             const version = this.availableVersions.values().find((version) => version.gameVersion === gameVersion);
             assert(version, `No installed version found for game version: ${gameVersion}`);
@@ -182,14 +183,19 @@ export class GameContentAPI extends PrDownloaderAPI<string, GameVersion> {
             log.error(`Error getting scenarios: ${err}`);
             return [];
         }
-    }
+    };
 
-    public async uninstallVersionById(gameVersion: string) {
+    public async uninstallVersionById(gameVersion: string | undefined) {
+        if (!gameVersion) return;
+
         const version = this.availableVersions.values().find((version) => version.gameVersion === gameVersion);
-        await this.uninstallVersion(version);
+
+        if (version) await this.uninstallVersion(version);
     }
 
     public async uninstallVersion(version: GameVersion) {
+        if (!version) return;
+
         // TODO: Uninstall game version through prd when prd supports it
         assert(!version.packageMd5.endsWith(".sdd"), "Cannot uninstall local/custom game versions");
         await fs.promises.rm(path.join(CONTENT_PATH, "packages", `${version.packageMd5}.sdp`));
--- a/src/main/content/maps/map-image-worker.ts
+++ b/src/main/content/maps/map-image-worker.ts
@@ -4,11 +4,11 @@ if (isMainThread) {
     throw new Error("This script should be run in worker thread.");
 } else {
     // listen to messages from the main thread
-    parentPort.on("message", async (imageSource: string) => {
+    parentPort?.on("message", async (imageSource: string) => {
         try {
             const response = await fetch(imageSource);
             const arrayBuffer = await response.arrayBuffer();
-            parentPort.postMessage({ imageSource, arrayBuffer });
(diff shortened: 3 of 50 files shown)


ehsan-ahmadi-g:
@bcdrme

src/main/content/engine/engine-content.ts
```diff
@@ -84,7 +85,9 @@ export class EngineContentAPI extends AbstractContentAPI<string, EngineVersion>
             });
     }
 
-    public async downloadEngine(engineVersion: string) {
+    public downloadEngine: DownloadEngine = async (engineVersion) => {
+        if (!engineVersion) return;
```
bcdrme, line 89: You should either throw an Error or assert that engineVersion is defined and not just return silently. It's important because we want it to crash so we know something is not right.

That's true every time you did this in this PR.

src/main/content/engine/engine-content.ts
```diff
@@ -118,7 +121,7 @@ export class EngineContentAPI extends AbstractContentAPI<string, EngineVersion>
                 headers: { "Content-Type": "application/7z" },
                 onDownloadProgress: (progress) => {
                     downloadInfo.currentBytes = progress.loaded;
-                    downloadInfo.totalBytes = progress.total;
+                    downloadInfo.totalBytes = progress.total || 0;
```
bcdrme, line 124: 0 is not a great default value for total. You can be sure someone somewhere is doing a currentBytes / totalBytes for computing a progress percentage, thus dividing by 0. 

I'd set it to -1.

src/main/content/maps/map-image-worker.ts
```diff
@@ -4,11 +4,11 @@ if (isMainThread) {
     throw new Error("This script should be run in worker thread.");
 } else {
     // listen to messages from the main thread
-    parentPort.on("message", async (imageSource: string) => {
+    parentPort?.on("message", async (imageSource: string) => {
```
bcdrme, line 7: You should throw an error instead of relying on `?`. It's way easier to debug instead of failing silently. 
`if(!parentPort) throw new Error("...")`

You must review everywhere where you used `?`, most of the time you should either assert that the value is defined or throw an Error.

src/main/model/start-script.ts
```diff
@@ -1,9 +1,9 @@
 export interface Game {
     [key: string]: unknown;
 
-    myplayername: string;
-    gametype: string;
-    mapname: string;
+    myplayername?: string;
+    gametype: string | undefined;
```
bcdrme, line 5: `gametype?` for consistency

src/main/oauth2/oauth2.ts
```diff
@@ -77,7 +77,7 @@ export async function authenticate(): Promise<TokenResponse> {
         openInBrowser(url);
         const callbackUrl = await redirectHandler.waitForCallback();
         log.debug(`Received callback URL: ${callbackUrl}`);
-        const code = callbackUrl.searchParams.get("code");
+        const code = callbackUrl.searchParams.get("code") || "";
```
bcdrme, line 80: throw an error if we have no code instead of defining a default value

src/renderer/audio/audio.ts
```diff
@@ -29,7 +29,7 @@ class AudioAPI {
         this.settings = await window.settings.getSettings();
         console.debug("Loading music files...");
         for (const filePath in musicFiles) {
-            const name = filePath.split("/").pop().split(".")[0];
+            const name = filePath.split("/").pop()?.split(".")[0] || "";
```
bcdrme, line 32: You should console.error that this file has an issue with it's name and skip to the next file.

bcdrme (changes requested):
Your changes are to satisfy ESLint issues, but you need to consider the business implications (especially on the Node side of things; using `?` in Vue components is mostly ok).

I didn't put a comment every single place something is not handled properly. Most of the time you want to throw an Error instead of failing silently. If you start executing some function and some prerequisites are not met, **you want it to fail**. You do not want to return undefined or continue running, that's the rule of thumb. Basically don't use `?` too often and don't `if(!myParam) return` because it's gonna cause headaches in the long run.

ehsan-ahmadi-g:
@bcdrme

src/main/content/game/game-content.ts
```diff
@@ -182,14 +183,23 @@ export class GameContentAPI extends PrDownloaderAPI<string, GameVersion> {
...
 
-    public async uninstallVersionById(gameVersion: string) {
         const version = this.availableVersions.values().find((version) => version.gameVersion === gameVersion);
-        await this.uninstallVersion(version);
+
+        if (version) await this.uninstallVersion(version);
     }
 
     public async uninstallVersion(version: GameVersion) {
+        if (!gameVersion) {
```
CollinsSpencer, line 199: `gameVersion` is not defined here.

src/main/content/engine/engine-content.ts
```diff
@@ -13,6 +13,8 @@ import { extract7z } from "@main/utils/extract-7z";
 import { contentSources } from "@main/config/content-sources";
 import { AbstractContentAPI } from "@main/content/abstract-content";
 import { CONTENT_PATH } from "@main/config/app";
+import { DownloadEngine } from "@main/content/game/type";
+import { assert } from "assert";
```
CollinsSpencer, line 17: This import isn't used.

```suggestion
```

CollinsSpencer (commented):
I found a couple of errors after running `npm run typecheck`.

tsconfig.web.json
```diff
@@ -1,6 +1,6 @@
 {
     "extends": "./tsconfig.shared.json",
-    "include": ["src/renderer/env.d.ts", "src/renderer/typed-router.d.ts", "src/renderer/**/*", "src/renderer/**/*.vue", "src/preload/*.d.ts"],
+    "include": ["src/main/", "src/renderer/env.d.ts", "src/renderer/typed-router.d.ts", "src/renderer/**/*", "src/renderer/**/*.vue", "src/preload/*.d.ts"],
```
CollinsSpencer, line 3: We ideally don't want to include the main folder here. Everything in `main` runs in a Node context, so it should error if we try to use browser-only APIs, like `window`. It's covered by the `tsconfig.node.json` file, so things like intellisense should work correctly in VSCode.

I just checked and to actually enforce this, we'd need to specify `"lib": ["esnext"],` in `tsconfig.node.json`, since the default imports the DOM libs. It adds a few new errors, so it might be out of scope for this PR though.

```suggestion
    "include": ["src/renderer/env.d.ts", "src/renderer/typed-router.d.ts", "src/renderer/**/*", "src/renderer/**/*.vue", "src/preload/*.d.ts"],
```

src/renderer/views/library/replays.vue
```diff
@@ -165,12 +165,12 @@ function openReplaysFolder() {
     window.shell.openReplaysDir();
 }
 
-function watchReplay(replay: Replay) {
+function watchReplay(replay: Replay | undefined | null) {
```
bcdrme, line 168: This function shouldn't accept undefined or null

src/renderer/views/library/replays.vue
```diff
@@ -165,12 +165,12 @@ function openReplaysFolder() {
...
     window.game.launchReplay(replay);
 }
 
-function showReplayFile(replay: Replay) {
-    window.shell.showReplayInFolder(replay.fileName);
+function showReplayFile(replay: Replay | undefined) {
```
bcdrme, line 172: This function shouldn't accept undefined, this is not good. This should be handled by the caller imo

src/renderer/store/maps.store.ts
```diff
@@ -23,7 +23,7 @@ export const mapsStore = reactive({
     };
 });
 
-export async function getRandomMap(): Promise<MapData> {
+export async function getRandomMap(): Promise<MapData | undefined> {
```
bcdrme, line 26: This should never return undefined since we are checking for `mapCount === 0`

src/renderer/audio/audio.ts
```diff
@@ -29,10 +29,16 @@ class AudioAPI {
...
         console.debug("Loading music files...");
         for (const filePath in musicFiles) {
-            const name = filePath.split("/").pop().split(".")[0];
+            const name = filePath.split("/").pop()?.split(".")[0];
             console.debug(name);
             const src = musicFiles[filePath];
             const volume = this.settings.musicVolume / 100;
+
+            if (!name) {
+                console.error(`something wrong with audio file named ${name}`);
```
bcdrme, line 38: this will always print out undefined ^^
maybe print out the full filePath

src/main/game/game.ts
```diff
@@ -93,35 +99,35 @@ export class GameAPI {
                 detached: true,
             });
 
-            this.gameProcess.stdout.on("data", (data) => {
+            this.gameProcess?.stdout?.on("data", (data) => {
```
bcdrme, line 102: if we don't have a gameProcess defined, something is very wrong. This function should throw an error instead of using `this.gameProcess?`

bcdrme (changes requested):
One last push ! Thank you for your work in fixing all those warnings

ehsan-ahmadi-g:
@bcdrme

bcdrme (approved):

PR #327, opened 2025-03-17
Scaling and window size fixes

p2004a:
- Minimal 1080p is unreasonable, especially when on HiDPI screens which would require native 4k
- Please don't force scale 1, everything is tiny and unusable


bcdrme:
With this set of changes this is what the UI looks like at 720p at zoom level 0. 

Screenshot From 2025-03-19 18-56-26
Screenshot From 2025-03-19 18-56-19

And here is the target for comparison 

Screenshot From 2025-03-19 18-58-04

src/main/main.ts
```diff
@@ -84,8 +84,6 @@ app.on("window-all-closed", () => app.quit());
 // Security
 app.enableSandbox();
 
-app.commandLine.appendSwitch("high-dpi-support", "1");
-app.commandLine.appendSwitch("force-device-scale-factor", "1");
```
bcdrme, line 88: Removing this line makes electron take into account the accessibility settings from the OS (size of text). Which is a good thing ? I guess ? But it comes with it's set of challenges when designing the UI.

I think ideally the app will have it's components always take the same amount of screenspace when in fullscreen independently of the pixel density.  Some more thoughts need to be put into this.

bcdrme: I need to elaborate a little bit. 
So my idea for this was to have the app compute a scaling factor automatically based on the height in pixels of your screen. That way independently of the resolution, the app would retain the same aspect, everything would be displayed the same on all computers. That was my idea at least, it's a bit buggy on startup but this is the logic behind the F11 keystroke, if you hit that twice you'll see that the UI is scaled up and down.

p2004a: > Removing this line makes electron take into account the accessibility settings from the OS (size of text). Which is a good thing ?

It forces everything to be scaled to 1, ignoring overall scaling factor. So the lobby window looks like this for me in comparison to browser window:

Screenshot From 2025-03-19 22-43-04

It's 2x too small, on every single possible way, it's not only text. Text scaling has some other way to be propagated: I can separately set the "scale in display setting" and "font scale" in some other location.

p2004a:
I prefer to have some stuff maybe not perfect, but functional, because without, it's just entirely broken:

Screenshot From 2025-03-19 22-53-33

Moreover, with changes here as on the branch, it can get to a state where it looks good:

image


Console says:
```
Enter full screen event
Window size:  [ 1440, 868 ]
Zoom factor:  0.8037037037037037
```

Thing is cut off in windowed mode, when console says:

```
Leave full screen event
Window size:  [ 1440, 900 ]
Zoom factor:  1
```

p2004a:
One more thing: I can still control the zoom level with `CTRL`+`-` `CTRL`+`+`, so can manually get it when I want, but, ofc breaking F11 breaks that manual change of zoom level.

bcdrme:
Closing this one in favor of #329 
Thank you for bringing up this issue :pray:

PR #328, opened 2025-03-19
Add code for pre-downloading data from pool cdn

### LINKED ISSUE #309: Download initial game version from CDN
The task is to extend `src/renderer/components/misc/InitialSetup.vue` with additional step: detect if there exists a `pool` directory inside of the install path and if it doesn't, before calling `downloadGame`, first fetch and unpack https://pool-init.beyondallreason.dev/data.7z there.

This is to save the bandwidth cost of the initial game install from one CDN, to a different one.

The file is large, so it's recommended to use library like `node-downloader-helper`. If the download fails: that's fine, the lobby should ignore and fallback to the regular `pr-downloader` download.

wereii:
Fixes #309

### DIFF
--- a/src/main/content/game/pool-cdn.ts
+++ b/src/main/content/game/pool-cdn.ts
@@ -1,0 +1,103 @@
+import { CONTENT_PATH } from "@main/config/app";
+import * as fs from "fs";
+import path from "path";
+import { DownloaderHelper } from "node-downloader-helper";
+import { logger } from "@main/utils/logger";
+
+import type { DownloadInfo } from "@main/content/downloads";
+import { Downloader } from "@main/content/abstract-content";
+import { removeFromArray } from "$/jaz-ts-utils/object";
+import { GameContentAPI } from "@main/content/game/game-content";
+import { extract7z } from "@main/utils/extract-7z";
+
+const log = logger("pool-cdn.ts");
+
+export class PoolCdnDownloader extends Downloader {
+    cdnUrl = "https://pool-init.beyondallreason.dev";
+    poolDataUrl = `${this.cdnUrl}/data.7z`;
+
+    constructor(gameApi: GameContentAPI) {
+        super();
+        this.onDownloadStart.add((downloadInfo) => {
+            gameApi.onDownloadStart.dispatch(downloadInfo);
+        });
+        this.onDownloadComplete.add((downloadInfo) => {
+            gameApi.onDownloadComplete.dispatch(downloadInfo);
+        });
+        this.onDownloadProgress.add((downloadInfo) => {
+            gameApi.onDownloadProgress.dispatch(downloadInfo);
+        });
+        this.onDownloadFail.add((downloadInfo) => {
+            gameApi.onDownloadFail.dispatch(downloadInfo);
+        });
+    }
+
+    /**
+     * Download and extract pool data from the pool CDN, if pool directory does not exist in content folder.
+     *
+     * Will try to reuse existing download if it exists (DownloadHelper will resume download).
+     */
+    public async preloadPoolData() {
+        const poolDirPath = path.join(CONTENT_PATH, "pool");
+        try {
+            await fs.promises.stat(poolDirPath);
+            log.debug("Pool folder exists, skipping download");
+            return;
+        } catch (e) {
+            if ((e as NodeJS.ErrnoException).code !== "ENOENT") {
+                console.error("Error checking pool folder", e);
+                throw e;
+            }
+            // Pool folder does not exist, create and continue
+            await fs.promises.mkdir(poolDirPath);
+        }
+        log.info("Pool folder does not exist, downloading pool data");
+
+        const downloadInfo: DownloadInfo = {
+            type: "game",
+            name: "pool-data",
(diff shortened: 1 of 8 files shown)


bcdrme (approved):
LGTM.

src/main/content/game/pool-cdn.ts
```diff
@@ -0,0 +1,103 @@
...
+            log.info("Pool data download started");
+            this.downloadStarted(downloadInfo);
+        });
+        dl.on("end", () => {
+            log.info("Pool data download complete");
+            this.downloadComplete(downloadInfo);
+        });
+        dl.on("error", (error) => {
+            console.error("Pool data download failed", error);
+            this.downloadFailed(downloadInfo);
```
p2004a, line 83: This doesn't fully fulfill the

> If the download fails: that's fine, the lobby should ignore and fallback to the regular pr-downloader download.

from the issue

p2004a: It's not a big thing, fine to keep it as is, until people start to complain.

src/main/content/game/pool-cdn.ts
```diff
@@ -0,0 +1,103 @@
...
+        });
+
+        try {
+            await dl.start();
+        } finally {
+            removeFromArray(this.currentDownloads, downloadInfo);
+        }
+
+        await extract7z(dlFilePath, poolDirPath);
+        log.info("Pool data extracted");
```
p2004a, line 101: The downloaded `data.7z` file needs to be removed, it's missing here.

src/main/content/game/pool-cdn.ts
wereii: Is there a special handler that "stops" the whole lobby in case something calls `donwloadFailed` ?
I've probably not noticed it then, as my understanding was that it would flash the error for a bit but because no error is propagated out and nothing is stopped, the usual process of initial-setup (pr-downloader) continues anyway.

p2004a: Oh, ok, you are right. I've misread it, I've tested now and works as intended, sorry!!!

> Edit: Does this also mean, pool cdn download should have replaced the pr-downloader step for maps? And only fallback to pr-d if pool fails?

I don't follow this question, pool is only for game files, it has nothing to do with maps.

src/main/content/game/pool-cdn.ts
```diff
@@ -0,0 +1,103 @@
...
+        this.currentDownloads.push(downloadInfo);
+
+        const dlFilePath = path.join(CONTENT_PATH, "data.7z");
+        const dl = new DownloaderHelper(this.poolDataUrl, CONTENT_PATH, {
+            fileName: "data.7z",
+            override: true,
+            resumeIfFileExists: true,
+            resumeOnIncomplete: true,
+            removeOnStop: false,
+            removeOnFail: false,
```
p2004a, line 71: Okey, while testing failures, I've notices that no timeouts etc are set here. So, if I go offline, but just disabling my wifi connection, it never fails, it also doesn't fail when I reconnect.

I would recommend taking a look at config in spring-launcher: https://github.com/beyond-all-reason/spring-launcher/blob/2c0a6e17b50ab04ea592c53963c571e5ccb071ba/src/http_downloader.js#L60-L66

src/main/content/game/pool-cdn.ts
wereii: > I don't follow this question, pool is only for game files, it has nothing to do with maps.

Typo, I meant the game files, and no worries, more eyes means better code:)

src/main/content/game/pool-cdn.ts
wereii: I forgot to test timeouts/disconnects - will look into it, will come with the data.7z deletion PR.

Should I also set User-Agent to let's say "bar-lobby" while I am at it ?

src/main/content/game/pool-cdn.ts
p2004a: Ah, then the answer is no, pr-downloader must be always called. `pool` directory is only "raw content", updated only once a week, and pr-downloader still need to download "index files" that point into this content and fetch the small number of missing files.

src/main/content/game/pool-cdn.ts
p2004a: > Should I also set User-Agent to let's say "bar-lobby" while I am at it ?

This one *shouldn't* matter, unless that package starts to be used for some abuse and e.g. Cloudflare will block [default user agent](<https://github.com/hgouveia/node-downloader-helper/blob/e5cc3a5884fb4f90f86b2dae1b918450b6d24d42/example/index.js#L26>) like they do with python urllib. So, yeah, you can set it to bar-lobby, but we aren't consistently setting it across the whole codebase, all http invocations anyway ;).

wereii: data.7z deletion fix in #336 - I didn't add the user-agent, believe that should be done properly (global constant and changed everywhere) when really needed

wereii: dammit, I forgot to include the timeout changes

PR #329, opened 2025-03-20
Rework scaling and add window size settings

bcdrme:
Should be better like this hopefully @p2004a 

- rework the scaling calculation to always use the same baseline
- add window size settings
- disable zoom/unzoom key bindings

image

Related #327

### DIFF
--- a/src/main/json/model/settings.ts
+++ b/src/main/json/model/settings.ts
@@ -2,6 +2,7 @@ import { Type } from "@sinclair/typebox";
 
 export const settingsSchema = Type.Object({
     fullscreen: Type.Boolean({ default: true }),
+    size: Type.Number({ default: 900 }),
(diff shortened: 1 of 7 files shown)


p2004a (changes requested):
in this PR
```
app.commandLine.appendSwitch("high-dpi-support", "1");
app.commandLine.appendSwitch("force-device-scale-factor", "1");
```
are still set, so it's not going to be fine.

image

But, when I drop those lines, all works as expected, at least I can't spot anything broken.

p2004a:
Also, it stays as expected if I artificially bump up only *font* scaling (not the base display scaling, so from native display resolution, ui scale is 2x, font scale is 3x)

image

bcdrme:
I ended up removing those two lines, as per our discussion. I allowed resizing and zoom in dev mode for convenience. It should work as expected now.

PR #330, opened 2025-03-22
Improve the logging system

### LINKED ISSUE #311: Allow user to upload logs
This is a basic functionality to help debug the issues with launcher and engine.

There is plenty of possible implementations but we can start with the one that exists already in spring-launcher as there is already all the infra for it:

1. Save logs from client and engine invocations in dated files
2. When user wants to upload logs package together last N (e.g. we use 7) log files into zip
3. Upload to https://log.beyondallreason.dev

See implementation in launcher for details: https://github.com/beyond-all-reason/spring-launcher/blob/master/src/log_uploader.js

JasonReed-2:
Fixes #311 by improving the logging system.

Features:
- Writes logs to dated files.
- Purges outdated logs automatically.
- Archives and upload logs on user request.
- Both main and renderer processes write to the same log file.

3/24 Additions:
- Users can upload logs either via the "Settings" menu or when an error occurs.
- Users clicking the button to send logs copies the URL to their clipboard (the UI for this does not look great but it works for now).
- Actually purges outdated [more than 14 logs ago] log files [no longer only when uploaded].


src/renderer/components/misc/DebugSidebar.vue
```diff
@@ -99,6 +100,13 @@ async function openStartScript() {
 function openSyncLobbyContentTool() {
     syncLobbyContentToolOpen.value = true;
 }
+
+function causeError() {
+    const x = null;
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    const j = x.hello;
```
bcdrme, line 108: It's still getting picked up by the linter apparently.

bcdrme (changes requested):
There is just that linter issue for the causeError function. Other than that it looks solid.

bcdrme (approved):

src/main/services/log.service.ts
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    return remainingFiles;
+}
+
+export async function packLogFiles() {
+    // Send only the 7 most recent log files.
+    const filesToPack = await purgeLogFiles();
+
+    // Copy the current log file into it's own file to upload to avoid lock issue.
```
p2004a, line 65: What lock issue are you referring to here?

src/main/services/log.service.ts
```diff
@@ -0,0 +1,152 @@
...
+    const archivePath = await packLogFiles();
+    if (archivePath == null) {
+        return;
+    }
+
+    const archiveFile = path.basename(archivePath);
+
+    axios({
+        method: "put",
+        url: "https://log.beyondallreason.dev/" + archiveFile,
```
p2004a, line 105: nit: Would be cool if you already put the url in the configuration file as constant

src/main/services/log.service.ts
```diff
@@ -0,0 +1,152 @@
...
+        headers: {
+            "content-type": "application/zip",
+        },
+        data: readFileSync(archivePath),
+        validateStatus: function (status) {
```
p2004a, line 110: I believe default validate status should be sufficient.

src/main/services/log.service.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    const archiveFile = path.basename(archivePath);
+
+    axios({
+        method: "put",
+        url: "https://log.beyondallreason.dev/" + archiveFile,
+        headers: {
+            "content-type": "application/zip",
+        },
+        data: readFileSync(archivePath),
```
p2004a, line 109: any reason the function is async and you are reading file in async mode?

src/main/services/log.service.ts
```diff
@@ -0,0 +1,152 @@
...
+}
+
+export async function purgeLogFiles() {
+    const fileList = await getSortedLogFiles();
+
+    // Keep only the 7 most recent log files.
+    const filesToDelete = fileList.slice(7);
+    const remainingFiles = fileList.slice(0, 7);
+
+    for await (const logFilePath of filesToDelete) {
```
p2004a, line 48: Why prune older logs?

src/main/services/log.service.ts
```diff
@@ -0,0 +1,152 @@
...
+        await rm(logFilePath);
+    }
+
+    // Delete all zip files
+    for await (const entry of glob("*.zip", { cwd: LOGS_PATH })) {
```
p2004a, line 53: I think removal of the zip, could be triggered just after the upload in addition to here, wdyt?

p2004a (changes requested):
- The uploaded log url is not presented to used in any way so user can't share it with anybody. This is required for functionality to be useful.
- It doesn't fully resolve the issue, because issue also asks for engine logs upload. It can be done in separate PR, but it just means this PR doesn't fully resolve the issue.
- It should be possible to just trigger logs upload, not only in case there is the big red error case (maybe option in settings?). I'm fine if that is separate github issue and PR

p2004a:
> It doesn't fully resolve the issue, because issue also asks for engine logs upload. It can be done in separate PR, but it just means this PR doesn't fully resolve the issue.

I take this one back, lobby contains the engine log inline, so that is fine :).

I'm not sure if the https://imolarpg.dyndns.org/bar/stacktrace_translator/frontend/ will work correctly in case of crashes though, that would need to be tested. That's the only "format" restriction I forgot to mentions, it it might be useful to capture engine logs separately, but, we can keep that are a new separate issue.

src/main/services/log.service.ts
JasonReed-2: If the process tries to archive the current log file: you get an error due to "another process is using the file" (assuming this is in Pino) and the current log file does not get included the archive file.

So this is to take a copy of the current log file and archive that instead.

src/main/services/log.service.ts
JasonReed-2: No reason other than I saw this done someplace else in the code base. But I can update this to use the async read file instead.

src/main/services/log.service.ts
JasonReed-2: Maybe this is not a practical concern. But I was worried about disk space usage over time if we just kept creating new log files every time the application was launched.

JasonReed-2:
Thanks for this feedback! I'll address these in this PR.

src/main/services/log.service.ts
p2004a: Hmm, ok, interesting, I wonder how it's possible we don't run into it [in launcher](<https://github.com/beyond-all-reason/spring-launcher/blob/master/src/log_uploader.js>).

src/main/services/log.service.ts
p2004a: I think just last 7 files is too little to keep on disk, just in case. But I see at the moment that it's only called when uploading log, so it doesn't really prevent the buildup (unless I miss some call?)

src/main/services/log.service.ts
p2004a: I think just last 7 files is too little to keep on disk, just in case. But I see at the moment that it's only called when uploading log, so it doesn't really prevent the buildup (unless I miss some call?)

src/main/services/log.service.ts
JasonReed-2: Agree. I added this in.

src/main/services/log.service.ts
JasonReed-2: I upped the # of saved logs to 14 but still only upload the 7 most recent.

I also added the automatic purge to application launch.

JasonReed-2:
I addressed the requested changes and edited my original PR comment to reflect the new changes.

The only one I did not address is the https://imolarpg.dyndns.org/bar/stacktrace_translator/frontend/ need. That one probably requires more research and is out of the scope for my desired changes in this PR. Can that get documented as a separate issue?

src/renderer/components/misc/Error.vue
```diff
@@ -35,12 +52,31 @@ window.addEventListener("error", (event) => {
...
+    // Store off event and target to avoid async issue
+    const curE = event;
+    const curTarget = curE.currentTarget;
+
+    // Do the operations
+    try {
+        const url = await uploadLogs();
+        await navigator.clipboard.writeText(url);
+        tooltipMessage.value = "Log URL was copied to clipboard.";
+    } catch (_) {
```
p2004a, line 73: nit: you could just `} catch {`, it will also remove warning. Same in [src/renderer/components/navbar/Settings.vue](https://github.com/beyond-all-reason/bar-lobby/pull/330/files#diff-b87832973649a2b045194ffa49bdc5f94f7f2dd5be9fff8f57121850348ebed4)

p2004a (commented):
Just this one issue with types and it's ready to merge :)

p2004a (approved):

Issue #202, opened 2023-04-16
Hovering on the edge of controls spams their hover event and sound


Jazcash:
[ice_video_20230416-172050.webm](https://user-images.githubusercontent.com/1434248/232326315-c21909a2-6a0f-4d13-804b-0a7dbd25fff7.webm)

This doesn't seem to be consistent, even with the same controls. Could possibly be related to the tooltip but not sure.

p2004a:
obsolete

p2004a:
Or maybe, first

@bcdrme is this still an issue you see in current version?

bcdrme:
Yes it is still around, I could reproduce

gdoteof:
@bcdrme if you have the opportunity, i would appreciate it if you can see if #529 fixes it for you

PR #338, opened 2025-03-27
fixup of #202: Hovering on the edge of controls spams their hover event and sound 

CamelCaseFan:
- added border to the tooltip so that mouse hover won't trigger event spam anymore 
- tooltips more robust by using a debounce mechanism

### DIFF
--- a/src/renderer/styles/_tooltip.scss
+++ b/src/renderer/styles/_tooltip.scss
@@ -3,6 +3,9 @@ $color: #222;
 .p-tooltip {
     z-index: 10;
     max-width: 400px;
+    transform: translateY(5px);
+    border-radius: 20px;
+    border: 1px solid rgba(255, 255, 255, 0.08);
     &-text {
         background: $color;
         padding: 2px 7px;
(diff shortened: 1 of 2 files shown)


src/renderer/styles/_tooltip.scss
```diff
@@ -3,6 +3,9 @@ $color: #222;
 .p-tooltip {
     z-index: 10;
     max-width: 400px;
+    transform: translateY(5px);
+    border-radius: 20px;
```
bcdrme, line 7: Why did you add a border-radius ? 
Looks super weird

image

src/renderer/styles/_tooltip.scss
```diff
@@ -3,6 +3,9 @@ $color: #222;
 .p-tooltip {
     z-index: 10;
     max-width: 400px;
+    transform: translateY(5px);
```
bcdrme, line 6: Can you explain this change ? This looks unnecessary.

bcdrme (changes requested):
This was supposed to fix the event spamming but I can reproduce the issue with your changes. 
The CSS changes you've made are very questionable, check comments.

PR #339, opened 2025-03-29
Fix replay cache issue

### LINKED ISSUE #332: Replay parsing triggered on new-in progress game.
New replay file is created as soon as engine starts up, it's empty, and it's filled up as the game progresses. Looks like currently lobby tries to parse replay as soon as the engine starts up which results in error and the new replay file is deleted! Both of those things are unintended, moreover, I don't think we should ever just remove demo files if we fail to parse them.

<details>
<summary>Log</summary>

```
[19:04:53.854] DEBUG: replay-content.ts - Chokidar -=- Replay added: /home/p2004a/.local/share/Beyond All Reason/demos/2025-03-23_18-04-51-841_Wanderlust 2_2025.01.6.sdfz
[19:04:53.855] DEBUG: replay-content.ts - Caching: /home/p2004a/.local/share/Beyond All Reason/demos/2025-03-23_18-04-51-841_Wanderlust 2_2025.01.6.sdfz
[19:04:53.859] ERROR: replay-content.ts - Error caching replay: /home/p2004a/.local/share/Beyond All Reason/demos/2025-03-23_18-04-51-841_Wanderlust 2_2025.01.6.sdfz
[19:04:53.859] ERROR: replay-content.ts - unexpected end of file
    err: {
      "type": "Error",
      "message": "unexpected end of file",
      "stack":
          Error: unexpected end of file
              at genericNodeError (node:internal/errors:983:15)
              at wrappedFn (node:internal/errors:537:14)
              at Zlib.zlibOnError [as onerror] (node:zlib:185:17)
    }
Error: unexpected end of file
    at genericNodeError (node:internal/errors:983:15)
    at wrappedFn (node:internal/errors:537:14)
    at Zlib.zlibOnError [as onerror] (node:zlib:185:17) {
  errno: -5,
  code: 'Z_BUF_ERROR'
}
[19:04:53.961] DEBUG: replay-content.ts - Chokidar -=- Replay removed: /home/p2004a/.local/share/Beyond All Reason/demos/2025-03-23_18-04-51-841_Wanderlust 2_2025.01.6.sdfz
```

</details>

JasonReed-2:
Fixes #332 by queuing the replay files to be cached only after the game closes.

### DIFF
--- a/src/main/content/replays/replay-content.ts
+++ b/src/main/content/replays/replay-content.ts
@@ -67,7 +67,7 @@ export class ReplayContentAPI {
         existingFiles
             .filter((fileName) => !replayFileNames.includes(fileName))
             .map((fileName) => path.join(REPLAYS_PATH, fileName))
-            .forEach(this.cacheReplay);
+            .forEach((fileName) => this.cacheReplay(fileName));
     }
 
     protected async scanFolderForReplays() {
@@ -84,7 +84,7 @@ export class ReplayContentAPI {
         }
     }
 
-    cacheReplay = async (replayFilePath: string) => {
+    protected async cacheReplay(replayFilePath: string) {
         if (gameAPI.isGameRunning()) {
             log.debug(`Queuing replay to be cached: ${replayFilePath}`);
             this.replayCacheQueue.add(replayFilePath);
@@ -115,7 +115,7 @@ export class ReplayContentAPI {
             log.error(err);
             //TODO emit error signal
         }
-    };
+    }
 
     public async cacheReplaysInQueue() {
         for (const replayFilePath of this.replayCacheQueue) {
--- a/src/main/content/replays/replay-content.ts
+++ b/src/main/content/replays/replay-content.ts
@@ -7,7 +7,7 @@ import { gameAPI } from "@main/game/game";
 import { logger } from "@main/utils/logger";
 import chokidar from "chokidar";
 import fs from "fs";
-import fsPromises from "fs/promises"
+import fsPromises from "fs/promises";
 
 import path from "path";
 
--- a/src/main/content/replays/replay-content.ts
+++ b/src/main/content/replays/replay-content.ts
@@ -120,6 +120,7 @@ export class ReplayContentAPI {
     public async cacheReplaysInQueue() {
         for (const replayFilePath of this.replayCacheQueue) {
             await this.cacheReplay(replayFilePath);
+            this.replayCacheQueue.delete(replayFilePath);
         }
     }
 }
--- a/src/main/content/replays/replay-content.ts
+++ b/src/main/content/replays/replay-content.ts
@@ -38,7 +38,7 @@ export class ReplayContentAPI {
                     return;
                 }
                 log.debug(`Chokidar -=- Replay added: ${filepath}`);
-                this.replayCacheQueue.add(filepath);
+                this.cacheReplay(filepath);
             })
             .on("unlink", (filepath) => {
                 if (!filepath.endsWith("sdfz")) {
@@ -85,6 +85,12 @@ export class ReplayContentAPI {
     }
 
     cacheReplay = async (replayFilePath: string) => {
+        if (gameAPI.isGameRunning()) {
+            log.debug(`Queuing replay to be cached: ${replayFilePath}`);
+            this.replayCacheQueue.add(replayFilePath);
+            return;
+        }
+
         log.debug(`Caching: ${replayFilePath}`);
         try {
             const replayData = await asyncParseReplay(replayFilePath);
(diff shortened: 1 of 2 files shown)


src/main/content/replays/replay-content.ts
```diff
@@ -84,7 +86,11 @@ export class ReplayContentAPI {
         }
     }
 
-    protected async cacheReplay(replayFilePath: string) {
+    protected async cacheReplay(replayFilePath: string, obj: ReplayContentAPI) {
```
p2004a, line 89: I don't understand why you are passing `obj`, the arrow function captures `this` from outer scope

[Quoting MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this):

> [Arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) differ in their handling of this: they inherit this from the parent scope at the time they are defined. This behavior makes arrow functions particularly useful for callbacks and preserving context

inside of `this.cacheReplay(filePath, this)` invocation `obj == this`.

src/main/content/replays/replay-content.ts
```diff
@@ -84,7 +86,11 @@ export class ReplayContentAPI {
...
+        // Don't try to cache an empty replay file
+        const stat = await fsPromises.stat(replayFilePath);
+        if (stat.size === 0) return;
```
p2004a, line 92: We skip file size if it's 0:
- Will this file caching be re-triggered once the engine finishes running?
- There is still race condition here of engine writing a few bytes but still not full header and this won't be parsable

p2004a (commented):
One more thing I highlight in that issue that lobby shouldn't really remove those under any circumstance and leave that to the user, do you have a good reason to keep the deletion?

src/main/content/replays/replay-content.ts
JasonReed-2: Right but cacheReplay is not an arrow function. Now that I think on this more: making that change was probably the right decision. I'll update that.

bcdrme:
I doesn't crash but it doesn't parse the replay either. I had to restart the app (or hit F5) to get the replay parsing running for a game I started and stopped.

src/main/content/replays/replay-content.ts
p2004a: All calls to `cacheReplay` are via `this.cacheReplay`, the `this` inside of the method should be properly bound, so `obj == this`.

JasonReed-2: Gotcha. I was encountering an error with "this" being undefined when I was testing. That is what led me to this change. I'll set it back and give it another test.

JasonReed-2:
I can update to remove the rename action to avoid "deleting" the replay file if it couldn't be cached.

I think I need a better understanding of the desired outcome.

Should the replay be continuously parsed and cached as the .sdfz file is being written to by the engine?
If not: is it sufficient to cache the generated replay files only after the game closes?

bcdrme:
Sure. 

A replay is an archive containing a bunch of files. We want to extract metadata from theses, this is the "caching" part. 
When you start a game and stop it, a new .sdfz file should have been generated and needs parsing. This used to work on detecting a new .sdfz file in the replay dir. 
The same mechanism is used when downloading and dropping a file in the replay dir (chokidar).

Now you could just start a scan of new files to "cache" after a game quit event, that would be a way of doing it. Or ask for a game engine change, so the engine generates a `.sdfz.tmp` until it's complete, then renames to `.sdfz` when the file is writter fully.

p2004a:
IMHO, easiest would be to pause all replay parsing while there is engine instance running, and resume when engine is stopped. When engine is running, nobody should be trying to start some replay anyway.

JasonReed-2:
Thanks for the feedback and suggestions @p2004a & @bcdrme. I made the update to queue the replay files to be cached and cache them after the game closes.

bcdrme (changes requested):
1. Downloading a replay and dropping it in the `demos` folder doesn't trigger a parsing. This is a regression, this is why this folder was watched by chokidar.

2. After watching a replay, the game closes, it scans all the files again. This is more minor, but some people will have hundreds of files in their `demos` folder.

src/main/content/replays/replay-content.ts
```diff
@@ -84,7 +84,13 @@ export class ReplayContentAPI {
         }
     }
 
-    protected async cacheReplay(replayFilePath: string) {
+    cacheReplay = async (replayFilePath: string) => {
```
p2004a, line 87: Why make this change? Why not keep it `protected async cacheReplay` ?

p2004a: The only place where I see usage of this is incorrect is above

```js
.forEach(this.cacheReplay);
```

It needs to be

```js
.forEach((file) => this.cacheReplay(file));
```

instead.

That's a bug in that line, not here.

JasonReed-2: Made an update for this. Thank you!

JasonReed-2:
Sorry for the amount of back and forth on this. Next time: I'll do better to understand the feature and ask clarifying questions before jumping into a solution.

p2004a:
No worries :D! Thanks a lot for this fix!

p2004a (approved):

PR #346, opened 2025-03-29
Update readme to include nvm use

desean1625:


p2004a:
There are many ways node can be installed, there is also e.g. `.mise.toml` in the repo.

Instead of change like this, I would add a paragraph in the Requirements section like:

> We recommend usage of tools management software like [NVM](https://github.com/nvm-sh/nvm) or [mise-en-place](https://mise.jdx.dev/) for installation of correct node version.

p2004a:
Deprecated by https://github.com/beyond-all-reason/bar-lobby/commit/b4cd53b28cae9d607c3fd2904cf537a73e16f01e

Issue #344, opened 2025-03-29
Error page is not top level dialog always


p2004a:
1. Open single player, "Custom 2"
2. Try to add bot, select circuit, it crashes
3. The error page with reload button is under bot selection dialog

![Image](https://github.com/user-attachments/assets/4352b2b9-883e-4ac1-be0b-ecedd4bdf166)

PR #347, opened 2025-03-30
Fix errors getting overlaid by other modals

wereii:
My solution to #344

The problem is that Error is modal too, so they all teleport to same space, specifically get appended to \#wrapper - I don't know why in this specific case the bot selection modal ends up lower in the tree (error should be appended after the click right?) then the Error modal but it does.  
I am of the opinion that Errors should be special modals with their own dedicated place that ensures it renders over anything else, see the solution.

Though I am not that experienced with Vue and frontend work so there is probably a better solution to this. Modals should not care about what they are displaying so the isError is bad, maybe noTeleport or something like that would be better.

### DIFF
--- a/src/renderer/App.vue
+++ b/src/renderer/App.vue
@@ -43,9 +43,11 @@
             </div>
         </Transition>
         <Settings v-model="settingsOpen" />
-        <Error />
         <ChatComponent v-if="state === 'default' && me.isAuthenticated && tachyonStore.isConnected" />
         <FullscreenGameModeSelector v-if="state === 'default'" :visible="battleStore.isSelectingGameMode" />
+    </div>
+    <div id="error-container">
+        <Error />
     </div>
 </template>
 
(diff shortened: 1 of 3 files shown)


wereii:
Note, OverlayPanel does have appendTo prop and the way I understood it was that that element would teleport itself out to body by default but it either does not work, or does something else I don't understand. The documentation for that element ~~sucks~~ explains nothing.

wereii:
Had a quick look into code of OverlayPanel - it does have Portal (which is primevue custom Teleport component) so errors should by default teleport to body but if I remember the tree correctly it doesn't, the whole element tree stays in \#wrapper, I will look into why later.

p2004a (commented):
We are sorry this PR is not reviewed yet :(.

I took a look but I'm not really experienced with vue/rusty with frontend, so some dumb questions:
- Instead of making isError special, maybe it should be a more generic "teleport" property or something like that?
- Does it need to be handled like that or instead being able to override z-index is maybe a better idea...?

bcdrme:
Sorry I haven't been very reactive theses days. 

Yeah I think the approach in the PR isn't the best. As you mentioned the whole teleport thing seems a bit weird. One could write a standalone Error component that always sits on top of everything else and that doesn't necessarily re-use the existing Modal code.

p2004a:
Deprecated by #404 where I've rewritten the `Error` component to not depend on `Modal`.

PR #359, opened 2025-04-04
Display errors that caused a disconnect

desean1625:
Before
image

after

image

### DIFF
--- a/src/main/tachyon/tachyon-client.ts
+++ b/src/main/tachyon/tachyon-client.ts
@@ -61,7 +61,6 @@ export class TachyonClient {
                         reject(details.error_description);
                         return;
                     }
-                    console.log(res, chunk.toString());
                     const error = `HTTP Error ${res.statusCode}: ${chunk.toString()}`;
                     log.error(error);
                     reject(error);
--- a/src/main/tachyon/tachyon-client.ts
+++ b/src/main/tachyon/tachyon-client.ts
@@ -16,6 +16,11 @@ export type TachyonClientRequestHandlers = {
     ) => Promise<Omit<GetCommands<"user", "server", "response", CommandId>, "type" | "commandId" | "messageId">>;
 };
 
+type TachyonError = {
+    error: string;
+    error_description: string;
+};
+
 export class TachyonClient {
     public socket?: WebSocket;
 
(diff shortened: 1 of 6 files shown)


desean1625:
fixes !358

src/main/tachyon/tachyon-client.ts
```diff
@@ -16,6 +16,11 @@ export type TachyonClientRequestHandlers = {
     ) => Promise<Omit<GetCommands<"user", "server", "response", CommandId>, "type" | "commandId" | "messageId">>;
 };
 
+type TachyonError = {
+    error: string;
+    error_description: string;
+};
```
p2004a, line 22: This is not defined anywhere in the specification https://github.com/beyond-all-reason/tachyon/blob/master/docs/connection.md, I don't like depending on this not agreed on teiserver implementation detail. So, I would expect it to be first specified in protocol.

`TachyonError` is also too generic, it sounds like internal protocol error, when it's just initial websocket connection error.

src/main/tachyon/tachyon-client.ts
```diff
@@ -45,6 +50,17 @@ export class TachyonClient {
             });
             this.socket.on("unexpected-response", async (req, res) => {
                 res.on("data", (chunk: Buffer) => {
+                    let details: TachyonError | undefined;
+                    try {
+                        details = JSON.parse(chunk.toString());
+                    } catch {
+                        //eat json parsing errors
+                    }
```
p2004a, line 58: This parsing is buggy because it reads only a single chunk from response, not whole response. It was maybe sub-optimal but a bit acceptable for stringification below, but when parsing, not really.

src/renderer/components/navbar/NavBar.vue
```diff
@@ -107,7 +107,7 @@ const router = useRouter();
 const allRoutes = router.getRoutes();
 const primaryRoutes = computed(() => {
     return allRoutes
-        .filter((r) => ["/singleplayer", "/multiplayer", "/library", "/learn", "/store", "/development"].includes(r.path))
+        .filter((r) => ["/singleplayer", "/library", "/multiplayer", "/learn", "/store", "/development"].includes(r.path))
```
p2004a, line 110: let's not mix in unrelated changes

src/renderer/store/tachyon.store.ts
```diff
@@ -23,7 +23,11 @@ async function connect() {
         tachyonStore.error = undefined;
     } catch (err) {
         console.error("Failed to connect to Tachyon server", err);
-        tachyonStore.error = "Error";
+        if (err instanceof Error && err.message.split(": ")[1]) {
```
p2004a, line 26: This is nasty parsing. Proper way is to subtype `Error` and pass in custom properties.

p2004a (changes requested):
We are sorry for dalay in reviewing this PR

Hectate:
Closing as stale.

PR #362, opened 2025-04-09
Default engine version logic

### LINKED ISSUE #349: The "selected engine" logic is broken
Single player uses this idea of "selected engine" to start local session, but the selection logic doesn't make sense:

1. The version references from https://github.com/beyond-all-reason/bar-lobby/blob/bc0163cae0fe86c37224de58ebed7ac9aa7da12a/src/main/config/default-versions.ts#L11 is only ensured that it's installed, it's not used as default for single player
2. The "selected engine" is picked on start as last element of available engines (sorted list?) https://github.com/beyond-all-reason/bar-lobby/blob/bc0163cae0fe86c37224de58ebed7ac9aa7da12a/src/renderer/components/misc/InitialSetup.vue#L29 which is not the default engine that should be used by single player
3. The download engine handler sets the selected version https://github.com/beyond-all-reason/bar-lobby/blob/bc0163cae0fe86c37224de58ebed7ac9aa7da12a/src/renderer/store/engine.store.ts#L30-L34 that means that if:
    - You download old replay
    - Try to watch it which triggers engine download
    - Next single player run will use that old engine version leading to broken game

The whole "engine selected" logic, what is used as default for single player, and overriding engine version when dev mode is enabled need to be entangled from each other.

JasonReed-2:
Fixes #349.

Refactors the engine store concept and enforces the "default engine version" [the hard-coded engine version]. Also: downloading an engine does not set the selected engine version to that downloaded engine version.

### DIFF
--- a/src/main/config/default-versions.ts
+++ b/src/main/config/default-versions.ts
@@ -8,4 +8,4 @@
 
 export const LATEST = "LATEST";
 export const LATEST_GAME_VERSION = "byar:test";
-export const LATEST_ENGINE_VERSION = "2025.01.6";
+export const DEFAULT_ENGINE_VERSION = "2025.01.6";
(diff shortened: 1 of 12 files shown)


p2004a (approved):
Looks good, thank you :smile:!

PR #365, opened 2025-04-21
[FIX-351] watch replay - prevent multiple launches

### LINKED ISSUE #351: Clicking "watch" for replay multiple times, will start downloads and games mutliple times
Just spam the "watch" button in replay, and it will trigger multiple launches in parallel. Quite easy to do by user because #350 is not implemented, maybe this issue should be just tackled as part of #350 too, but I guess disabling the watch button without propagating all the watch progress would be a good start.

maciejbaba:
closes #351

### DIFF
--- a/src/renderer/views/library/replays.vue
+++ b/src/renderer/views/library/replays.vue
@@ -72,8 +72,12 @@
                     <Panel class="flex-grow">
                         <ReplayPreview v-if="selectedReplay" :replay="selectedReplay" :showSpoilers="showSpoilers">
                             <template #actions="{ replay }">
-                                <DownloadContentButton v-if="map && replay" :map="map" @click="watchReplay(replay)"
-                                    >Watch</DownloadContentButton
+                                <DownloadContentButton
+                                    v-if="map && replay"
+                                    :map="map"
+                                    @click="watchReplay(replay)"
+                                    :disabled="isLaunching || gameStore.isGameRunning"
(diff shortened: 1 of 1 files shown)


src/renderer/views/library/replays.vue
```diff
@@ -72,8 +72,12 @@
...
                         <ReplayPreview v-if="selectedReplay" :replay="selectedReplay" :showSpoilers="showSpoilers">
                             <template #actions="{ replay }">
-                                <DownloadContentButton v-if="map && replay" :map="map" @click="watchReplay(replay)"
-                                    >Watch</DownloadContentButton
+                                <DownloadContentButton
+                                    v-if="map && replay"
+                                    :map="map"
+                                    @click="watchReplay(replay)"
+                                    :disabled="isLaunching || gameStore.isGameRunning"
+                                    >{{ getWatchButtonText() }}</DownloadContentButton
```
p2004a, line 80: Not a vue dev, but I see in other parts of the code it's handled as

```vue
<template v-if="gameStore.isGameRunning">Game Running</template>
<template v-else-if="isLaunching.value">Launching...</template>
<template v-else>Watch</template>
```

I guess to have presentation in single place.

maciejbaba: yup, probably a react habit from my side plus i looked at getFriendlyDuration which has a bit different use than this, i will change that

PR #367, opened 2025-04-27
Reorganize Installation and File Paths

### LINKED ISSUE #310: Correct installation paths on Windows and Linux
We have installer, but we also need to verify that the installation paths for game files under both Linux and Windows are appropriate by following the recommendations.

Electron has a very basic https://www.electronjs.org/docs/latest/api/app#appgetpathname available, but it might not be sufficient.

### Mixed data concern

In the old setup, all game files are mixed up together in a single directory: base game files, configuration, "state" files like replays. That's not necessarily correct. E.g. when uninstalling it's common that base game files are expected to be removed, but not e.g. replays https://github.com/beyond-all-reason/spring-launcher/issues/48.

The Engine offers mechanism of layering folders with data, using `SPRING_DATADIR` env variable and `SpringData` springsettings option (See [`DataDirLocater.cpp`](https://github.com/beyond-all-reason/spring/blob/0e02c56c8dc282deed3d4706bb6d392ba437e0d6/rts/System/FileSystem/DataDirLocater.cpp)) so it's possible to separate base game files, and user content into separate directories.

### Documentation

Windows:

- https://learn.microsoft.com/en-us/windows/deployment/usmt/usmt-recognized-environment-variables?utm_source=chatgpt.com
- https://learn.microsoft.com/en-us/windows/win32/shell/csidl

Linux:

- https://specifications.freedesktop.org/basedir-spec/latest/

JasonReed-2:
Fixes #310

Everything is user-specific now [per-machine = false]. This also means that the SDFZ file association will no longer work.

Solves the mixed data concerns by separating out the files as follows:
- State Files (Preserve on uninstall; stored in application data folders)
  - Replays
  - Logs
  - Configuration
- Data Files (Preserve on update via avoiding deleting these files in the customRemoveFiles macro in the NSIS installer script; remove on uninstall)
  - Engine
  - Maps
  - Packages
  - Pool
  - etc. basically everything else that lobby downloads that isn't classified as a "State File"
- Package Files (Remove on update and uninstall)
  - Resources [contains app.asar]
  - Locales

@p2004a : I have not tested this on Linux. Are you able to do so? If not: I can install WSL2 and give it a go there.

Arguably: the replays should live in a users' documents folder instead of the app data folders for easier user accessibility. I haven't implemented that but let me know your thoughts and I can make that change.


src/main/config/app.ts
```diff
@@ -2,15 +2,24 @@ import { contentSources } from "@main/config/content-sources";
 import envPaths from "env-paths";
 import path from "path";
 import { env } from "process";
+import { app } from "electron";
 
 export const APP_NAME = "Beyond All Reason";
 
 const paths = envPaths(APP_NAME, { suffix: env.APP_NAME_SUFFIX || "" });
-export const CONTENT_PATH = paths.data;
+const INSTALL_PATH = path.parse(app.getPath("exe")).dir;
```
bcdrme, line 10: I had to change this line back to `const INSTALL_PATH = paths.data;` to make it work on linux

bcdrme (changes requested):
This is not working on Linux as-is, check comment below

JasonReed-2:
Ok coming back to this after a hiatus. I setup everything in WSL now so I can test this on both linux and windows :)

I see the problem here, and I think the solution needs to be that linux always treats "Data Files" as "State Files" [i.e. preserve on uninstall].

That is the way that launcher handles this based on my reading of https://github.com/beyond-all-reason/spring-launcher/blob/2c0a6e17b50ab04ea592c53963c571e5ccb071ba/src/write_path.js#L29

p2004a:
Half of this works, half doesn't.

- The whole separation of state and data doesn't work, because engine isn't configured properly to read from data and save to state, the replays will still be jut saved in data so this breaks that.
- There is still mixing of state and data on linux and so on.

I'm merging this in current half broken state, and will just follow up with fixes.

PR #370, opened 2025-05-02
Configurable Lobby Server settings added

### LINKED ISSUE #144: An option to choose a different server
It would be useful to have a way to switch to another server, mostly localhost, for testing and development.

Example of other lobbies:
Chobby has this server dropdown with the default server and localhost as a second option.
image

Skylobby has more options, entering any address and port.
image

Hectate:
This PR would close #144 by adding the ability for a user to select the lobby server they intend to authenticate the bar-lobby client to.

The following changes are reflected in this PR.

**User Experience Changes**
- Users with Dev Mode _enabled_ will see an additional button in the Settings modal, labeled "Lobby Server Settings".
- On click, the Lobby Server Settings modal will open. Settings modal will close.

The Lobby Server Settings modal has the following elements:

- Active Server dropdown has a list of servers to select from. They are grouped into two sets, "Default Servers" and "Custom Servers", in that order.
- Upon first open, the "Custom Servers" group will be empty.
- The Default Servers group contains 4 items (see image below).
- Beneath the dropbox field, a "Custom Server" textbox is available. Users can enter a string, with the intention of it being a valid server URL or IP, and optionally port.
- Beneath the textbox input are two buttons. One labeled "Add" and the other "Remove".
- The "Add" button, or pressing Enter while focused on the textbox, will submit the value in the textbox into the dropdown list under the Custom Server group. The textbox will be blanked upon success.
- The "Add" button will not submit if the value is a blank string, or if the value is an exact match to a Default Server item. No other validation is performed on the string.
- The "Remove" button is disabled unless an existing item from the Custom Server group is in the Active Server field.
- Clicking "Remove" with a Custom Server item in the Active Server field will remove that item from the Custom Server group, and set the Active Server to the first item of the Default Server group.

**Application Changes**
The persistent storage for Settings has been modified in the following ways:

- Addition of a `lobbyServer` containing the string of the currently selected Active Server
- Addition of a `customServerList` containing an Array of Strings equal to all strings submitted by the user

`src>main>config>server.ts` has been modified in the following ways:

- Import of the Settings store service
- Usage of the `lobbyServer` value in two new functions
- Replacement of two constants with exported functions to return appropriate strings; `getOAuthAuthorizationServerURL()` and `getOAuthWellKnownURL()` which are equivalent to the original constants, except with the server portion being taken from `lobbyServer`

`src>main>oauth2>oauth2.ts` has been modified in the following way:

- Replaced references to the two constants noted above with the imported functions so authentication will utilize the user's selected server

**Additional Considerations**
The default list of 4 servers is hard-coded into `ServerSettings.vue`. This list was made with assumptions about what would be useful here. If any entries should be added/removed, let me know.

As noted, the servers URLs are just strings with almost no validation. It is my view that this does not represent a security risk to servers or the application. Anyone taking effort to inject code into this string could do it more easily with developer tools. Any deliberately malformed server URLs could only, at worst, degrade the experience of the user attempting this. Servers receiving bar-lobby OAuth requests have their own criteria for valid request/response, and a malformed server URL is not likely to be a risk to the server from an untrusted client.

As noted, changing Active Server has an immediate effect on the OAuth process, meaning the application will immediately expect to be authenticating and refreshing from the currently active server URL. No auto-logout is performed upon switching, which means the refresh token, etc, will cease to match the active server. It is my view that this is not a problem for the following reasons:
1. This feature is most likely to be used by developers, who are assumed to have more resilience to errors caused by changing developer settings on the fly without re-authenticating.
2. A user/developer will likely only infrequently change the active server, and predominantly do so before a login attempt.
3. Forced immediate logout upon changing the active server has a negative side effect for users that change this in error. Currently, if the active server is changed in error, it can be switched back immediately without problems.
4. Until more multi-player/lobby features are added to bar-lobby, it is premature to assume how server connectivity will work in the final product. As such, forcing a logout now has minimal added value to the use of this feature, but could be added if needed in the future.

Finally, this PR does not include any changes to the UI that will identify to the user what server they are connected to, except for in the Lobby Server Setting modal. I believe as more multiplayer features are added that can be addressed on a "server status" element, but it is not important for the use of this feature at this time.

**Images**
image
image
image

### DIFF
--- a/src/main/config/server.ts
+++ b/src/main/config/server.ts
@@ -3,8 +3,10 @@
 // see https://github.com/beyond-all-reason/teiserver/pull/555
 // UPDATE: above issue has been resolved. Leaving note for posterity at the moment.
 
-export const TEISERVER_HOSTNAME_PORT = "server4.beyondallreason.info";
 // export const TEISERVER_HOSTNAME_PORT = "lobby-server-dev.beyondallreason.dev";
+import { settingsService } from "@main/services/settings.service"
+
+const TEISERVER_HOSTNAME_PORT = "server4.beyondallreason.info"
 
 export const OAUTH_AUTHORIZATION_SERVER_URL = `https://${TEISERVER_HOSTNAME_PORT}`;
 export const OAUTH_WELL_KNOWN_URL = `${OAUTH_AUTHORIZATION_SERVER_URL}/.well-known/oauth-authorization-server`;
@@ -12,3 +14,19 @@ export const OAUTH_CLIENT_ID = "generic_lobby";
 export const OAUTH_SCOPE = "tachyon.lobby";
 
 export const WS_SERVER_URL = `wss://${TEISERVER_HOSTNAME_PORT}/tachyon`;
+
+function getLobbyServer() {
+	return settingsService.getSettings().lobbyServer;
+}
+
+function getOAuthAuthorizationServerURL() {
+	const url = (`https://${getLobbyServer()}`);
+	return url
+}
+
+function getOAuthWellKnownURL() {
+	const url = (`${getOAuthAuthorizationServerURL()}/.well-known/oauth-authorization-server`);
+	return url;
+}
+
+export { getOAuthAuthorizationServerURL, getOAuthWellKnownURL }
--- a/src/main/config/server.ts
+++ b/src/main/config/server.ts
@@ -1,32 +1,25 @@
-// TODO: this should reference a real server like server5.beyondallreason.info
-// there is an issue with port handling right now, so we can't use the real server
-// see https://github.com/beyond-all-reason/teiserver/pull/555
-// UPDATE: above issue has been resolved. Leaving note for posterity at the moment.
+import { settingsService } from "@main/services/settings.service";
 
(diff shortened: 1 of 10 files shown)


bcdrme:
Hello there :wave: Nice feature. 

Can you move the new `Lobby Server Settings` button from the settings to the dev panel (this is where we put everything related to dev mode) :pray: 

image

src/main/config/server.ts
```diff
@@ -1,13 +1,32 @@
 // TODO: this should reference a real server like server5.beyondallreason.info
 // there is an issue with port handling right now, so we can't use the real server
 // see https://github.com/beyond-all-reason/teiserver/pull/555
+// UPDATE: above issue has been resolved. Leaving note for posterity at the moment.
```
bcdrme, line 4: You can delete all comments related to resolved issues

bcdrme (changes requested):
Make sure to run theses pre-checks and fix/commit the changes :

`npm run typecheck`
`npm run format`
`npm run lint`

Hectate:
Requested changes made.

I also identified that the websocket server URL was not updated to utilize the selected lobby server, so that has been fixed as well.

The typecheck, format, and lint pre-checks have been run as well now.

bcdrme (approved):

PR #375, opened 2025-05-11
Fixed Scenario launching crash

Hectate:
Render process was passing a `script: string` into IPC, but main was expecting a `settings: ScriptLaunchSettings` instead. This led to crashes when trying to access properties of an undefined object.

The typecheck, format, and lint scripts have been run.


p2004a:
Thank you!

PR #377, opened 2025-05-12
[Draft] Example of using a typed wrapper to reduce errors in IPC calls

jkells:


src/main/services/settings.service.ts
```diff
@@ -41,7 +41,7 @@ function toggleFullscreen() {
 
 function registerIpcHandlers() {
     ipcMain.handle("settings:get", () => getSettings());
-    ipcMain.handle("settings:update", (_, data: Partial<typeof settingsSchema>) => updateSettings(data));
+    ipcMain.handle("settings:update", (_, data: Partial<Settings>) => updateSettings(data));
```
jkells, line 44: Other side used `Settings`

src/main/typed-ipc.ts
```diff
@@ -0,0 +1,125 @@
...
+import { ipcRenderer as electronIpcRenderer, ipcMain as electronIpcMain, IpcMain, IpcMainEvent, IpcMainInvokeEvent, IpcRenderer, IpcRendererEvent, WebContents } from "electron";
+import { logLevels } from "@main/services/log.service";
+import { MapData } from "@main/content/maps/map-data";
+import { MultiplayerLaunchSettings } from "@main/game/game";
+import { NewsFeedData } from "@main/services/news.service";
+import { Replay } from "@main/content/replays/replay";
+import { Scenario } from "@main/content/game/scenario";
+import { Settings } from "@main/services/settings.service";
+import { TachyonEvent } from "tachyon-protocol";
+
```
jkells, line 15: To start with I've just typed these as per the existing handle and listener functions whatever they happened to be.

Note: Some of these are returning un-necessary `| null | undefined`, one even returns `void[]` which is a little odd. We can tidy that up in the future.

src/preload/preload.ts
```diff
@@ -39,11 +37,11 @@ export type MainWindowApi = typeof mainWindowApi;
 contextBridge.exposeInMainWorld("mainWindow", mainWindowApi);
 
 const shellApi = {
-    openConfigDir: (): Promise<void> => ipcRenderer.invoke("shell:openConfigDir"),
-    openContentDir: (): Promise<void> => ipcRenderer.invoke("shell:openContentDir"),
-    openSettingsFile: (): Promise<void> => ipcRenderer.invoke("shell:openSettingsFile"),
-    openStartScript: (): Promise<void> => ipcRenderer.invoke("shell:openStartScript"),
-    openReplaysDir: (): Promise<void> => ipcRenderer.invoke("shell:openReplaysDir"),
+    openConfigDir: (): Promise<string> => ipcRenderer.invoke("shell:openConfigDir"),
```
jkells, line 40: These do actually really return strings which I don't think we use. We could update the handle function to a void return type if we wanted to clean this up.

src/preload/preload.ts
```diff
@@ -93,7 +91,7 @@ contextBridge.exposeInMainWorld("engine", engineApi);
 const gameApi = {
     // Content
     downloadGame: (version: string): Promise<void> => ipcRenderer.invoke("game:downloadGame", version),
-    getScenarios: ((version) => ipcRenderer.invoke("game:getScenarios", version)) as GetScenarios,
+    getScenarios: (version: string) => ipcRenderer.invoke("game:getScenarios", version),
```
jkells, line 94: A few of these had explicit casts like this that I've removed. They weren't all accurate.

src/preload/preload.ts
```diff
@@ -115,45 +113,45 @@ contextBridge.exposeInMainWorld("game", gameApi);
...
-    onDownloadMapProgress: (callback: (downloadInfo: DownloadInfo) => void) => ipcRenderer.on("downloads:map:progress", (_event, downloadInfo) => callback(downloadInfo as DownloadInfo)),
-    onDownloadMapFail: (callback: (downloadInfo: DownloadInfo) => void) => ipcRenderer.on("downloads:map:failed", (_event, downloadInfo) => callback(downloadInfo as DownloadInfo)),
+    onDownloadMapStart: (callback: (downloadInfo: DownloadInfo) => void) => ipcRenderer.on("downloads:map:start", (_event, downloadInfo) => callback(downloadInfo)),
+    onDownloadMapComplete: (callback: (downloadInfo: DownloadInfo) => void) => ipcRenderer.on("downloads:map:complete", (_event, downloadInfo) => callback(downloadInfo)),
+    onDownloadMapProgress: (callback: (downloadInfo: DownloadInfo) => void) => ipcRenderer.on("downloads:map:progress", (_event, downloadInfo) => callback(downloadInfo)),
+    onDownloadMapFail: (callback: (downloadInfo: DownloadInfo) => void) => ipcRenderer.on("downloads:map:fail", (_event, downloadInfo) => callback(downloadInfo)),
```
jkells, line 147: This is an actual bug, `failed` on one side and `fail` on the other. I went with `fail` for consistency with the other events.

src/preload/preload.ts
```diff
@@ -162,13 +160,13 @@ contextBridge.exposeInMainWorld("misc", miscApi);
 function request<C extends GetCommandIds<"user", "server", "request">>(
     ...args: GetCommandData<GetCommands<"user", "server", "request", C>> extends never ? [commandId: C] : [commandId: C, data: GetCommandData<GetCommands<"user", "server", "request", C>>]
 ): Promise<GetCommands<"server", "user", "response", C>> {
-    return ipcRenderer.invoke("tachyon:request", ...args);
+    return ipcRenderer.invoke("tachyon:request", ...args) as Promise<GetCommands<"server", "user", "response", C>>;
```
jkells, line 163: These tachyon messages are the only remaining messages with casts. Unfortunately the trick we've used with conditional types to support 1 or 2 arguments gets lost when you pass it to the `Parameters` utility type, it's just a bit too magic.

 I couldn't get past it so just explicitly cast this one.

src/preload/preload.ts
```diff
@@ -162,13 +160,13 @@ contextBridge.exposeInMainWorld("misc", miscApi);
...
 }
 
 function onEvent<C extends GetCommandIds<"server", "user", "event">>(eventID: C, callback: (event: GetCommandData<GetCommands<"server", "user", "event", C>>) => void) {
     return ipcRenderer.on("tachyon:event", (_event, event) => {
-        if (event.commandId === eventID) {
-            callback(event.data);
+        if (event.commandId === eventID && "data" in event) {
```
jkells, line 168: Possibly not necessary to check for "data" key here but it's not guaranteed to be there by Typescript so it can't hurt to guard against it and fix the error.

src/renderer/store/maps.store.ts
```diff
@@ -72,6 +72,10 @@ export async function fetchMissingMapImages() {
 async function fetchMapImages(map: MapData) {
     if (map.imagesBlob?.preview) return;
     const arrayBuffer = await window.maps.fetchMapImages(map.images?.preview);
+    if (!arrayBuffer) {
```
jkells, line 75: I believe this is a real bug, This can return undefined.

src/renderer/views/singleplayer/scenarios.vue
```diff
@@ -74,7 +74,8 @@ import Markdown from "@renderer/components/misc/Markdown.vue";
 import DownloadContentButton from "@renderer/components/controls/DownloadContentButton.vue";
 import { gameStore } from "@renderer/store/game.store";
 
-const loadedScenarios = await window.game.getScenarios(gameStore.selectedGameVersion?.gameVersion);
+const gameVersion = gameStore?.selectedGameVersion?.gameVersion;
```
jkells, line 77: Small tweaks to eliminate an undefined argument which the other side didn't handle.

src/main/typed-ipc.ts
```diff
@@ -0,0 +1,125 @@
...
+import { Info } from "@main/services/info.service";
+import { ipcRenderer as electronIpcRenderer, ipcMain as electronIpcMain, IpcMain, IpcMainEvent, IpcMainInvokeEvent, IpcRenderer, IpcRendererEvent, WebContents } from "electron";
+import { logLevels } from "@main/services/log.service";
+import { MapData } from "@main/content/maps/map-data";
+import { MultiplayerLaunchSettings } from "@main/game/game";
+import { NewsFeedData } from "@main/services/news.service";
+import { Replay } from "@main/content/replays/replay";
+import { Scenario } from "@main/content/game/scenario";
+import { Settings } from "@main/services/settings.service";
+import { TachyonEvent } from "tachyon-protocol";
```
p2004a, line 14: I guess plenty of those could be `import type`.

jkells: Yep good call, will update thanks.

p2004a:
Sorry for delay in the final review, this is awesome improvement, thanks a lot!

PR #379, opened 2025-05-15
Default lobby server added

Hectate:
Fixes an error where the default server was missing for first-launch users.

The lint, format, and typecheck scripts have been run.

### DIFF
--- a/src/main/services/settings.service.ts
+++ b/src/main/services/settings.service.ts
@@ -19,7 +19,7 @@ const settingsStore = new FileStore<typeof settingsSchema>(path.join(CONFIG_PATH
     battlesHideLocked: false,
     battlesHideEmpty: true,
     logUploadUrl: "https://log.beyondallreason.dev/",
-    lobbyServer: "",
+    lobbyServer: "server4.beyondallreason.info",
     customServerList: [],
 });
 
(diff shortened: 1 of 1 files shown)


src/main/services/settings.service.ts
```diff
@@ -19,7 +19,7 @@ const settingsStore = new FileStore<typeof settingsSchema>(path.join(CONFIG_PATH
     battlesHideLocked: false,
     battlesHideEmpty: true,
     logUploadUrl: "https://log.beyondallreason.dev/",
-    lobbyServer: "",
+    lobbyServer: "server4.beyondallreason.info",
     customServerList: [],
 });
```
p2004a, line 24: I'm confused why this default object here is needed at all when the defaults are already specified in the json schema in src/main/json/model/settings.ts.

This shouldn't be needed.

Hectate:
I thought so as well, but it seems as if it is needed. What I suspect is happening is that the v-model is binding the value to the default in the settings, not from the Schema. Since the v-model is the source of truth for the setting, the Schema default is ignored. Thus, on first launch it finds an empty string and sets the value to that.

Hectate:
Oh, I think I see what you mean. Since the default is overridden by the empty string, we never get the default from the Schema.

Is it just meant to be left out entirely then? I've not used these JSON Schemas before.

Hectate:
Yes, that's exactly what is happening. The defaults existed in both the `settingsSchema` and were repeated, unnecessarily, in the `settings.store.ts`. This updated PR now removes all the redundant defaults and makes the schema the single source of defaults for the setting store.

p2004a (approved):

PR #386, opened 2025-05-25
Launch replay with non-live maps

### LINKED ISSUE #348: Watching replays using non-live maps
If the replay uses a map that is no longer in the live map set, lobby doesn't support downloading it and playing such replay.

1. Download replay like https://bar-rts.com/replays/de127e664eab39d78fca7a1bdda31eca
2. It's impossible to watch it

JasonReed-2:
Fixes #348 by supporting tracking of download progress and install status of "non-live" maps and updating the UI to allow launching replays for non-live maps.

Launching games with non-live maps is not supported by these changes.

This also fixes a pre-existing bug I found when downloading a map, deleting said map, then redownloading the map again. The lobby would never download the map until you restarted the lobby because the system never removed the map from `currentDownloads` array

### DIFF
--- a/src/renderer/views/library/replays.vue
+++ b/src/renderer/views/library/replays.vue
@@ -126,6 +126,7 @@ import { useDexieLiveQueryWithDeps } from "@renderer/composables/useDexieLiveQue
 import ReplayPreview from "@renderer/components/battle/ReplayPreview.vue";
 import DownloadContentButton from "@renderer/components/controls/DownloadContentButton.vue";
 import { gameStore } from "@renderer/store/game.store";
+import { MapDownloadData } from "@main/content/maps/map-data";
 
 const endedNormally: Ref<boolean | null> = ref(true);
 const showSpoilers = ref(true);
@@ -154,9 +155,13 @@ const replays = useDexieLiveQueryWithDeps([endedNormally, offset, limit, sortFie
     return query.reverse().sortBy(sortField.value);
 });
 
-const map = useDexieLiveQueryWithDeps([() => selectedReplay.value?.mapSpringName], () => {
-    if (!selectedReplay.value) return;
-    return db.maps.get(selectedReplay.value.mapSpringName);
+let map = useDexieLiveQueryWithDeps([() => selectedReplay.value?.mapSpringName], async () => {
+    let selected = selectedReplay.value;
+    if (!selected) return;
+
+    let selectedMap = (await db.maps.get(selected.mapSpringName)) as MapDownloadData | undefined;
+
+    return selectedMap ?? db.nonLiveMaps.get(selected.mapSpringName);
 });
 
 function onPage(event: DataTablePageEvent) {
--- a/src/renderer/views/library/replays.vue
+++ b/src/renderer/views/library/replays.vue
@@ -126,7 +126,6 @@ import { useDexieLiveQueryWithDeps } from "@renderer/composables/useDexieLiveQue
 import ReplayPreview from "@renderer/components/battle/ReplayPreview.vue";
 import DownloadContentButton from "@renderer/components/controls/DownloadContentButton.vue";
 import { gameStore } from "@renderer/store/game.store";
-import { MapDownloadData } from "@main/content/maps/map-data";
 
 const endedNormally: Ref<boolean | null> = ref(true);
 const showSpoilers = ref(true);
@@ -159,9 +158,9 @@ let map = useDexieLiveQueryWithDeps([() => selectedReplay.value?.mapSpringName],
     let selected = selectedReplay.value;
     if (!selected) return;
 
-    let selectedMap = (await db.maps.get(selected.mapSpringName)) as MapDownloadData | undefined;
+    const [live, nonLive] = await Promise.all([db.maps.get(selected.mapSpringName), db.nonLiveMaps.get(selected.mapSpringName)]);
 
-    return selectedMap ?? db.nonLiveMaps.get(selected.mapSpringName);
+    return live ?? nonLive;
 });
 
 function onPage(event: DataTablePageEvent) {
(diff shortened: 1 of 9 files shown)


p2004a:
> Launching games with non-live maps is not supported by these changes.

Do you mean downloading?

Without this it partially, not fully, resolves #348.

p2004a:
For this issue, you don't need to support starting games with custom maps, only allow to start download of custom maps for replays.

So user can download any replay from site, and lobby will always download correct assets and start it.

p2004a:
> but you can select a replay, download the map for that replay (even if non-live) and launch it.

That doesn't work, with this PR user needs to manually download the map they need, and only then they are able to play the replay. The "watch" button is greyed out when the map isn't already available locally.

JasonReed-2:
Ah sorry. I had this working at one point during the development and did not catch this during testing after I refactored. Working on this now.

p2004a (changes requested):
Almost there :)

It doesn't handle the case where non-live map was on disk, then was deleted, and then user wants to watch replay wit it. Basically the list of non-live maps is never cleared up from the database (Maybe sounds like yet another argument to see if we could cleanly not store it in indexed db, but only in memory, but I don't know yet how to do it cleanly).

src/renderer/views/library/replays.vue
```diff
@@ -154,9 +155,23 @@ const replays = useDexieLiveQueryWithDeps([endedNormally, offset, limit, sortFie
...
+    const [live, nonLive] = await Promise.all([db.maps.get(selected.mapSpringName), db.nonLiveMaps.get(selected.mapSpringName)]);
+
+    let map = live ?? nonLive;
+
+    if (!map) {
+        map = {
+            springName: selected.mapSpringName,
+            isDownloading: false,
+            isInstalled: false,
+        } satisfies MapDownloadData;
```
p2004a, line 171: Ok, I have idea, what if we delete the whole `nonLive` table, keep only this snippet here so that something is always returned, and just verify here `isInstalled` by calling the backend which should at all times know all live and non live maps because it did the scanning? Do I miss something?

p2004a: Maybe you had some idea how to start game with custom map that will require to have this table?

JasonReed-2: Your approach would work. But what are you thinking would be the benefit? I followed the design of how we handle the live maps to keep a somewhat consistent way of handling map installed and download status across both "live" and "non-live" maps (i.e store both in the indexDB). But nothing really further.

Is there a way to derive the metadata from a map file directly? If so: maybe there is no need to distinguish between the two in two separate tables. Instead: it could be an additional column. That would be the ideal in my mind.

JasonReed-2:
> Almost there :)
> 
> It doesn't handle the case where non-live map was on disk, then was deleted, and then user wants to watch replay with it. Basically the list of non-live maps is never cleared up from the database (Maybe sounds like yet another argument to see if we could cleanly not store it in indexed db, but only in memory, but I don't know yet how to do it cleanly).

For this one: I think this should be fine because deleting the file should automatically update the nonlive DB to flag that the map is no longer installed and thus prompt the user to download the map file. So even though the map is in the nonLiveDB: the install status is "false".

gif of the workflow:
electron_kXlcQxVrC9

p2004a:
> For this one: I think this should be fine because deleting the file should automatically update the nonlive DB to flag that the map is no longer installed and thus prompt the user to download the map file. So even though the map is in the nonLiveDB: the install status is "false".

Not if map is deleted when lobby is not running.

src/renderer/views/library/replays.vue
p2004a: > Is there a way to derive the metadata from a map file directly?

To some degree yes, it actually existed already and was dropped in https://github.com/beyond-all-reason/bar-lobby/pull/278, I've commented on that PR, but the [follow up didn't happen](https://github.com/beyond-all-reason/bar-lobby/pull/278#issuecomment-2466903283)

JasonReed-2: > Ok, I have idea, what if we delete the whole `nonLive` table, keep only this snippet here so that something is always returned, and just verify here `isInstalled` by calling the backend which should at all times know all live and non live maps because it did the scanning? Do I miss something?

Oh yeah: I forgot. Preserving the reactivity was part of the decision to use the indexDB as well. This is how it looks with a simple implementation of your idea:
electron_Xm9lvv5Me3

The `useDexieLiveQueryWithDeps` function gets triggered when the DB is updated. If we don't update the DB anymore because we dont store this in nonLive DB: we lose that reactivity.

There probably is some way around this but would introduce more complexity to the front-end component.

JasonReed-2:
That's fair. I'll fix this. Thanks!

JasonReed-2:
Ok I've documented a testing plan for this to make sure I cover all bases.

Scenario 1: PASS
- Non-live map not in DB and not in folder before lobby starts.
- Launch lobby
- Able to download map? Yes
- Upon clicking download map: downloads map and updates UI to show "Launch" as expected.
- Nonlive DB updated as expected.

Scenario 2: PASS
- Lobby is launched already and nonlive map is in DB and in folder.
- Delete map from the folder
- Nonlive DB updated to reflect that the map is no longer installed and UI updates to show "Download Map"

Scenario 3: PASS
- Lobby is launched already and nonlive map is in DB [with installed = false] but map is not in folder.
- Lobby shows "Download Map" button.
- Clicking the download button updates the nonlive map in the DB [installed = true]
- After download: can launch the replay

Scenario 4: PASS
- Lobby is not launched and nonlive map is in the DB [installed = true] but removed from the folder.
- Launch Lobby
- The nonlive DB shows installed = false and the UI allows user to download the map
- After download: can launch the replay

Scenario 5: PASS
- Lobby is not launched and nonlive map is not in the DB [installed = false] but a user manually downloads the map into the map folder
- Launch lobby
- The nonlive DB shows installed = true and the UI allows user to launch the replay

p2004a (approved):
Great, thanks a lot for this!

Issue #387, opened 2025-05-25
Allow customization of install directories


JasonReed-2:
See #367 for discussion on the types of files that lobby uses / downloads.

We should have some way for users to customize where these files are downloaded if they desire to.

p2004a:
See also https://github.com/beyond-all-reason/bar-lobby/pull/395 for even more analysis.

Issue #384, opened 2025-05-25
Initial engine download should be opt-in


Hectate:
Upon launching, if there is no engine installed one is automatically downloaded. There is no prompt to the user to start this download, nor is there any way to opt out or cancel if it has started. I think the intent is to eventually ship the launcher with an initial install of the engine, so nothing is required if the user is simply playing offline or creating a portable installation?

In any case, the presumptive installation of the initial engine should at least require some action by the user outside of launching bar-lobby, and have a cancel option while it is in progress. This would protect against complaints from users with limited bandwidth and data caps.

The login option should be available even if no engine is downloaded. To be fair, there should be warnings or notices that no engine is currently installed if the user is logged in and accessing relevant sections of the application without an engine installed. Various actions, like joining an active lobby, playing a scenario, or trying to watch a replay (semi-related: https://github.com/beyond-all-reason/bar-lobby/issues/350) should probably be blocked until the appropriate version is installed.

Hectate:
> I'm way less sure about this one. I'm not convinced spending effort of supporting this half-installed state is worth it at all. With the steam install ~everything will be also packaged as part of steam install.

I don't disagree, I also think the support burden is non-ideal. The only two edge cases I could think of that would be impacted by this decision were:

1. Users desiring to run older, custom, or development versions of the engine will still be forced to install the current engine regardless.
2. Users that delete/uninstall all versions could find themselves in this half-installed state regardless.

The first is a fairly extreme edge case in my mind, so we can probably disregard and just say this is a design choice.

The second could be handled by not allowing the deletion of whatever the "current" or "stable" version is in whatever "installed version management" function we have for the app. Should a user (or an overzealous anti-virus on their machine?) manually delete versions, however, we need to be able to notice and warn the user that they need to re-download.

p2004a:
For second one, we won't offer deleting of default engine version. So far there isn't way to delete old engines in UI anyway. If user deletes engine manually, it will be just fetched on next lobby run. If they delete engine while lobby is running: we can just let lobby error out.

Hectate:
Closing this because the behavior is now to prompt the user for an install location first. This solved my concerns.

PR #395, opened 2025-05-29
Rework data install locations

### LINKED ISSUE #310: Correct installation paths on Windows and Linux

p2004a:
Change location of installation and files manager by the lobby:
- Splits the files into static assets and more dynamic config/state/logs/replays location. That includes how engine operates on data, it's no longer a single directory! Logs, replays, files created by game are in entirely separate "data" directory from where rapid pool, maps, engines etc are stored.
- Change Windows install directory name and Linux files to the same `BeyondAllReason` (it was either "bar-lobby", or "Beyond All Reason" depending on context. Currently in old launcher it's a mix of "Beyond All Reason" and "Beyond-All-Reason")
- Allow to override location using env variables `BAR_ASSETS_PATH` and `BAR_STATE_PATH`.
- Separate location for data in development vs normal install

#### Examples

See the detailed description about locations in `src/main/config/app.ts`, copying from there simplified summary:

- Windows:
  - Lobby program install location: `AppData\Local\Programs\BeyondAllReason`
  - Assets: `AppData\Local\Programs\BeyondAllReason\assets` (maybe we will over convenient customization for this one, and in my opinion only this one, in the future https://github.com/beyond-all-reason/bar-lobby/issues/387)
  - State: `AppData\Roaming\BeyondAllReason`
    - engine *write dir* `AppData\Roaming\BeyondAllReason\data`
- Linux:
  - Lobby program install location: N/A, user puts AppImage wherever they want. 
  - Assets: `~/.local/share/BeyondAllReason/assets`
  - State: `~/.local/state/BeyondAllReason`
    - engine *write dir* `~/.local/state/BeyondAllReason/data`
- Under development (npm run start):
  - Assets: `./assets`
  - State: `./state`

So, for example:

  - `AppData\Local\Programs\BeyondAllReason\assets\engine\2025.01.6\spring.exe`
  - `~/.local/share/BeyondAllReason/assets/pool/00/009a03b038e7f87ddc9cc11b7ce523.gz`
  - `AppData\Roaming\BeyondAllReason\data\demos\2025-05-29_23-08-17-805_Supreme%20Isthmus%20v2_2025.04.04.sdfz`
  - `~/.local/state/BeyondAllReason/data/LuaUI/Config/BYAR.lua`
  - `~/.local/state/BeyondAllReason/config/account.json`

### Direction for Steam

Lobby executable + Assets will be under `[Steam Install]\SteamApps\common\[Game Folder]\`. The location of state files will be mostly as defined here, the only thing to tune is to add one more subdirectory with steam account id as described here https://partner.steamgames.com/doc/features/cloud (you can change steam account while signed in to the same system account, state files should be then separate).

Zero-K, as far as I understand and tested currently, puts *everything* under SteamApps, which means that all users on the PC share all data, including the login, settings and so on: IMHO we should do better.

### Custom install location

This is likely going to be a feature that will be requested by many players, primarily on Windows. Just enabling assisted installer in electron-builder (see https://www.electron.build/nsis.html) has many implications that make it problematic:
- Naturally less seamless
- It also allows to select per-machine install and users run into permission issues
- The installer also allows to change parameters during the *update*, and in launcher we had issues some users always breaking their own install

To support this there are 3 paths then:
- Just accept that it will be required, and do exactly what launcher does, and unfortunately it will cause the same issues
- Pick the assisted installer, but also write custom NSIS scripts which is allowed by electron-builder and make the update process more seamless, maybe add a bit more tuning
- We will have some in-lobby first setup, there are already requests like https://github.com/beyond-all-reason/bar-lobby/issues/384, and there will be more. We could manage the *assets*, not full lobby install path from inside of lobby.

### WHY split assets and state

Pros:

- Uninstalling keeps the state around.
- Multi-user installation support (for steam for sure, for standalone installer if we would support per-machine install (see previous point)) with separate per-user data.
- Easier for users to back up their configuration without having to separate it from game assets. Also easier to purge configuration without reinstall.
- It much better follows the respective operating system standards, and other software can make assumptions based on that (e.g. profile roaming under Windows, file history, backups, etc).
- There is also benefit for development where we can have multiple independent state directories with different configs without a bunch of manual symlinks.
- There isn't really a good reason for users to touch anything in assets directory. Custom maps, widgets etc? They can put it in state. It actually potentially allows us to fully automatically manage old versions of maps and so on, while still allowing users to place custom ones and not fear interference.

Cons:
- It might be more confusing for users to locate the files they are searching for, and overall some users will prefer to have all the data together in single place for convenience and what not.
- As visible in the PR, increases the lobby code complexity a bit.

---

Fixes #310


p2004a:
Maybe assets under Linux should live under `~/.local/share/BeyondAllReason/assets` for *some* consistency :thinking: ?

Issue #400, opened 2025-06-03
Add a "Downloaded" filter for Map Previews


Hectate:
It is not currently possible to filter the Library > Maps list by "already downloaded" maps. I recommend adding this as a filter option just below the "Favorites Only" checkbox.

There is also no way to know from the list if a map is already downloaded. An icon, perhaps in the top-right combined with the "Favorite" heart would be an ideal, minimal way to display this. The icon best to be used would probably be one from a "file-check" group, in my opinion. (https://icon-sets.iconify.design/mdi/file-check/)

PR #403, opened 2025-06-09
Add downloaded filter for maps

DarkEyeDragon:
Implements #400 
image

Both filters can be selected and will only show what matches both filters.
image

image
image


p2004a (approved):
Nice! Thank you!

PR #405, opened 2025-06-11
Fix username display and terrain icon layout issues

### LINKED ISSUE #391: Display correct username when logged in
Image
The name displayed on the top right seems to be somewhat random.
Upon connection, the server sends a [user/self](https://github.com/beyond-all-reason/tachyon/blob/master/docs/schema/user.md#self) event with the player name in it but the client doesn't seem to handle it.

### LINKED ISSUE #399: Terrain Icons can overlap other map details on preview card
This is visible under Library > Maps. The terrain icons on the map previews can overlap the other details (size, player count) to the left, if there are many of them. See image below.

Image

In theory, we want to stack these start a second row above. An alternative option would be to simply rank/prioritize the icons, and only show the X-highest ranking ones that would fit. Filtering would still work, of course.

akluu:
# Changes

- Added CSS rules so terrain icons create another row instead of overlapping

Fixes #399

Screenshot 2025-06-09 181856

- Added an event handler for `user/self` that populates `me` with relevant data

Fixes #391

> **Note:** The data doesn't persist with UI refreshes, but this should be a dev-only issue.

### DIFF
--- a/src/renderer/components/maps/MapOverviewCard.vue
+++ b/src/renderer/components/maps/MapOverviewCard.vue
@@ -140,6 +140,9 @@ const imageUrl = computed(() =>
     &.br {
         bottom: 10px;
         right: 10px;
+        flex-wrap: wrap-reverse;
+        justify-content: flex-end;
+        max-width:55%;
     }
 }
 </style>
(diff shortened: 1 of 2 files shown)


p2004a:
Thank you :smile: !

Issue #350, opened 2025-03-30
Map and engine download progress when watching replay


p2004a:
When you click watch, there needs to be indicator of download progress for the missing resources.

- [X] map (resolved in #407)
- [ ] engine (see https://github.com/beyond-all-reason/bar-lobby/issues/350#issuecomment-2974656637)
- [ ] game (see https://github.com/beyond-all-reason/bar-lobby/issues/350#issuecomment-2974656637)

Hectate:
OK, more digging shows that game/engine downloads, if missing for the replay, are handled here: https://github.com/beyond-all-reason/bar-lobby/blob/2aa4302c4269d66e3195cc979f6a2614f9636e31/src/main/game/game.ts#L94

Definitely deeper than I expected. I suggest a separate issue to raise this up to the UI level.

p2004a:
> Definitely deeper than I expected. I suggest a separate issue to raise this up to the UI level.

Ugh, indeed that's _deep_. I think it's fine to repurpose this issue for this. I will just update the top level description.

Hectate:
FYI; there are sub-tasks still to be completed that were not done as part of #473 as merged, so I am re-opening this. I'll come back to the rest of it in the future unless someone else does first.

PR #407, opened 2025-06-11
Map download progress added to Library > Map/Replay

Hectate:
This adds a small progress bar that appears when a map is being downloaded from either the Library > Replays or Library > Maps > [id] views. The progress bar appears below the buttons, and then disappears when complete. It displays the progress as both a horizontal progress bar as well as a two-digit number with percentage sign after (e.g. "50%").

Lint, Typecheck, and Format have been run.

Fixes https://github.com/beyond-all-reason/bar-lobby/issues/350

### DIFF
--- a/src/renderer/views/library/maps/[id].vue
+++ b/src/renderer/views/library/maps/[id].vue
@@ -68,9 +68,14 @@
                                 <Button v-else-if="map.isDownloading" class="green inline" disabled>Downloading map...</Button>
                                 <Button v-else class="red inline" @click="downloadMap(map.springName)">Download</Button>
                             </div>
-							<div class="padding-top-md padding-bottom-md">
-								<Progress :class="{ pulse: isDownloading }" :percent="downloadPercent" v-if="isDownloading" :text="(downloadPercent * 100).toFixed(0) + '%'"></Progress>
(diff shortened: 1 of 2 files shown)


Hectate:
This branch was from an older master, so I'm touching up now.

p2004a:
This doesn't actually show the download progress for the map the is being downloaded. It shows download progress for *everything* that is currently in the download queue, I don't believe that this is intended behavior.

There is also code duplication across 2 components, I would expect such component to be refactored out.

Hectate:
Ah, yes I forgot to clean up the duplication. The download progress I'll have to look at closer then. Thanks.

Hectate:
@p2004a This should be ready to merge now. I solved the progress math issue with the de-duplication, as well as the per-map download (and not all downloads). With this, we have a specific component for MapDownloadProgress that just needs a map's "Spring name" to function. This will give us the benefit of being able to easily add it to multiplayer lobby screens if a selected map requires download.

Minimal styling is done on this, but as a new SFC any desired styling can be applied directly to the component, or the Progress component it uses, should appearance changes be needed in the future.

p2004a:
Thank you!!

PR #408, opened 2025-06-14
fix bug when dragging coop bots

webdevsavvy:
when dragging coop bots to the same team or outside the component the ui becomes unresponsive, this PR fixes that bug


p2004a:
Thank you!

PR #417, opened 2025-06-15
Set up basic application auto updates

### LINKED ISSUE #314: Set up basic application auto updates
This isssue tracks configuring electron updater following: https://www.electron.build/auto-update.html to enable automatic client updates of client from GitHub releases.

We should try as hard as possible to use one-click installer, not assisted where somebody needs to select something, so maybe the update can be downloaded in background and applied only on next application start.

This basic request doesn't include anything like [update dialog](https://github.com/beyond-all-reason/spring-launcher/blob/master/src/update_dialog.js) with changelog or anything like that.

webdevsavvy:
This PR aims to add basic auto updating of the game client through GitHub releases.

Resolves #314 

### Added
- IPC bridge for managing client updates from the renderer
- Visual feedback when initializing the app to indicate there is an update in progress

### Changed
- Encapsulated the logic for updating the game client


webdevsavvy:
There was already a previous attempt to add this, so far I've fixed this code but I'm having trouble testing as it would require a new release I think.

webdevsavvy:
> What I did in the past for spring-launcher to test is to set locally in package.json version to be smaller then current latest version, build and install that package, and then it will trigger the update flow.

It worked! pushing this as ready to review.

webdevsavvy:
Now that I think about it I didn't test the auto updates on windows, whats the best way to setup a test env for windows, docker?

webdevsavvy:
Auto updater works on windows and linux.

webdevsavvy:
> What do you think about hooking it up into the `InitialSetup.vue` that handles all other updates, and not make it like here out of bound?

I didn't understand what you mean :thinking:

webdevsavvy:
Ahh ok, yes that would be preferable, but I thought this was outside the scope for this issue, I can add the necesary hooks to have it be managed by the client

webdevsavvy:
I've added basic ipc handlers and events for the auto updates but testing I realized that `InitialSetup.vue` might not be the place to inject the auto updater service, as it is conditionally rendered depending on the `state` ref in `App.vue` and it should run even when the app has already ran the _initial setup_.

The way I imagine the auto updater working is:
1. check for available updates
2. download the new update if exists
3. if the update is a major version change (following semver) force restart to update
4. if the update is a minor version prompt the user with the new update and a choice to restart or apply on next start up
5. finally install the update on application restart

webdevsavvy:
Added the update process to  `InitialSetup.vue` and made it a one click install.

Right now when the download finishes it automatically restarts the application, should we prompt the user before restart?

src/main/content/auto-updater.ts
```diff
@@ -0,0 +1,116 @@
...
+    public async downloadUpdate(): Promise<void> {
+        if (!this.intialized) return;
+        return await new Promise<void>((resolve) => {
+            autoUpdater.on("download-progress", (progressInfo) => {
+                this.downloadProgress({
+                    type: "update",
+                    name: this.updateInfo?.version,
+                    currentBytes: progressInfo.transferred,
+                    totalBytes: progressInfo.total,
+                } as DownloadInfo);
```
p2004a, line 90: no casts please

p2004a (changes requested):
Overall looks good, I've not tested myself yet though.

> Right now when the download finishes it automatically restarts the application, should we prompt the user before restart?

I'm fine with current behavior, in the future we can add showing release notes and confirmation.

p2004a (approved):
Looks good, thank you! I will test it tomorrow/saturday and merge if all works correctly :slightly_smiling_face:

Issue #316, opened 2025-02-23
Basic support for internatiolization


p2004a:
This is task to adopt some internationalization framework solution to the code base early, so we can easily plug in translations later.

Hectate:
I'm doing some testing of this. Worth to note, there is already some basic framework setup using [vue-i18n](https://github.com/beyond-all-reason/bar-lobby/blob/f46091b639d4a1fe0b33570338c3d85fb7964f5f/src/renderer/index.ts#L41), so my plan is to validate that it is set up correctly by implementing a few translated strings.

PR #418, opened 2025-06-15
Add basic support for internatiolization

webdevsavvy:
This PR aims to add basic support for i18n through game translation files and vue-i18n

Contributes to #316 

### Added
- Copied translation files from main game repo into assets folder
- Utility function for processing translation file data
- Bash script to look for translation targets

### Changed
- Modified the setup of `vue-i18n` to manage new translation file directory structure


webdevsavvy:
Should I create a new issue for tracking translations needed to be implemented?

p2004a:
The part where translators provide text, and what to request etc, I'm not familiar with. Watch The Fork set it up for BAR. You can open tracking issue if you want to figure that out.

src/renderer/utils/i18n.ts
```diff
@@ -7,3 +7,12 @@ export const pluralize = (str: string, quantity: number, locale: string = "en-US
     if (rule == "one") return str;
     return str + "s"; // TODO: link to i18n translation files
 };
+
+export const processTranslationData = (jsonData) => {
+    for (const key in jsonData) {
+        if (typeof jsonData[key] === "object") {
+            jsonData[key] = processTranslationData(jsonData[key]);
+        } else if (typeof jsonData[key] === "string") jsonData[key] = jsonData[key].replace(/%\{(.*?)\}/, "{$1}");
```
p2004a, line 15: hmm, is it guaranteed there is always going to be at most one substitution?

src/renderer/views/development/game-dev.vue
```diff
@@ -19,15 +19,18 @@ SPDX-License-Identifier: MIT
     - Unit Definitions
     "
         />
+        {{ t("interface.ui.topbar.resources.incomeTooltip", { resource: "9000" }) }}
```
p2004a, line 22: what would happen if this field didn't exist? I would assume full crash?

webdevsavvy: it would output the fallback or if that doesnt exists then the string `interface.ui.topbar.resources.incomeTooltip`

webdevsavvy: i believe there is already some check from vue-i18n that can be hooked into, but it is runtime bound so i dont think it would be good for proper tests, so maybe a script that checks calls to `t()` and constrasts them with the json files?

webdevsavvy: there is this package [https://github.com/Spittal/vue-i18n-extract](https://github.com/Spittal/vue-i18n-extract)

webdevsavvy: I created the issue but lack the permissions to add labels or tag as a subissue. I'll leave that to you if thats ok :)

p2004a: thanks, done, I've also send you invite that gives you Triage permission in the repo

PR #420, opened 2025-06-16
Add explicit protocol selection for lobby servers

### LINKED ISSUE #382: The server selection should be able to select plain TCP (as opposed to TLS)
By default it's using TLS, even for things like localhost where there's usually no TLS setup.
There should be an option to disable TLS, so do the auth flow using http, and then establish the websocket connection using `ws` (not `wss`)
Or maybe, add the protocol in the server selection, so one can type `http://localhost:4000`, and base the ws connection off that as well.

akluu:
Fixes #382

## Changes

- Server config now requires explicit protocol
- Updated default servers to include protocols
- Also added protocol conversion for OAuth endpoints (ws -> http, wss->https)
- Small change to UI placeholder to show protocol example

### DIFF
--- a/src/main/config/server.ts
+++ b/src/main/config/server.ts
@@ -11,14 +11,6 @@ function getLobbyServer() {
     return settingsService.getSettings().lobbyServer;
 }
 
-function parseServerURL(server: string, defaultProtocol: string): URL {
-
-    if (!/^\w+:\/\//.test(server)) {
-        server = `${defaultProtocol}://${server}`;
-    }
-    return new URL(server);
-}
-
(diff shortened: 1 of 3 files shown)


p2004a (changes requested):
This can be done in more idiomatic way

- Use `new URL` constructor to parse url, then operate on parsed object. E.g. you can do `url.protocol = 'wss'` instead of spring splitting
- Combine it with something like `function isSecure(url) { return url.protocol == 'https:' || url.protocol == 'wss:'; }`

p2004a (approved):
Thank you!

PR #422, opened 2025-06-17
Replay screen updates

### LINKED ISSUE #406: Map image and Team list in Library > Replays do not share space well
Once you select a replay, the team list and map image displayed to the side end up scaled wildly inconsistently depending on how well the map fits within the preview space.

For comparison; here is a "tall" versus a "wide" map.

Image

Image

Outside of completely changing the setup of this page, there's probably nothing we can do about the wider maps having tiny preview images. However, giving the team list only a single line of text in a scrolling box is probably non-ideal also.

A simple fix would be to simply give the map and text box maximum sizes as a percentage. Another option would be to switch to a tabbed setup for the preview, so the full preview can be either the team list _or_ the map preview, but not both at the same time. Potentially relevant for extremely large player lists.

Hectate:
Fixes #406

The right panel of the Replay view now displays a tabbed interface. Two tabs are presented when a replay is selected; Map and Details.

The **Map** tab will display the existing map preview with player positions. It is the default view.

The **Details** tab will display:
* Engine version
*  Game version
* Team/player list 

We could extend this to include advanced replay details as well, if desired in bar-lobby.

I have also corrected two issues with the "Download Map" button.
* The string "Download Map" was not styled to match the other versions of the buttons. It now is properly styled.
* The string "Downloading Map..." was so large it would wrap to a second line, causing the UI to jump around. This text only is now shrunk slightly to fit on a single line to prevent this.

- [x] I have read the [**CONTRIBUTING.md**](CONTRIBUTING.md) file.
- [x] I've scoped pull request to a single change or feature.
- [x] I have tested my changes locally.
- [x] I have updated the documentation, if necessary.

**Images:**
image
image
image


Hectate:
FYI, I have left this as a draft for UI approval first.

Hectate:
Updates made per UI review.
1. Button lettering is now sentence case.
2. Team rosters and replay details positions reversed as requested.
3. Table styling changes to match website.
4. Downloading map progress bar now overlaps the (disabled) button. Button color also changed to grey during the downloading state.
5. Additional details can be easily added later once it is determined exactly which details are desirable. The content will match appropriate style and overflow will result in a scrollbox in the same way that the team's list does as needed.

**Bug Fixes**
Separately, I identified that the list of Players was not displaying if the Replay was a FFA match. This was a preexisting issue that had not yet been identified. This bug has been fixed also as part of this PR.

**Example Images**
image
image
image

Hectate:
Did some more UI review with Scopa. Almost all requested changes have been submitted. However, there is an issue with the scrollbox I need to correct, so this has been moved back to Draft for now.

akluu:
Hey, I am working on opening replays and I noticed that if I delete all my replays from replays folder the left panel updates with "No replays found" but the right panel still has the previous replay information.

Screenshot 2025-06-23 151818

It's problematic because you can still click Watch and it will throw a does not exist error.

Hectate:
> Hey, I am working on opening replays and I noticed that if I delete all my replays from replays folder the left panel updates with "No replays found" but the right panel still has the previous replay information.
> 
> [...]
> 
> It's problematic because you can still click Watch and it will throw a does not exist error.

Thanks for the report! If I can fix it with this update I'll do so. If not, I'll open a new Issue.

Hectate:
@akluu Your bug report is now fixed with this PR also. Thanks for the find!

p2004a:
@Hectate is all the UI design part of this PR finalized with Scopa or not? I've also send out https://discord.com/channels/549281623154229250/927564746104905728/1388862758510137354 to figure out best method to not get PRs stuck in a weird state where part of review happens in one place, part in different.

Hectate:
@p2004a Yes, my understanding is that it is good; all changes requested were made. We had a Discord chat with screen-sharing where we implemented everything except for the scrolling container, but that was clearly defined also.

I do not anticipate that to always be possible though, so I agree that we need to figure out a permanent solution for approvals.

PR #427, opened 2025-06-20
Revert "Add basic support for internatiolization"

p2004a:
Reverts beyond-all-reason/bar-lobby#418

- Doesn't work correctly when app is packaged (primary reason for revert)
- Invalid promise resolution
- Incorrect type for glob return


p2004a:
@webdevsavvy FYI

PR #428, opened 2025-06-20
Fix - Add Basic support for internatiolization

webdevsavvy:
Fixes a previously merged and reverted PR (#418 #427), due to errors on packaged versions of the app.

Additionaly a testing JSON file was added under `en` locale `lobby.json` for testing purposes, as well as a translation in the main view for the _Quick play_ button.

### DIFF
--- a/src/main/content/auto-updater.ts
+++ b/src/main/content/auto-updater.ts
@@ -1,0 +1,96 @@
+// SPDX-FileCopyrightText: 2025 The BAR Lobby Authors
+//
+// SPDX-License-Identifier: MIT
+
+import { Downloader } from "@main/content/abstract-content";
+import { logger } from "@main/utils/logger";
+import { app } from "electron";
+import electronUpdater from "electron-updater";
+const { autoUpdater } = electronUpdater;
+
+const log = logger("auto-updater.ts");
+
+export type UpdateInfo = electronUpdater.UpdateInfo;
+
+export class AutoUpdaterAPI extends Downloader {
+    private updateInfo?: UpdateInfo;
+    private intialized: boolean = false;
+    public readonly currentVersion: string = autoUpdater.currentVersion;
+
+    public async init() {
+        log.info("Initializing AutoUpdaterAPI");
+        if (!app.isPackaged) {
+            log.info("App is not packaged. Skipping initializaion");
+            return this;
+        }
+        autoUpdater.autoDownload = false;
+        autoUpdater.logger = log;
+        autoUpdater.autoInstallOnAppQuit = true;
+        this.intialized = true;
+        return this;
+    }
+
+    public get updateInformation(): UpdateInfo | null {
+        return this.updateInfo ?? null;
+    }
+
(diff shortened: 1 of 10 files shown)


src/renderer/index.ts
```diff
@@ -44,11 +45,35 @@ async function setupVue() {
...
+    // `any` required as these are loaded from json files with no defined format
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const messages: Record<string, Record<string, any>> = {};
+    const translationFiles: Record<string, Array<string>> = {};
+
     for (const filePath in localeFilePaths) {
-        const localeCode = filePath.match(/([a-z]{2})\.json$/)![1];
-        messages[localeCode] = localeFilePaths[filePath];
+        const localeCode = filePath.match(/\/([a-z]{2})\/.+?\.json$/)![1];
+        const translationFileURI = `${localeFilePaths[filePath]}`;
```
p2004a, line 55: Just `localeFilePaths[filePath]`, no need for string interpolation, this is why the previous bad typing of `localeFilePaths` wasn't cought.

src/renderer/index.ts
```diff
@@ -44,11 +45,35 @@ async function setupVue() {
 
 async function setupI18n() {
```
p2004a, line 46: How are you testing it? is there some way to switch locales for testing?

src/renderer/index.ts
```diff
@@ -44,11 +45,35 @@ async function setupVue() {
...
+
+    for (const locale in translationFiles) {
+        // prevent unnecesary processing of translation files and load only client locale and fallback
+        if (locale != myLocale || locale != "en") continue;
+        for (const translationFile of translationFiles[locale]) {
+            try {
+                const processedData = await fetch(translationFile)
+                    .then((res) => res.json())
+                    .then((jsonData) => processTranslationData(jsonData))
+                    .catch(() => console.error(`Error processing translation file (${translationFile}) for locale (${locale})`));
```
p2004a, line 68: It also could be just

```js
const resp = await fetch(translationFile);
const processedData = processTranslationData(await resp.json());
```

src/renderer/index.ts
webdevsavvy: hard coding values into `locale` property of `createI18n()' function in `src/renderer/index.ts`

src/renderer/index.ts
webdevsavvy: true, will change asap

src/renderer/index.ts
webdevsavvy: a better way is change locale in sensors in chrome dev tools

webdevsavvy: https://stackoverflow.com/questions/37221494/how-to-change-the-locale-in-chrome-browser

webdevsavvy: and yeah it was broken, didnt test after adding the continue statement and didnt realize it didnt work anymore

p2004a (approved):

PR #437, opened 2025-07-07
Fix Downloading Non-live Maps For Scenarios

### LINKED ISSUE #435: Map name versioning can prevent a scenario from being played
Certain scenarios cannot be played. The button to "Launch" them is forced to a disabled state. [The code](https://github.com/beyond-all-reason/bar-lobby/blob/29e1dc37394c65b521a0f7774b0522700d664bd9/src/renderer/views/singleplayer/scenarios.vue#L60) for this state is specifically driven by the existence of a `map<MapData>`, as determined by the filename available in each scenario's data (for example, "AcidicQuarry 5.16").

Vue Devtools look at one of the offending scenarios; "The Sky is the Limit".
Image

Unfortunately for this, it appears that the version name of the map has changed but the scenario list is using the old name, the failure to match results in a disabled scenario. Since the older version of the map is not currently installed, and only the newer version is available to download, the scenario cannot be played.

Vue Devtools look at the currently available filename for Acidic Quarry; "AcidicQuarry 5.17".
Image

It is not immediately obvious to me how scenario map filenames get updated. It appears to be part of the game content downloading API. I presume that it is possible that the game content version has an impact on this, but as such information is not exposed in the scenario screen it cannot be definitively determined. For what it is worth, the Vue Devtools show the loaded ``gameVersion`` is "Beyond All Reason test-28332-7027e9e".

Image showing the issue.

Image

For what it is worth; I did confirm that this particular scenario is functional in Chobby.

JasonReed-2:
Fixes #435

Allows non-live maps to be downloaded from the "Scenarios" tab for use for launching a scenario. Uses the same solution as the one used for downloading non-live maps for replays. See #386.

### DIFF
--- a/src/renderer/views/singleplayer/scenarios.vue
+++ b/src/renderer/views/singleplayer/scenarios.vue
@@ -75,6 +75,7 @@ import { Scenario } from "@main/content/game/scenario";
 import { LATEST_GAME_VERSION, DEFAULT_ENGINE_VERSION } from "@main/config/default-versions";
 import Panel from "@renderer/components/common/Panel.vue";
 import { db } from "@renderer/store/db";
+import { MapDownloadData } from "@main/content/maps/map-data";
 import { useDexieLiveQueryWithDeps } from "@renderer/composables/useDexieLiveQuery";
 import Markdown from "@renderer/components/misc/Markdown.vue";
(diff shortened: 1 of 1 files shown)


bcdrme:
Looks like it's doing the trick, merging.

PR #441, opened 2025-07-15
Add blurb regarding dev for multiplayer

geekingfrog:
so we can link that when someone on discord attempts to use the lobby with the wrong server [like that](https://discord.com/channels/549281623154229250/927564746104905728/1394549339774386216)


webdevsavvy:
Thanks for the doc update! It was not at all clear how to do this before, a much needed addition.

Issue #423, opened 2025-06-19
Give map downloads a max concurrent download limit


DarkEyeDragon:
Right now all downloads run at the same time.
This slows down the download to play time significantly when multiple maps are downloaded.

My suggestion therefore would be to make all downloads sequential. 
There is no benefit in downloading multiple maps at once since they all come from the same origin anyway so you're just speading bandwidth over all downloads.

p2004a:
> There is no benefit in downloading multiple maps at once since they all come from the same origin anyway so you're just speading bandwidth over all downloads.

As I've commented on Discord, that is not true. It would be true in the naive model of 0 latency, streaming from a single machine that has all data already in ram and that server is able to saturate client connection. But that is very far from how it works in the real world.

Unbounded concurrency is bad, single is not great either. We should target 2-4 parallel downloads.

p2004a:
> Would it be worth allowing the user to set the limits? Realistically once you have most maps it wouldn't really be a big issue.

I don't think it's a configuration space that matters for user. We should avoid options space to blow up.

burnhamrobertp:
I think this should've been closed by https://github.com/beyond-all-reason/bar-lobby/pull/675

Hectate:
> I think this should've been closed by [#675](https://github.com/beyond-all-reason/bar-lobby/pull/675)

Correct, thanks for noting.
For any future readers, concurrency is gated by upstream work but is implemented here as a value that can be updated in the future when that issue is resolved.

p2004a:
https://github.com/beyond-all-reason/RecoilEngine/pull/3236 needs to be released

Hectate:
Fair point. I'll open a separate ticket that focuses on enabling the feature once in a release.

PR #442, opened 2025-07-16
Queue map downloads with a max limit of 4

antoni-g:
This is a small PR that addresses a minor issue raised, #423 

What this does is use a lightweight `p-queue` package (from sindresorhus, which there are multiple dependencies from in the project already) to add a max concurrent map downloads to `map.store.ts`. This prevents a user from downloading every map and waiting for them to complete simultaneously. This is a very minor performance/ux enhancement.

When a map is queued for download, it is marked into the `isQueued` state. It will be processed FIFO as queued. The UI reflects this by marking the map as `Queued...`. Once maps are done downloading it will pass along the next map and move to a `Downloading...` state.

The change was tested by artificially throttling the node app with `sudo tc qdisc add dev eth0 root tbf  /  rate 128kbit burst 32kbit latency 400ms` to simulate poor internet. Screenshots of the expected behavior are attached below.

There are a few extensibility improvements we can do from here but they also apply to map downloads overall. I think we could extend this in the future by:

1. allow the user to remove maps from the queue
2. allow the user to cancel existing downloads (an improvement that can be made regardless of this PR)
3. Change the size of the opened map downloads to either fit 4 maps or change the limit to 5 to fill out the opened window. Or simply have the window sized to fit.

**Map queued while 4 other maps are already downloading**
queued-testing-screenshot

**Queued map moved to downloading once the first in the queue has completed**
moved-to-download-testing-screenshot

### Contributor Checklist (**remove after completing**)

- [X] I have read the [**CONTRIBUTING.md**](CONTRIBUTING.md) file.
- [X] I've scoped pull request to a single change or feature.
- [X] I have tested my changes locally.
- [X] I have updated the documentation, if necessary.


bcdrme:
Hi there, thank you for your contribution !

A UX problem I see with this implementation is you'd have to wait for everything in front of the map you wanna play to finish downloading. I'd suggest implementing some sort of prioritisation. So not a queue but a FIFO.

That'd mean being able to pause some downloads to prioritise the latest thing.

bcdrme:
A stack, the latest thing you asked the client to download is the most important thing to download. In a mulitplayer lobby settting, you wouldn't want to wait for players to download 5 maps in the queue THEN the map that the lobby is set up with.

antoni-g:
I see the logic in enabling both.

Actually, your comment though raises what I would like to work on next - the ability for users to pause and stop and cancel downloads. Right now they are started and just go.

Let me know what y'all think about this PR. Can change to a stack, merge, and work on adding user controls around downloads next.

antoni-g:
Sounds good. I'll reapproach this from that perspective.

To clarify, do we want to limit to a single concurrent download? In the original issue I linked, it was discussed to do something like 4-6. I'm trying to figure out what the shape of this feature could look like then.

For context, the current impl is quite barebones; you can queue up any number of maps and they will just download in parallel. While it isn't a huge hit, I think folks with slower internet may experience a giant concurrent block of maps waiting to download.

p2004a:
People do want to download multiple things at the same time. Often e.g. all maps because they have slow internet connection. It's a valid use case to cover, so cancelling last download sounds like invalid behavior to me.

Hectate:
I'm going to close this PR as stale. There is now queue work being done in #635 which won't have merge conflicts to resolve.

PR #449, opened 2025-07-24
Replace hard coded strings with translations for maps files (#433)

wrath625:
Hardcoded string replacement for @webdevsavvy's issue #433 specifically for the maps views


webdevsavvy:
Nice work and thanks for the contribution!

PR #452, opened 2025-07-27
WASM replay parser

bcdrme:
Here is a PoC for a high performance replay parser (WASM, written in Zig). You can find my ugly undocumented code [here](https://github.com/bcdrme/zig-bar-replay-parser).

I'm able to open and parse 645MB worth of replays in 
- 14s if the workload isn't parallelized
- ~1.4s if it, 16 workers on my 7800x3d
- tens of minutes with the current typescript replay-parser (I didn't wait for it to finish tbh, I may or may not do a proper benchmark).

_This PR is WIP, I'm opening it to start the conversation._


p2004a:
The performance numbers look nice of course, but:
1. What is the primary motivation?
2. Are you comparing apples to oranges in terms of functionality? Are they doing the same?
3. Did you try to figure out why the current parser is so slow?

I would personally quite strongly discourage introducing more programming languages in the project.

p2004a:
I've looked into what you parse out in Zig and done some quick improvements to [typescript parser](https://github.com/beyond-all-reason/demo-parser/), result on single threaded parsing on my laptop:

```console
$ du ../replays -BMB
647MB	../replays
$ time NODE_ENV=production node dist/parsing_bench.js
210/210
NODE_ENV=production node dist/parsing_bench.js  12.50s user 1.24s system 105% cpu 12.967 total
```

that in my opinion kills this proposal.

bcdrme:
Great ! Looking forward to test that

PR #458, opened 2025-07-30
Fetch and display dynamic matchmaking queues (#392)

### LINKED ISSUE #392: Display dynamic matchmaking queues
The matchmaking screen currently has 3 hardcoded queues, but that list should instead come from the server with the [matchmaking/list](https://github.com/beyond-all-reason/tachyon/blob/master/docs/schema/matchmaking.md#list) request/response.

wrath625:
Added fetchAvailableQueues which retrieves the matchmaking/list data.  This is called in window.tachyon.onConnected.  Maybe there's a better place for this but it seemed appropriate.

ranked.vue displays the dynamic playlist data now, and sorts it by total player count

All of the queues are displayed using the 'classic' background image defined in css currently as it doesn't look like there are specific background images for multiplayer yet.

Closes #392

### DIFF
--- a/lang/en/lobby.json
+++ b/lang/en/lobby.json
@@ -26,11 +26,8 @@
             "ranked": {
                 "title": "Ranked",
                 "description": "Join a multiplayer ranked queue.",
-                "modes": {
-                    "2v2": "2 vs 2",
-                    "1v1": "DUEL",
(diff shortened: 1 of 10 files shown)


bcdrme:
Cool ! Just fix the warning on the any value and we good :)

Issue #461, opened 2025-08-01
"Sort By" has no effect on the map list


Hectate:
While working on the i18n for [src/renderer/components/maps/MapListComponent.vue](https://github.com/beyond-all-reason/bar-lobby/blob/master/src/renderer/components/maps/MapListComponent.vue) I noted that the "Sort By" choice has no effect on the list of maps displayed. I confirmed this was the case in master, and not caused by my current work.

PR #463, opened 2025-08-02
Quick fix for broken sorting in the MapListComponent

akluu:
**Fix for #461** 

The `dbkey` field name for the sort methods was incorrect. Causing the query to fail and probably returned some unsorted results. After checking the schema in  `map-metadata.ts`, fixing the fieldname fixed the issue.


p2004a:
Thnaks!

PR #465, opened 2025-08-02
Fix tachyon-client request/response binding.

bcdrme:
It was not using the generated tracking uuid for binding a response handler to a specific request.


p2004a (commented):
Sorry for delay... no email notifications for events like "marked this pull request as ready for review"... just missed it :face_with_diagonal_mouth:

The one case that's still missing here is cancelling of ongoing requests and cleaning queue in case of socket disconnection. Up to you to handle it in this PR or do a separate one/file issue.

bcdrme:
I now purge handlers when the socket is closed. I'm not sure I see a use case for cancelling requests though... I'll merge like that, happy to discuss that cancelling request thing elsewhere.

p2004a:
Ah sorry it was just me not being clear, by canceling I had in mind exactly calling the handlers with `socket_closed` before cleaning :)

PR #469, opened 2025-08-09
added a proof-of-concept fix where the game would freeze when clicking on the 'open replays folder' button in the replays section, without a filemanager installed (fix only works for i3)

### LINKED ISSUE #123: Integrate common SPADS battle commands
image

TeamSize: `!set teamSize 4`
#Teams: `!nbTeams 5`
Preset: `!preset coop` (dynamically fetch presets from autohost?)
Autobalance: `!autobalance advanced` (on/off/advanced)
BalanceMode: `!balanceMode random` (skill, clan;skill, random)
Locked: `!lock`, `!unlock`
Balance: `!balance`

ChocoboCoding:
Please describe your changes clearly and concisely. Focus on the "what" and "why" rather than the "how." If your changes are related to an existing issue, please reference it by number (e.g., "Fixes #123").

I've added a functionality to shell:openReplaysDir, where it would check if it is running in I3, and if no filemanager is installed. If so, it will spawn a terminal detached from the bar-lobby application, so that electron won't wait for it to close. 
Additionally I've added 3 functions, to check if it is running in I3, if a filemanager is installed and to focus the terminal into the front, so that it is not stuck behind bar-lobby

### DIFF
--- a/src/main/services/shell.service.ts
+++ b/src/main/services/shell.service.ts
@@ -6,9 +6,80 @@ import { STATE_PATH, CONFIG_PATH, WRITE_DATA_PATH, REPLAYS_PATH, ASSETS_PATH } f
 import { shell } from "electron";
 import { ipcMain } from "@main/typed-ipc";
 import path from "path";
+import { execSync, spawn, exec } from 'child_process';
 
 const REPLAY_SERVICE_URL = "https://bar-rts.com/replays";
 const NEWS_SERVICE_URL = "https://www.beyondallreason.info/news";
+
+// Check for file managers in PATH
+function hasFileManager() {
+    const managers = ['nautilus', 'thunar', 'pcmanfm', 'dolphin', 'nemo'];
+    return managers.some(cmd => {
+        try {
+            execSync(`command -v ${cmd}`, { stdio: 'ignore' });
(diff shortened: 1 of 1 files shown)


p2004a:
Thanks for looking into this, but this looks very hacky, for very specific configuration.

There must be a way to make this generic and not include code like that. All the interesting code is in https://github.com/electron/electron/blob/d6c0691a63c04ca3ffe88154ded595b763a69327/shell/common/platform_util_linux.cc#L270-L368. It looks like we only need to ensure we aren't triggering functions that trigger `XDGUtil` with `wait_for_exit=true`.

There is already also upstream issue for this https://github.com/electron/electron/issues/47668

One of the functions that triggers `wait_for_exit=false` is OpenExternal, we can abuse it. Please check if my workaround on https://github.com/beyond-all-reason/bar-lobby/tree/shell-open-freeze-workaround fixes the issue for you.

p2004a:
Closing because latest electron release has a proper fix for that https://github.com/electron/electron/releases/tag/v37.3.1 and I've pushed 2d34d15aa01dd14c547783ad2cc284d43974dc33 that updates this to that version.

Issue #468, opened 2025-08-09
Replays do not always sort properly over multiple pages


Hectate:
Pagination on the Watch/Replay view does not seem to always properly respect the requested sorting order. Specifically, while the results shown in a single page appear to be sorted properly, if you move forward/backwards in the page count you can sometimes find replays that should have appeared earlier in the list (e.g. a replay at the top of page 2 might have a date more recent than the last replay on page 1, despite being sorted by startTime).

While investigating errors in the console related to the issue, it was identified that the [integrated demo-parser code](https://github.com/beyond-all-reason/bar-lobby/tree/master/vendor/sdfz-demo-parser) is in need of an update. The intent is to integrate the parser from https://github.com/beyond-all-reason/demo-parser/ which will resolve the errors. It is possible that the sorting issues are related to this, so we will need to revisit this issue once that change is implemented.

Example images provided by Aurora on Discord.

Page 1
<img width="1108" height="548" alt="Image" src="https://github.com/user-attachments/assets/3de75163-4da8-4e8f-a09a-990e342b8c49" />

Page 2
<img width="1118" height="549" alt="Image" src="https://github.com/user-attachments/assets/434b26bf-12c0-448f-a82e-7b3f2b87487b" />

As can be seen, we would expect the 2025/08/08 demo on P2 to appear on P1 instead above the 2025/08/06 demo, among other changes in the sorting.

PR #477, opened 2025-09-12
Replays Pagination Sorting Fix

taylorjohannsen:
Fixes: https://github.com/beyond-all-reason/bar-lobby/issues/468

Issue was identified with pagination on the replays page, where the sorting was not working correctly.
This PR address the issue due to `.offset` and `.limit` being called before sorting, which causes the sorting to only be applied to the 15 replays it was filtered to. 
Updated the logic and cleaned up the code a bit here to get it working.

Also, replaced the Demo Parser with it's NPM version (https://github.com/beyond-all-reason/demo-parser/). I can reverse this if this is not correct.

### Contributor Checklist (**remove after completing**)

- [x] I have read the [**CONTRIBUTING.md**](CONTRIBUTING.md) file.
- [x] I've scoped pull request to a single change or feature.
- [x] I have tested my changes locally.
- [x] I have updated the documentation, if necessary.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,6 +35,7 @@
                 "pino-pretty": "^13.0.0",
                 "primeicons": "^6.0.1",
                 "primevue": "3.23.0",
+                "sdfz-demo-parser": "^5.12.0",
                 "tachyon-protocol": "^1.12.0",
                 "vue-i18n": "^11.1.11",
                 "vue-router": "^4.5.1",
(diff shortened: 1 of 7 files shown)


p2004a:
Please revert the demo parser dependency change for now. We want to do this, but the upstream demo parser package is not ready: it pulls in too many dependencies.

Yes, upstream should have been fixed instead of vendoring, vendoring was supposed to be temporary to unblock things, but here we are.

taylorjohannsen:
@p2004a Reverted NPM demo parser import 👍

Issue #307, opened 2025-02-16
Custom battles support


p2004a:
Overall issue for support of custom battles.

Currently blocked entirely by tachyon protocol definition and protocol specification.

PR #486, opened 2025-09-28
Adds working lobbies for online games via Tachyon Protocol

Hectate:
Initial implementation, closes or partially progresses #307 

# Notable Changes
- Significant updates to ``tachyon.store.ts`` to support the new requests, responses, and events, and related logic to support client usage of online lobbies.
- ``OfflineBattleComponent`` was renamed to ``BattleComponent``. 
- ``customLobbies.vue`` ~~was moved into folder alongside the dev-only ``lobby.vue``~~ (reverted change), and received significant updates to utilize the changes to the Tachyon store.
- ``hostBattle.vue`` modal was updated to provide configurable options as needed for a ``lobby/createLobby`` request.
- Components found in ``components/battle/`` have been made "online-aware" to function correctly both online and offline, and/or online-specific components were created.

# Notes
This uses the most current version of tachyon-protocol. As a result, there are UI elements that do nothing because there is no support for them yet on the server side. This includes:
* ~~Any modifications to the lobby settings after creation (map, startboxes, team compositions, etc)~~ **Feature added**
* ~~Joining specific teams, spectating, or queuing on demand~~ **Feature added.**
* Chat messages
* No ability to "Ready Up" or indicate that you are ready to start the battle

No UI team approval was sought for this MVP, the goal is simply to prove functionality. Existing components were used wherever possible, however, to ensure some consistency.

There are a handful of items impacted by dev mode being enabled or not.

# Known Issues
- ~~It's possible to navigate to the Custom Lobbies screen without seeing a Login button, or being logged in. Need to add some sort of status checking to force a user to login first. (fix reverted due to main menu alterations, see changes needed in #498 for the final requirements)~~ **Fixed**
- ~~Playing a custom offline battle while in an online lobby will de-synchronize the state of the UI until a new ``lobby/join`` request is made.~~ **Fixed**
- ~~_Short_ connection losses to the server could de-synchronize the state of the UI. Unsubscribe and then Subscribe to Lobby updates will generally correct this, as will reloading the client (CTRL + SHIFT + R).~~ **Fixed**
- ~~Due to existing issues with the Number component, ``<input>`` tags are in use in the ``HostBattle`` modal.~~ **Fixed**
- The ``lobby/left`` event has only received incidental testing, because there is no support for requesting to kick another member from the lobby.
- Game and Engine versions are not automatically downloaded for the lobby itself, although currently the version used by the server for a lobby is installed by the client during initial launch (See #503 for pending fix)
- A client can receive a ``battleStart`` event even if they have not downloaded the map for the lobby yet (see "Ready Up" note)

### DIFF
--- a/src/renderer/components/battle/BattleComponent.vue
+++ b/src/renderer/components/battle/BattleComponent.vue
@@ -178,7 +178,6 @@ const gameListOptions = computed(() => {
     return Array.from(gameStore.availableGameVersions.values());
 });
 
-
 const map = useDexieLiveQueryWithDeps([() => battleStore.battleOptions.map], () => {
     if (!battleStore.battleOptions.map) return;
     return db.maps.get(battleStore.battleOptions.map.springName);
@@ -205,7 +204,7 @@ async function onGameSelected(gameVersion: string) {
     //FIXME: why do we have both 'gameStore.selectedGameVersion' as well as 'battleStore.battleOptions.gameVersion'??
     //It looks like it's because in offline battles we select from available versions?
     //gameStore.selectedGameVersion = await db.gameVersions.get(gameVersion);
-	gameStore.selectedGameVersion = gameStore.availableGameVersions.get(gameVersion);
+    gameStore.selectedGameVersion = gameStore.availableGameVersions.get(gameVersion);
     battleStore.battleOptions.gameVersion = gameVersion;
 }
 


src/renderer/components/battle/TeamComponent.vue
```diff
@@ -54,37 +62,87 @@ import BotParticipant from "@renderer/components/battle/BotParticipant.vue";
...
+});
+
 const memberCount = computed(() => {
-    return battleWithMetadataStore.teams[props.teamId]?.participants.length || 0;
+    if (battleStore.isOnline && tachyonStore.activeLobby && tachyonStore.activeLobby.allyTeams) {
+        if (tachyonStore.activeLobby.members) {
+            let count = 0;
+            for (const memberKey in tachyonStore.activeLobby.members) {
+                const member = tachyonStore.activeLobby.members[memberKey];
+                if (parseInt(member!.allyTeam) == props.teamId) count++;
```
Hectate, line 101: Need to compare strings, not index and a parsedInt(string).

src/renderer/store/battle.store.ts
```diff
@@ -320,10 +327,10 @@ function removeCustomStartBox(boxId: number) {
     }
 }
 
-function defaultOfflineBattle(engine?: EngineVersion, game?: GameVersion, map?: MapData) {
+function defaultOfflineBattle(engine?: EngineVersion, game?: GameVersion, map?: MapData, online?: boolean) {
```
Hectate, line 330: Rename this for clarity.

src/renderer/store/tachyon.store.ts
```diff
@@ -40,6 +67,221 @@ async function connect() {
...
+        console.error("Error with request lobby/unsubscribeList:", error);
+        tachyonStore.error = "Error with request lobby/unsubscribeList";
+    }
+}
+
+async function createLobby(data: LobbyCreateRequestData) {
+    try {
+        battleActions.resetToDefaultBattle(undefined, undefined, undefined, true);
+        battleStore.isOnline = true;
+        battleStore.isLobbyOpened = true;
```
Hectate, line 98: This will open the BattleDrawer even if the request is not successful. This line should be after the request is made so it does not open when it should not (e.g. on failure).

src/renderer/store/tachyon.store.ts
```diff
@@ -40,6 +67,221 @@ async function connect() {
...
+    };
+    battleStore.battleOptions.engineVersion = data.engineVersion;
+    battleStore.battleOptions.gameVersion = data.gameVersion;
+    db.maps.get(data.mapName).then((map) => {
+        battleStore.battleOptions.map = map;
+    });
+    for (const allyKey in data.allyTeamConfig) {
+        const item = data.allyTeamConfig[allyKey];
+        lobbyObject.allyTeams![allyKey] = {
+            startBox: data.allyTeamConfig[allyKey].startBox,
```
Hectate, line 160: Unclear why I didn't use the ``item`` const here... need to check.

src/renderer/store/tachyon.store.ts
```diff
@@ -40,6 +67,221 @@ async function connect() {
...
+            maxTeams: item.maxTeams,
+            teams: {},
+        };
+        for (const teamKey in item.teams) {
+            const team = item.teams[teamKey];
+            lobbyObject.allyTeams![allyKey].teams![teamKey] = { maxPlayers: team.maxPlayers };
+            lobbyObject.maxPlayerCount += team.maxPlayers; //Lobby maximum is sum of all AllyTeams>Teams>MaxPlayers
+        }
+        //Here we assign the startbox for the AllyTeam to the battlestore so they match what we set when the lobby was created.
+        battleStore.battleOptions.mapOptions.customStartBoxes.push(data.allyTeamConfig[allyKey].startBox);
```
Hectate, line 170: This should have also used the ``item`` const probably.

src/renderer/store/tachyon.store.ts
```diff
@@ -40,6 +67,221 @@ async function connect() {
...
+            type: "player",
+        };
+    }
+    return lobbyObject;
+}
+
+async function leaveLobby() {
+    try {
+        tachyonStore.error = undefined;
+        const response = await window.tachyon.request("lobby/leave");
```
Hectate, line 190: The ``lobby/leave`` request is basically always successful or it fails for reasons that you're not in a lobby anyway. If, for some reason, we have to clear the lobby data _but_ cannot get a success response here, the code below will not clear the status. Move it up so it happens either outside of the try/catch or before the request is sent.

src/renderer/store/tachyon.store.ts
```diff
@@ -40,6 +67,221 @@ async function connect() {
...
+    tachyonStore.activeLobby = applyPatch(tachyonStore.activeLobby, data);
+    //Recalculate player counts afterward.
+    let maxPlayerCount: number = 0;
+    for (const allyKey in tachyonStore.activeLobby!.allyTeams) {
+        for (const teamKey in tachyonStore.activeLobby!.allyTeams[allyKey]!.teams) {
+            maxPlayerCount += tachyonStore.activeLobby!.allyTeams[allyKey]!.teams[teamKey]!.maxPlayers!;
+        }
+    }
+    tachyonStore.activeLobby!.maxPlayerCount = maxPlayerCount;
+    const playerCount: number = Object.keys(tachyonStore.activeLobby!.members!).length;
```
Hectate, line 243: Add a reminder to change this when spectators exist.

src/renderer/store/tachyon.store.ts
```diff
@@ -40,6 +67,221 @@ async function connect() {
...
+    tachyonStore.activeLobby!.playerCount = playerCount;
+    // We need to sub to new members, but if members go away we unsub instead. Ugh we need to maintain a list and diff it also lol
+    if (data.members) {
+        const userSubList: UserId[] = [];
+        for (const memberKey in data.members) {
+            if (data.members[memberKey]) {
```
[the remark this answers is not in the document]
Hectate: Right now this just would resub to them.

src/renderer/views/play/customLobbies/customLobbies.vue
```diff
@@ -0,0 +1,205 @@
+<!--
```
Hectate, line 1: Better to move this back to the original directory. We get a better diff, but more importantly the ``lobby.vue`` file is currently dev-only since we implemented the Battle Drawer setup instead. Maybe later if we end up with a specific view for a lobby (i.e. once UI Team provides the spec) we can review if the structure needs to be separated or not like this.

src/renderer/views/play/customLobbies/lobby.vue
```diff
@@ -0,0 +1,73 @@
...
+</template>
+
+<script lang="ts" setup>
+import Panel from "@renderer/components/common/Panel.vue";
+import Button from "@renderer/components/controls/Button.vue";
+import { tachyonStore, tachyon } from "@renderer/store/tachyon.store";
+import { router } from "@renderer/router";
+import { battleStore } from "@renderer/store/battle.store";
+
+//TODO: This is a dev-only page, but if we are going to keep it we should still apply translation strings
```
Hectate, line 53: Remove; this page will be fully replaced by whatever the UI team provides for spec and translations will happen at that point, or it will be deleted entirely. All functions in here are provided in the Battle Drawer except for the separation of the configuration components.

src/renderer/components/battle/Playerlist.vue
```diff
@@ -14,12 +14,32 @@ SPDX-License-Identifier: MIT
...
         @bot-selected="onBotSelected"
     />
-    <div class="scroll-container padding-right-sm">
+    <div v-if="battleStore.isOnline" class="scroll-container padding-right-sm">
+        <div class="playerlist" :class="{ dragging: draggedBot || draggedPlayer }">
+            <TeamComponent
+                v-for="(team, key, index) in tachyonStore.activeLobby != undefined ? tachyonStore.activeLobby!.allyTeams : {}"
+                :key="key"
+                :teamId="index"
+                :teamKey="key as string"
```
Hectate, line 23: Once tachyon server sends us the allyteam/team with matching strings to the member's properties, we can probably remove this ``as``.

src/renderer/views/play/customLobbies.vue
```diff
@@ -101,36 +117,81 @@ SPDX-License-Identifier: MIT
...
 
 const { t } = useTypedI18n();
 
 const loading = ref(false);
-const hostBattleOpen = ref(false);
+const createLobbyModalIsOpen = ref(false);
+const leaveConfirmModalIsOpen = ref(false);
 const searchVal = ref("");
-const selectedBattle: Ref<OngoingBattle | null> = shallowRef(null);
+const autojoinLobbyId = ref();
```
Hectate, line 144: This should be typed as a string.

src/renderer/components/battle/Playerlist.vue
Hectate: Still required? This seems to be an issue where Vue treats the key as it will be ``string | number`` when we know it is always a string. Will keep the ``as`` here.

src/renderer/views/play/customLobbies.vue
Hectate: Fixed.

Hectate:
All issues found during self-review have been resolved and changes committed.

Hectate:
I'm going to revamp this so that the lobby API code is all in a separate file, instead of in the ``tachyon.store.ts`` file. It'll be cleaner that way.

src/renderer/store/lobby.store.ts
```diff
@@ -0,0 +1,288 @@
...
+
+const i18n = setupI18n();
+
+export const lobbyStore = reactive({
+    isInitialized: false,
+    error: undefined,
+    lobbyList: {}, //This will hold changes from ``lobby/listUpdated`` events
+    selectedLobby: {}, //This is the lobby we select from the datatable showing the lobbylist
+    activeLobby: undefined, //This will hold changes from ``lobby/updated`` events
+} as {
```
p2004a, line 34: Prefer how it's done in engine and map store please

src/renderer/store/lobby.store.ts
```diff
@@ -0,0 +1,288 @@
...
+import { battleStore, battleActions } from "@renderer/store/battle.store";
+import { db } from "@renderer/store/db";
+import { notificationsApi } from "@renderer/api/notifications";
+import { setupI18n } from "@renderer/i18n";
+
+const i18n = setupI18n();
+
+export const lobbyStore = reactive({
+    isInitialized: false,
+    error: undefined,
```
p2004a, line 30: I don't think I understand the concept of single global error. All the error handling below is also just mostly setting this value and I believe nothing happens with it, and it's just overridden in next call?

src/renderer/store/lobby.store.ts
```diff
@@ -0,0 +1,288 @@
...
+}
+
+function clearLobbyAndListInfo() {
+    lobbyStore.lobbyList = {};
+    lobbyStore.selectedLobby = null;
+    lobbyStore.activeLobby = undefined;
+}
+
+export async function initLobbyStore() {
+    window.tachyon.onEvent("lobby/left", (data) => {
```
p2004a, line 272: nit pick: it could be just `window.tachyon.onEvent("lobby/left", onLobbyLeftEvent)`

src/renderer/store/lobby.store.ts
```diff
@@ -0,0 +1,288 @@
...
+        maxPlayerCount: 0, //Placeholder value; Ditto for calculating max players
+        mapName: data.mapName,
+        engineVersion: data.engineVersion,
+        gameVersion: data.gameVersion,
+        allyTeams: {},
+        members: {},
+    };
+    battleStore.battleOptions.engineVersion = data.engineVersion;
+    battleStore.battleOptions.gameVersion = data.gameVersion;
+    db.maps.get(data.mapName).then((map) => {
```
p2004a, line 125: Assert for when it's not found?

Overall, the more I stare at the map store the more I think we need to attempt to rework it again to have better separation from live maps list and not have all tied to it... But that's not for this PR...

src/renderer/model/lobby.ts
```diff
@@ -0,0 +1,41 @@
...
+// With changes this is now just the same as the tachyon type for LobbyJoinOkResponseData | LobbyCreateOkResponseData
+export type Lobby = {
+    id: string;
+    name: string;
+    mapName: string;
+    engineVersion: string;
+    gameVersion: string;
+    playerCount: number;
+    maxPlayerCount: number;
+    allyTeams?: {
```
p2004a, line 18: is there good reason for this to be optional property? similar members etc below

src/renderer/model/lobby.ts
```diff
@@ -0,0 +1,41 @@
...
+        [k: string]: {
+            startBox?: StartBox;
+            maxTeams?: number;
+            teams: {
+                [k: string]: {
+                    maxPlayers?: number;
+                } | null;
```
p2004a, line 25: similar to above, why it's ok for this to be nullable?

src/renderer/store/lobby.store.ts
```diff
@@ -0,0 +1,288 @@
...
+    // We need to sub to new members, but if members go away we unsub instead. Ugh we need to maintain a list and diff it also lol
+    if (data.members) {
+        const userSubList: UserId[] = [];
+        const userUnsubList: UserId[] = [];
+        for (const memberKey in data.members) {
+            const member = data.members[memberKey];
+            if (member != null) {
+                userSubList.push(member.id);
+            } else {
+                userUnsubList.push(memberKey);
```
p2004a, line 226: Annoying, this can unsubscribe from user to whom we subscribed for some other reason, e.g. friend list, so looks like we need to maintain some single global list of who we need to subscribe unsubscribe too e.g. with kind of "reference count".

src/renderer/store/tachyon.store.ts
```diff
@@ -94,6 +99,8 @@ export async function initTachyonStore() {
         if (me.isAuthenticated) {
             // Try to connect to Tachyon server periodically
             tachyonStore.reconnectInterval = setInterval(connect, 10000);
+            // We also want to clear the lobby list, etc
+            lobby.clearLobbyAndListInfo();
```
p2004a, line 103: This one is tricky. Maybe it's ok for now, but not finally.

Effectively, the lobby view for UI, and lobby view from protocol perspective will need to be separated for disconnect support. When websocket connection drop, maybe there could be some spinner or something, but overall the user in UI should still stay in lobby and so on and we reconcile only after reconnection.

It might be a bit too much to implement it also as part of this PR, but we will need to bite this problem at some point.

p2004a (commented):
For now just quite surface level reading of parts of the PR...

src/renderer/store/lobby.store.ts
Hectate: Yeah, and this is relevant for parties, matchmaking, and friends too. I’m thinking we can keep it as part of another store (for example Me or Friends) and just do a check from here, as well as any of those other places.

src/renderer/store/tachyon.store.ts
Hectate: Agreed. I noted it as a known issue but am undecided on how best to handle it.

src/renderer/model/lobby.ts
Hectate: Both properties are duplicating the response data for create/join in the protocol. I’ll have to look of I made an error, or if there are reasons for these to be absent. I can’t immediately think of a reason why. A lobby should have at least 1 allyteam with 1 or more team per allyteam.

src/renderer/store/lobby.store.ts
Hectate: Speaking of; once the update to the friends feature is merged, I was going to take a look at this again. It wasn’t clear where we would end up with either yet. I needed to see the final implementation to figure out the approach.

src/renderer/store/lobby.store.ts
Hectate: We could do something here, yes. If we somehow create or join a lobby and get the success response with an unidentified map name, then the only reasonable response is to bail on the lobby altogether? It should be pretty rare, I hope.
Do we have a way to essentially say “I think my maps list may be inaccurate or out of date, let’s refresh it”?

src/renderer/model/lobby.ts
Hectate: I see that I used the LobbyUpdatedEventData as the base, which has optional properties everywhere because of the patch used.  I'm updating it to the latest tachyon spec anyway, so I'll fix this too.

src/renderer/store/lobby.store.ts
Hectate: So I ran into a "command_unimplemented" error (``lobby/joinAllyTea``) and it did a full stop of the UI with the giant red error box. It looks like the "catch" isn't even happening because the ``tachyon-client`` is treating this as an unhandled rejection before it even gets to ``lobby.store``.

```
[23:58:40.087] ERROR: main - 2025-10-05T03:57:35.361Z - tachyon-client - Error response received: {"reason":"command_unimplemented","status":"failed","type":"response","commandId":"lobby/joinAllyTeam","messageId":"4d2aa29c-030c-49d8-8864-e0cd069c4d65"}
Error occurred in handler for 'tachyon:request': Error: command_unimplemented
    at <anonymous> (D:\git\bar-lobby\src\main\tachyon\tachyon-client.ts:142:28)
    at TachyonClient.handleResponse (D:\git\bar-lobby\src\main\tachyon\tachyon-client.ts:199:9)
    at TachyonClient.handleMessage (D:\git\bar-lobby\src\main\tachyon\tachyon-client.ts:166:18)
    at WebSocket2.<anonymous> (D:\git\bar-lobby\src\main\tachyon\tachyon-client.ts:67:26)
    at callListener (D:\git\bar-lobby\node_modules\ws\lib\event-target.js:290:14)
    at WebSocket2.onMessage (D:\git\bar-lobby\node_modules\ws\lib\event-target.js:209:9)
    at WebSocket2.emit (node:events:518:28)
    at Receiver.receiverOnMessage (D:\git\bar-lobby\node_modules\ws\lib\websocket.js:1220:20)
    at Receiver.emit (node:events:518:28)
    at Receiver.dataMessage (D:\git\bar-lobby\node_modules\ws\lib\receiver.js:596:14)
[23:58:40.151] ERROR: renderer - 2025-10-05T03:58:40.151Z - Error.vue - Error: Error invoking remote method 'tachyon:request': Error: command_unimplemented
```

This is from me trying it like so:
```ts
async function joinAllyTeam(allyTeam: string) {
    try {
        window.tachyon.request("lobby/joinAllyTeam", { allyTeam: allyTeam });
    } catch (error) {
        console.error("Tachyon error: lobby/joinAllyTeam:", error);
        notificationsApi.alert({
            text: "Error with request lobby/joinAllyTeam",
            severity: "error",
        });
    }
}
```
So we'll probably need to check into bubbling these errors up as non-fatal errors that essentially crash the client even though they really don't need to.

p2004a: Looks like it's just missing `await` in `tachyon.request` call.

Hectate: Ah interesting. Thanks. I'll add that to all of them just in case. With the notificationAPI additions too, we actually get alerts in the UI now too. So this will be super helpful. The actual details will be in the console(s).

Later, after we finish this, we can go back and update the other stores to do the same.

src/renderer/store/lobby.store.ts
Hectate: This is resolved. If the `db.maps.get()`` returns ``undefined``, a console error and notification will be generated. The user will still join the lobby, however. Since an ``undefined`` mapname cannot be downloaded, instead the user will see a disabled "Start the game" button and a blank map preview.
To resolve, they'll have to leave the lobby. Once all users leave, the lobby, along with the undefined mapname, will be disposed of. Ideally, proper lobby creation will prevent this situation from happening, but at least it will be handled gracefully now.

src/renderer/store/tachyon.store.ts
Hectate: This should be resolved, and will be better for the entire client. Your comment about separating UI and protocol made me look closer at what I had to work with.

Upon connection loss, the following happens:
1. NotificationAPI Alerts pop up advising the user. No more "crashes" with Unhandled Exceptions upon ``tachyon.store`` Errors thrown.
2.  Custom Lobbies view specifically will go into a "Loading" state, and the Battle Drawer will close if necessary.
3. After a few seconds, if reauth/reconnect fails, the user will be put "offline" (existing behavior).
4. For any routes that have the ``meta.onlineOnly`` property, the user will be sent back to the login/offline route ``"/"`` so they can't interact with online-only views.
5. If the user reconnects before (3) and (4) above instead, the client will attempt to restore lobby state by reading the ``user.currentLobby`` property from the ``user/self`` event that is sent automatically upon connection/reconnection.
- (5a): If ``currentLobby`` is present, a ``lobby/join`` request will be sent for that lobby ID. By design, the protocol states that this will auto-succeed if the user is in the lobby already. The result will be a full parse of the lobby data upon success, restoring the ``lobbyStore.activeLobby`` state. _Note: this property is always ``null`` in the event data at this time, so this requires server changes to work correctly._
- (5b): If a lobby ID is not provided, the lobby information is purged and then a ``lobby/subscribeList`` request is sent. This triggers the server to send a ``lobby/resetList`` event, which will fully restore the ``tachyonStore.lobbyList`` state, as well as ensure further updates are sent.

src/renderer/store/lobby.store.ts
Hectate: Tell me what you think of this idea before I start writing the logic. This is also the first overloads and JSDocs I've done, but I felt to maximize future usefulness, it was appropriate.

Inside ``me.store.ts`` we export ``subscriptions`` which contains ``attach``, ``detach``, and two different getters. Elsewhere in the client (e.g. ``lobby.store.ts``) we use these to add/remove users to specific subscription lists. Meanwhile, the ``me.store.ts`` will actually manage the logic to decide if we need to subscribe, unsubscribe, etc. based on the changes made to the overall list.

For example, if user ``"123"`` is in both ``CURRENTFRIEND`` and ``LOBBYUSER``, and we do ``detach("123", SubLists.LOBBYUSER);``, no unsubscribe will be sent because the overall list still contains "Hectate" inside the Friend list.

I've written the initial framework already, but I didn't want to go into the logic in case you saw something useful. If we get new categories of lists to maintain, we simply add to the Enum. The ``IGNOREUSER`` probably won't end up used in this context, but it's there for now.

```ts
export enum SubLists {
    OUTGOINGFRIEND,
    INCOMINGFRIEND,
    CURRENTFRIEND,
    IGNOREUSER,
    LOBBYUSER,
    PARTYUSER,
    MATCHUSER,
}
export const subscriptions = { attach, detach, getUsersInSublist, getSublistsFromUsers };

/**
 * Attach the UserId(s) to the selected subscription list. A subscription request will be automatically sent to the Tachyon server if required.
 * @param users UserId, or Array of UserId, that will be attached
 * @param list SubLists Enum indicating which subscription list to attach them to.
 */
function attach(users: UserId, list: SubLists): void;
function attach(users: UserId[], list: SubLists): void;
function attach(users: UserId | UserId[], list:SubLists): void {
	if(typeof users === 'string') {
		console.log("Attaching single user to ", list);
	} else {
		console.log("Adding multple users to ", list);
	}
}

/**
 * Detach the UserId(s) from the selected subscription list. An unsubscribe request will be automatically sent to the Tachyon Server if required.
 * @param users UserId, or Array of UserId, that will be detached
 * @param list SubLists Enum indicating which subscription list to detach them from
 */
function detach(users: UserId, list: SubLists): void;
function detach(users: UserId[], list: SubLists): void;
function detach(users: UserId | UserId[], list: SubLists): void {

}

/**
 * Provides all UserId(s) currently part of the subscription list.
 * @param list Enum indicating which subscription list to get the UserIds for.
 * @returns An Array of UserId, or an empty Array if the list has none.
 */
function getUsersInSublist(list: SubLists): UserId[] {

}

/**
 * Provides all Sublists the identified UserId(s) are currently attached to.
 * @param users UserId, or Array of UserId
 * @returns An Array of Enums, or an empty Array if the UserId(s) are in none of the lists.
 */
function getSublistsFromUsers(users: UserId): SubLists[];
function getSublistsFromUsers(users: UserId[]): SubLists[];
function getSublistsFromUsers(users: UserId | UserId[]): SubLists[] {

}

// Example
if(getSublistsFromUsers("123").includes(SubLists.CURRENTFRIEND)) {
	console.log("User 123 is a friend.");
}
```

Hectate: This is now resolved with the changes to utilize the new ``SubsManager`` class.

src/renderer/components/battle/AddBotModal.vue
```diff
@@ -46,11 +46,11 @@ const { t } = useTypedI18n();
 const props = defineProps<{
     engineVersion: string;
     gameVersion: string;
-    teamId: number;
+    teamId: number | string;
```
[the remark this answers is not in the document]
Hectate: With refactoring, I’m sure it’s possible. My initial plan was to just get all this functional so we can test against it. Then later revisit with UI later to see what kind of overlap needs to exist between all the different, possibly shared, components and client UI.
The deeper we get, however, the more these two are getting intertwined - and that might be undesired and we just don’t know yet because these elements haven’t been defined.
I have noted that prior development incurred assumption about how client<>server would function, all prior to tachyon developments. There’s a whole host of refactoring that will be needed, and it would be worth discussing in depth, perhaps as part of future Milestone conversation.

Hectate:
Notable updates made recently:
* Addition of the Lobby "preview" in the NavBar space
* Custom Lobbies view is closer (not complete) to the UI Team spec in #494 
* Bot configs are no longer subject to desync issues.

Hectate:
Closing PR per conversation in Discord. Portions of this branch will be successively prepared for separate, smaller pull requests.

PR #488, opened 2025-10-03
Matchmaking updated and cleaned up

Hectate:
This PR updates client to the Tachyon 1.14.2 specification, and updates ``matchmaking.store.ts``. The file has also been cleaned up to be more readable, maintainable, and ready for future development.

Note that this does not implement any functionality that was not already present, nor fix any functionality that was previously unimplemented. This is just laying the groundwork for UI-side advancement of Matchmaking, as well as matchmaking client logic.

As an aside, the main issue from matchmaking being functional is ensuring the player has the map for their match downloaded before needing to Ready.

### DIFF
--- a/lang/en/lobby.json
+++ b/lang/en/lobby.json
@@ -1,13 +1,83 @@
 {
     "lobby": {
+        "api": {
+            "commands": {
+                "help": "Displays this help text.",
+                "whoami": "Sends back information about who you are.",
+                "whois": "[user] Sends back information about the user specified.",
+                "discord": "Allows linking of your discord account to your BAR account.",
+                "mute": "[username] Mutes that user and prevents you seeing their messages.",
+                "unmute": "[username] Un-mutes that user and allows you to see their messages.",
+                "coc": "[term] Searches the code of conduct and returns items with a textual match in them.",
+                "joinq": "Adds you to the queue to join when a space opens up, you will be automatically added to the game as a player. If already a member it has no effect.",
+                "leaveq": "Removes you from the join queue.",
+                "status": "Status info about the battle lobby.",
+                "afks": "Lists possible afk players.",
+                "password": "Tells you the room password",
+                "splitLobby": "[minimum players] Causes a vote to start where other players can elect to join you in splitting the lobby, follow someone of their choosing or remain in place. After 30 seconds, if at least the minimum number of players agreed to split, you are moved to a new (empty) lobby and those that voted yes or are following someone that voted yes are also moved to that lobby.",
+                "roll": "[range] Rolls a random number based on the range format.\n- Dice format: nDs, where n is number of dice and s is sides of die. E.g. 4D6 - 4 dice with 6 sides are rolled\n- Max format: N, where N is a number and an integer between 1 and N is returned\n- Min/Max format: MN MX, where each is a number and an integer between them (inclusive) is returned",
+                "explain": "Lists a log of the steps taken to calculate balance for the lobby",
+                "resetApproval": "Resets the list of approved players to just the ones present at the moment (approved players are able to join even if it is locked and without needing a password). Requires boss privileges.",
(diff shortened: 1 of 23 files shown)


Hectate:
Pulling back to draft so feedback from the lobby store PR can be incorporated here as well.

Hectate:
Cleanup done, ready for review.

src/renderer/store/matchmaking.store.ts
```diff
@@ -31,88 +48,155 @@ export const matchmakingStore = reactive<{
...
+                matchmakingStore.downloadsRequired[queue.id] = { needed: checkIfAnyMapsAreNeeded(queue.maps) };
+            } else {
+                matchmakingStore.downloadsRequired[queue.id] = { needed: true };
+            }
+        });
     } catch (error) {
-        console.error("Failed to fetch available queues", error);
-        matchmakingStore.queueError = "Failed to fetch queues";
+        console.error("Tachyon error: matchmaking/list:", error);
+        matchmakingStore.queueError = "Failed to retrieve available queues";
```
p2004a: notification here?

src/renderer/store/matchmaking.store.ts
```diff
@@ -31,88 +48,155 @@ export const matchmakingStore = reactive<{
...
-            console.error("Error starting matchmaking:", error);
-            matchmakingStore.errorMessage = "Error: Failed to join the matchmaking queue.";
-            matchmakingStore.status = MatchmakingStatus.Idle;
-        }
-    },
-    stopSearch() {
+async function sendQueueRequest() {
+    try {
+        matchmakingStore.errorMessage = null;
+        matchmakingStore.status = MatchmakingStatus.Searching;
```
p2004a: why set it both here and in `onQueuesJoinedEvent`?

p2004a (changes requested):

src/renderer/store/matchmaking.store.ts
Hectate: Technically, I think my plan when I did that was to have the UI start to treat us as in queue "early" rather than wait for the event, and then if the request failed we would just go back to Idle instead. But considering the possible failure points, I'm going to add a "Requested Join" status that permits us to be between the two.

p2004a (commented):
Nice, looks good, one small nit :)

src/renderer/store/matchmaking.store.ts
```diff
@@ -31,88 +48,155 @@ export const matchmakingStore = reactive<{
...
-        matchmakingStore.queueError = "Failed to fetch queues";
+        console.error("Tachyon error: matchmaking/list:", error);
+        matchmakingStore.queueError = "Failed to retrieve available queues";
     } finally {
         matchmakingStore.isLoadingQueues = false;
     }
 }
 
+// This is just to get is a quick idea if any maps will be needed to be downloaded
+function checkIfAnyMapsAreNeeded(maps: { springName: string }[]): boolean {
```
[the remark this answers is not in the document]
Hectate: Ok, I can work with that. Once #503 is merged with the ability to attempt to obtain “unknown” maps, we should have nothing that prevents that from working on this side too. Custom Lobbies can follow along also in the same vein.

src/renderer/store/matchmaking.store.ts
```diff
@@ -31,88 +49,150 @@ export const matchmakingStore = reactive<{
...
-            await window.tachyon.request("matchmaking/queue", { queues: [matchmakingStore.selectedQueue] });
-        } catch (error) {
-            console.error("Error starting matchmaking:", error);
-            matchmakingStore.errorMessage = "Error: Failed to join the matchmaking queue.";
-            matchmakingStore.status = MatchmakingStatus.Idle;
-        }
-    },
-    stopSearch() {
+async function sendQueueRequest() {
+    if (matchmakingStore.downloadsRequired[matchmakingStore.selectedQueue].needed) {
```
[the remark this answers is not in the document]
Hectate: Done. I also took the opportunity to reset the ``downloadsRequired`` object entirely whenever the client reloads the playlist, to ensure that the two are synchronized and "stale" entries aren't left inside it.

PR #492, opened 2025-10-09
User subscriptions API for me.store.ts

Hectate:
@p2004a FYI: seeking feedback

Initial pass at a concept of managing all possible user subscriptions from a tachyon server in a central place for the client. Changes are not yet tested, this is simply a initial review.

The concept is that anywhere else in the client, we can request adding or removing a UserId to one of the valid subscription sets/lists in ``me.store.ts``, all without worrying about the actual underlying subscription itself. This will prevent losing subscriptions, for example, if you "Friend" a user in your Lobby and then leave that Lobby afterward. There are some convenience getters also.

I'm not satisfied with how I have ``SubLists`` and ``meSets`` in this configuration, as it would be more ideal for them to either be merged into a single object so referencing a particular set (via the Enum) did not rely on that set being a particular index of an Array that matches that Enum's number.

### DIFF
--- a/src/renderer/utils/subscriptions-manager.ts
+++ b/src/renderer/utils/subscriptions-manager.ts
@@ -40,12 +40,9 @@ export class SubsManager {
         const userUnsubs: UserId[] = [];
         for (const user of typeof users === "string" ? [users] : users) {
             const uLists: Set<symbol> | undefined = this.userList.get(user);
-            if (uLists != undefined) {
-                uLists.delete(list);
-                if (uLists.size == 0) {
-                    userUnsubs.push(user);
-                    this.userList.delete(user);
-                }
+            if (uLists?.delete(list) && uLists.size === 0) {
+                userUnsubs.push(user);
+                this.userList.delete(user);
             }
         }
         if (userUnsubs.length > 0) {
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,4 @@ thumbs.db
 debug.log
 bun.lockb
 *.pfx
+*.ps1
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,7 +35,7 @@
                 "pino-pretty": "^13.0.0",
                 "primeicons": "^6.0.1",
                 "primevue": "3.23.0",
-                "tachyon-protocol": "^1.14.3",
+                "tachyon-protocol": "^1.15.1",
                 "vue-i18n": "^11.1.11",
                 "vue-router": "^4.5.1",
                 "ws": "^8.18.3"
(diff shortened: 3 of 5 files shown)


src/renderer/utils/subscriptions-manager.ts
```diff
@@ -0,0 +1,205 @@
...
+            let dirty: boolean = false;
+            let uLists = this.userList.get(user); //get the lists for the user
+            if (!uLists) {
+                //if undefined, there are none for this user
+                uLists = new Set(); //so we create an empty Set for this user
+                uLists.add(list); //we add the requested list to this set
+                userSubs.push(user); //add the user to the subscription list because they are entirely new
+                dirty = true;
+            }
+            if (!uLists.has(list)) {
```
p2004a: no need for this `has`, just add to set

src/renderer/utils/subscriptions-manager.ts
```diff
@@ -0,0 +1,205 @@
...
+                //if this list is missing from this user
+                uLists.add(list); //we add it!
+                dirty = true;
+            }
+            if (dirty) {
+                this.userList.set(user, uLists); //update the user's map with the changes
```
p2004a: do it when you create new `ulist` instead separately, no need to split it and introduce `dirty` bit.

src/renderer/utils/subscriptions-manager.ts
```diff
@@ -0,0 +1,205 @@
...
+     * Detach the UserId(s) from the selected subscription list. An unsubscribe request will be automatically sent to the Tachyon Server if required.
+     * @param users UserId, or Array of UserId, that will be detached
+     * @param list Symbol indicating which subscription list to detach them from
+     */
+    detach(users: UserId, list: symbol): void;
+    detach(users: UserId[], list: symbol): void;
+    detach(users: UserId | UserId[], list: symbol): void {
+        const userUnsubs: UserId[] = [];
+        for (const user of typeof users === "string" ? [users] : users) {
+            let dirty: boolean = false;
```
p2004a: this variable doesn't have any purpose

src/renderer/utils/subscriptions-manager.ts
```diff
@@ -0,0 +1,205 @@
...
+     * Provides all UserIds currently in any subscription list
+     * @returns An Array of UserIds, or an empty Array if there no subscriptions
+     */
+    getAllUsersSubscribed(): UserId[] {
+        const arr: UserId[] = [];
+        for (const [key] of this.userList.entries()) {
+            arr.push(key);
+        }
+        return arr;
+    }
```
p2004a: Just `return Array.from(this.userList.keys())`

src/renderer/utils/subscriptions-manager.ts
```diff
@@ -0,0 +1,205 @@
...
+                dirty = uLists.delete(list); //we delete the requested list
+                if (dirty) {
+                    //if a list was deleted successfully
+                    if (uLists.size == 0) {
+                        //we check if there are any lists left
+                        userUnsubs.push(user); //and add them to unsub if there are none
+                        this.userList.delete(user); //also delete from the list of subscribed users
+                    } else {
+                        //there are lists remaining
+                        this.userList.set(user, uLists); //save the updated set to the user's map
```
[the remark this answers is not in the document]
Hectate: Indeed, I forgot about that. I'll restructure it appropriately.

src/renderer/utils/subscriptions-manager.ts
Hectate: I was using it to determine if I needed to ``set`` again after ``delete``, but as you pointed out below that is unnecessary, so yes it's all redundant work. The ``size`` alone will tell me if it's time to unsub or not. Correcting, thanks!

src/renderer/utils/subscriptions-manager.ts
Hectate: I had overlooked that ``add`` wouldn't duplicate (one of the benefits of a Set!), so no need to pre-check. That definitely simplifies it.

p2004a (approved):
LGTM, thank you! 2 small nits, feel free to ignore them and just resolve comments or adjust, up to you, I'm happy to merge this in anytime :).

PR #496, opened 2025-10-19
Updated main menu to UI Team spec

Hectate:
Main menu now matches desired appearance of large viewport for artwork/3d scenes, with buttons on the bottom right for the main Play actions. There is a small, unrelated change to ``.gitignore`` to exclude Powershell scripts from being uploaded.

Waiting on final UI team approval.

Screenshot:

image

### DIFF
--- a/src/renderer/views/play/menu.vue
+++ b/src/renderer/views/play/menu.vue
@@ -19,21 +19,15 @@ SPDX-License-Identifier: MIT
                         </div>
                     </Panel>
 
-                    <Panel :no-padding="true" class="game-mode-card" @click="startCampaign">
+                    <Panel :no-padding="true" class="game-mode-card disabled" @click="startCampaign">
                         <div class="card-content">
-                            <h2>Campaign</h2>
+                            <h2>Campaign<span class="small-text margin-left-md">(Coming Soon)</span></h2>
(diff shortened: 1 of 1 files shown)


Hectate:
Updated per request. Animation is 0.1s in and 0.3s out. Background color for this is ``rgba(255,255,255,0.4);`` which is pure white with 40% alpha. The way a Panel is component is designed, it includes a black, very transparent, linear gradient which adds shading at the top and bottom. As a result, we get "free" shading when just changing the background like this.

image

p2004a:
> can we instead show players a popup that explains what they need to do in order for the button to work and then provide them with a way to move forward?

No, I'm against that. The purpose is to hide behind dev mode functionality that is not finished. Everything behind multiplayer is at the moment not finished and can't be used on main server etc.

Pop-ups "you need to be online" are fine, but there should never be a case of some popup showing "you need to open dev mode" because by definition of what dev mode is, everything related to it should be hidden without it enabled.

scottsaunders:
image

Recapping a decision. 

For features that are still under development and only accessible in developer mode, we will disable those buttons and add (coming soon).

p2004a (approved):
Thank you!

PR #503, opened 2025-10-30
DownloadContentButton can now handle Engine and Game download

### LINKED ISSUE #350: Map and engine download progress when watching replay

Hectate:
Fixes #350 

## UX changes
``DownloadContentButton`` now has optional props of ``map``, ``engine``, and ``game``. Behavior of the component is as follows:
1. For each of those properties, the component will determine if the content is installed already. If any one (or more) property requires a download, the button will become a "Download" button.
2. Clicking Download will result in an attempt to download only the missing content. Downloads are consecutive, not concurrent, as ``pr-downloader`` always crashed for me if triggered multiple times.
3. The button will become a disabled button with an internal progress bar overlay (as is already), moving from 0-100% for each item requested. Once all requested content is downloaded, the button will revert to the "no downloads required" state as currently in use in the client (_usually_ implemented as "Play" button).
4. Failed downloads will result in errors in console and via NotificationsAPI. The button will revert to the state as above in (1). This will prevent users from gaining access to "Play" if they do not have the required content and permit reattempts to download the content.

## Code changes/fixes
* ``downloadFailed`` was not being sent by ``pr-downloader``, this has been fixed.
* Component ``MapDownloadProgress`` renamed to ``DownloadProgress``
* Component ``DownloadContentButton`` updated per UX changes above
* Download failure logic added to ``map.store`` ``engine.store`` and ``game.store``
* Failed downloads were not removed from ``downloads.store``, this has been fixed.
* Updates to places where ``DownloadContentButton`` was already utilized to utilize the UX changes above.
* ``test2.vue`` now has a testable map/engine/game ``DownloadContentButton``, which incidentally can be used to attempt a forced download of specific map/engine/game content.

### DIFF
--- a/lang/en/lobby.json
+++ b/lang/en/lobby.json
@@ -131,6 +131,7 @@
                 "queueError": "Error loading queues",
                 "buttons": {
                     "searchGame": "Search game",
+                    "joinRequested": "Joining queue",
                     "searchingForOpponent": "Searching for opponent",
                     "matchFound": "Match found",
                     "accepted": "Accepted",
--- a/package-lock.json
+++ b/package-lock.json
@@ -8659,9 +8659,9 @@
             "license": "MIT"
         },
         "node_modules/js-yaml": {
-            "version": "4.1.0",
-            "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.1.0.tgz",
-            "integrity": "sha512-wpxZs9NoxZaJESJGIZTyDEaYpl0FKSA+FB9aJiyemKhMwkxQg63h4T1KJgUGHpTqPDNRcmmYLugrRjJlBtWvRA==",
+            "version": "4.1.1",
+            "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-4.1.1.tgz",
+            "integrity": "sha512-qQKT4zQxXl8lLwBtHMWwaTcGfFOZviOJet3Oy/xmGk2gZH677CJM9EvtfdSkgWcATZhj/55JZ0rmy3myCT5lsA==",
             "dev": true,
             "license": "MIT",
             "dependencies": {
--- a/src/renderer/assets/languages/cs.json
+++ b/src/renderer/assets/languages/cs.json
@@ -2112,6 +2112,7 @@
                 "queueError": null,
                 "buttons": {
                     "searchGame": null,
+                    "joinRequested": null,
                     "searchingForOpponent": null,
                     "matchFound": null,
                     "accepted": null,
--- a/src/renderer/assets/languages/de.json
+++ b/src/renderer/assets/languages/de.json
@@ -1983,6 +1983,7 @@
                 "queueError": null,
                 "buttons": {
                     "searchGame": null,
+                    "joinRequested": null,
                     "searchingForOpponent": null,
                     "matchFound": null,
                     "accepted": null,
--- a/src/renderer/assets/languages/en.json
+++ b/src/renderer/assets/languages/en.json
@@ -1962,6 +1962,7 @@
                 "queueError": "Error loading queues",
                 "buttons": {
                     "searchGame": "Search game",
+                    "joinRequested": "Joining queue",
                     "searchingForOpponent": "Searching for opponent",
                     "matchFound": "Match found",
                     "accepted": "Accepted",
--- a/src/renderer/assets/languages/fr.json
+++ b/src/renderer/assets/languages/fr.json
@@ -3965,6 +3965,7 @@
                 "queueError": null,
                 "buttons": {
                     "searchGame": null,
+                    "joinRequested": null,
                     "searchingForOpponent": null,
                     "matchFound": null,
                     "accepted": null,
--- a/src/renderer/assets/languages/ru.json
+++ b/src/renderer/assets/languages/ru.json
@@ -3934,6 +3934,7 @@
                 "queueError": null,
                 "buttons": {
                     "searchGame": null,
+                    "joinRequested": null,
                     "searchingForOpponent": null,
                     "matchFound": null,
                     "accepted": null,
--- a/src/renderer/assets/languages/zh.json
+++ b/src/renderer/assets/languages/zh.json
@@ -4076,6 +4076,7 @@
                 "queueError": null,
                 "buttons": {
                     "searchGame": null,
+                    "joinRequested": null,
                     "searchingForOpponent": null,
                     "matchFound": null,
                     "accepted": null,
--- a/src/renderer/store/matchmaking.store.ts
+++ b/src/renderer/store/matchmaking.store.ts
@@ -3,17 +3,27 @@
 // SPDX-License-Identifier: MIT
 
 import { reactive } from "vue";
-import { MatchmakingListOkResponseData } from "tachyon-protocol/types";
+import {
+    MatchmakingCancelledEventData,
+    MatchmakingFoundEventData,
+    MatchmakingFoundUpdateEventData,
+    MatchmakingListOkResponseData,
+    MatchmakingQueuesJoinedEventData,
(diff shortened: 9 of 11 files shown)


src/renderer/components/controls/DownloadContentButton.vue
```diff
@@ -7,43 +7,169 @@ SPDX-License-Identifier: MIT
...
 const { t } = useTypedI18n();
 
 export interface Props extends /* @vue-ignore */ ButtonProps {
-    map: MapDownloadData;
+    map?: MapDownloadData;
     disabled?: boolean;
     class?: string;
     onClick?: (event: MouseEvent) => void;
+    engineVersion?: string;
+    gameVersion?: string;
```
p2004a: those probably will need to be all `maps`, `gameVersions` and `engineVersions` to support also the e.g. download for matchmaking.

src/renderer/components/controls/DownloadContentButton.vue
```diff
@@ -7,43 +7,169 @@ SPDX-License-Identifier: MIT
...
+    const downloads = downloadsStore.gameDownloads;
+    for (const download of downloads) {
+        if (download.name === props.gameVersion) {
+            return true;
+        }
+    }
+    return false;
+});
+
+const isDownloading = computed(() => {
```
p2004a: same as in the other file, please remove duplication

src/renderer/components/controls/DownloadContentButton.vue
```diff
@@ -7,43 +7,169 @@ SPDX-License-Identifier: MIT
...
+    if (game && !(await window.game.isVersionInstalled(game))) {
+        try {
+            await downloadGame(game);
+        } catch (error) {
+            console.error(`DownloadGame for ${game} failed:`, error);
+            notificationsApi.alert({ text: "Game download failed.", severity: "error" });
+            refresh.value += 1;
+        }
+    }
+}
```
p2004a: Looks like something that should be handled in some store.

p2004a (changes requested):

src/renderer/components/common/DownloadProgress.vue
```diff
@@ -0,0 +1,81 @@
...
+    const downloads = downloadsStore.gameDownloads;
+    for (const download of downloads) {
+        if (download.name === props.game) {
+            return true;
+        }
+    }
+    return false;
+});
+
+const isDownloading = computed(() => {
```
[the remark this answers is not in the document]
Hectate: Ah yes, these were all ``computedAsync`` initially for various reasons, and I never brought them back together afterward. Will fix.

src/renderer/components/controls/DownloadContentButton.vue
Hectate: Multiple maps I could see, but will we really need to support multiple engine and game version downloads? Obviously, doing one means we just replicate it for the others, but I can't think of a scenario where it'd be wanted.

p2004a: Multiple game and engine versions allow for staged progressive rollouts and A/B testing.

p2004a: And by the way that's already in the protocol https://github.com/beyond-all-reason/tachyon/blob/bbc1aa4817a72f2fcd0d471fe1ca34148e9fa632/schema/matchmaking/list/response.json#L33-L54

src/renderer/components/controls/DownloadContentButton.vue
```diff
@@ -7,43 +7,169 @@ SPDX-License-Identifier: MIT
...
+        }
+    }
+    return false;
+});
+
+const isDownloading = computed(() => {
+    return isMapDownloading.value || isEngineDownloading.value || isGameDownloading.value;
+});
+
+// Note; we have to await each download because pr-downloader will crash if we do two/three at once.
```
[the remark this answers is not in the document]
Hectate: So looking at ``pr-downloader.ts`` and reading the [source about this](https://github.com/beyond-all-reason/pr-downloader/blob/8263d019b8b048f3939ddfbc2595d7f10ef45bf4/src/main.cpp#L50) I see what you mean. We would have to switch the ``downloadContent()`` call to use an array of objects and then handle the download progress appropriately for each download and the stores involved.
Weren't we needing to update ``pr-downloader`` anyway since it's vendored?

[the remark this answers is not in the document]
Hectate: Fair enough. I _did_ already update the code of the component to accept an array of maps, engines, and games, so at least that half will be ready to go.

src/renderer/store/game.store.ts
```diff
@@ -20,15 +21,18 @@ export const gameStore: {
...
     status: GameStatus.CLOSED,
+    availableGameVersions: [],
 });
 
 async function refreshStore() {
     await db.gameVersions.clear();
     const installedVersions = await window.game.getInstalledVersions();
     await db.gameVersions.bulkPut(installedVersions);
+    gameStore.availableGameVersions = Array.from(installedVersions);
     const latestGameVersion = await db.gameVersions.orderBy("gameVersion").last();
```
p2004a: If we do this, we are actually using the `db.gameVersions` for anything useful? Can't we replace remaining usage in `DebugSidebar.vue` and `OfflineBattleComponent.vue` with this property and drop that db table entirely?

src/renderer/components/controls/DownloadContentButton.vue
```diff
@@ -7,43 +7,91 @@ SPDX-License-Identifier: MIT
...
+
+function updateDownloadStatus(value: boolean) {
+    isDownloading.value = value;
 }
 
-const props = defineProps<Props>();
+// Note; we have to await each download because we need to update pr-downloader to accept concurrent downloads
+async function beginDownload(maps?: string[], engines?: string[], games?: string[]) {
+    if (maps) {
+        for (const map of maps) {
```
p2004a: nit: or just `for (const map of maps ?? []) {}`

src/renderer/store/game.store.ts
Hectate: Looking at how it's used, I definitely think it's not required to be in the DB. The one downside that comes to mind by dropping the DB is we lose the "key" of getting data out of the GameVersion type by finding it with a string. (e.g. ``db.gameVersions.get(version:string)``). We would have to search the array for the ``GameVersion.gameVersion`` value that matches the desired string.

As an alternative to the straight ``GameVersion[]`` in this PR, we could instead create a ``Map<string, GameVersion>`` to hold it keyed by version string. How do you feel about that?

p2004a: Sounds good to me, and `Map` has `.keys()` and `.values()` so it's easy to get what you need.

src/renderer/components/battle/OfflineBattleComponent.vue
```diff
@@ -127,7 +132,7 @@ import gridIcon from "@iconify-icons/mdi/grid";
 const mapListOpen = ref(false);
 const mapOptionsOpen = ref(false);
 const mapListOptions = useDexieLiveQuery(() => db.maps.toArray());
-const gameListOptions = useDexieLiveQuery(() => db.gameVersions.toArray());
+const gameListOptions = Array.from(gameStore.availableGameVersions.values());
```
p2004a: This should be computed property, afaiu this way it's just static value. Same issue in `DebugSidebar.vue`.

p2004a:
Ahh, sorry, you are right. I had selected wrong merge type in PR UI and it was not permitting that particular one.

PR #505, opened 2025-11-01
Login Confirmation modal and Subnavigation

### LINKED ISSUE #497: PLAY - Subnavigation
The sub navigation on the PLAY tab should read:
Home
Skirmish vs AI
Campaign
Matchmaking
Custom Lobbies
Scenarios
Tournaments

When players open the app for the first time, they should be on the PLATY - Home tab. 

Players can select either a gamemode from the primary buttons or from the tabs.

### LINKED ISSUE #498: Confirmation Modal for "You need to log in"
Per UI Team direction, users should be able to see all options whether logged in or not. In order to accommodate this, a new Modal is required. 

Requirements:
* Button to "Sign In" (use Green from current Button component)
  * ``@click`` will route the user to the Login screen
  * If possible, after logging in, it would be ideal to route the user onward to the originally intended online-only view
* Button to "Cancel" (use Red from current Button component)
  * ``@click`` will dismiss the Modal, no additional navigation.
* Text (i18n-compliant) stating "You need to be signed in."
* Modal should appear if the user is not logged in and they attempt to access any of the "online-only" features.
  * Matchmaking
  * Custom Lobbies
  * Tournaments
  * Friends
  * Any future enhancements that are online (e.g. Parties) should implement this as well.

robclarked:
Fixes #497 and Fixes #498 

Created the Login confirmation modal, I used the button styes from elsewhere in the app that matched he login button so it looked nice.
Used a composable so the state of the login modal could be accessed easily from other components
I decided to use "Login" and "Play Offline" to match what the login page says and reuse it's translation text. 
I made the onlineOnly UI items visible at all times, and then have them open the modal when the user is not authenticated.
Some of the subnavigation are not correct as they are already fixed in Pull Request #502 
I added guards to the routing to open the modal when unauthenticated users try to navigate to those views.

I have not been able to login to the dev server to fully test the functionality when logged in. Would appreciate if someone who can login is able to test.

Full list of changes:
- Created LogInConfirmationModal based on other modal templates
- Used Button styling from DownloadedContentButton.vue so it actually looks good. (It should be added to the default button template) 
- LogInConfirmationModal redirecs to login page with 'intendedRoute' to redirect back to previous page
- Added translation strings for confirmation modal. Decided to use "Login" to be consistent with the login page.
- Added LogInConfirmationModal to App.vue
- Added Guard to App router that prevents simpleRouterMemory trapping users in prompt hell

- Removed onlineOnly and isAuthenticated checks from the submenu to not hide onlieonly views
- Updated Direct messages in the navar to only show when authentiated isntead of never.
- Updated Friends icon in the navar to always show, but open the LogInConfirmationModal when not authenticated
- Added useLogInConfirmation to Navbar to toggle login modal

- Updated server status in the navar to always show 
- Refactored server status component to properly display the status, including offline.

- Create useLogInConfirmation composable to effectively conrtol the state of the login modal and intendedRoute
- Added useLogInConfirmation to App.vue to toggle login modal

- Updated router to guard the onlineOnly routes and trigger the modal through the composable

- Updated view index login function to redirect to intendedRoute if it is set
- Removed devOnly from the submenu views

### DIFF
--- a/src/renderer/views/play/matchmaking.vue
+++ b/src/renderer/views/play/matchmaking.vue
@@ -5,7 +5,7 @@ SPDX-License-Identifier: MIT
 -->
 
 <route lang="json5">
-{ meta: { title: "Matchmaking", order: 3, onlineOnly: true, transition: { name: "slide-left" } } }
+{ meta: { title: "Matchmaking", order: 3, devOnly: true, onlineOnly: true, transition: { name: "slide-left" } } }
 </route>
 
 <template>


src/renderer/components/misc/LogInConfirmationModal.vue
```diff
@@ -0,0 +1,121 @@
...
+}
+</script>
+
+<style lang="scss" scoped>
+.container {
+    width: 352px;
+    gap: 20px;
+}
+
+.quick-download-button {
```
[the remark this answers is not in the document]
robclarked: Yes. as I mentioned I copied it directly from another component.

I left the classname so it would be obvious what was copied. As I mentioned in the pull request, these classes probably should be moved into the the button, as I believe this is the Green and Red buttons that where being referred to. 

The issue asks for us to use the Green and Red from the current Button component. However the button components defaults where not being used else where. Each time a button component was used that was green or red, these custom styles where used.

robclarked:
Here is a screenshot of the login modal:
Screenshot 2025-11-02 101659

Here is a screenshot with the server status in the offline state, and the friends tab showing. clicking on these while not logged in will open the login popup.
Screenshot 2025-11-02 101614

src/renderer/components/misc/LogInConfirmationModal.vue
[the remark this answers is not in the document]
Hectate: I suspect there's a bit of reorganization/centralization needed across all component styles. Probably worth a separate issue to review and address, similar to the i18n approach we have.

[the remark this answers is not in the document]
p2004a: Ack, thanks, that makes sense. I've opened https://github.com/beyond-all-reason/bar-lobby/issues/506 to track this.

src/renderer/components/navbar/NavBar.vue
```diff
@@ -73,7 +72,7 @@ SPDX-License-Identifier: MIT
                     </Button>
                 </div>
                 <div class="secondary-right flex-row flex-right">
-                    <ServerStatus v-if="me.isAuthenticated" />
+                    <ServerStatus />
```
p2004a: Hide if not dev mode disabled

src/renderer/components/navbar/NavBar.vue
```diff
@@ -31,11 +31,10 @@ SPDX-License-Identifier: MIT
                         <div v-if="messagesUnread" class="unread-dot"></div>
                     </Button>
                     <Button
```
p2004a: Hide if not dev mode disabled

src/renderer/views/index.vue
```diff
@@ -60,7 +60,8 @@ async function login() {
         connecting.value = true;
         await auth.login();
         await tachyon.connect();
-        router.push("/play");
+        const redirect = router.currentRoute.value.query.redirect as string | undefined;
+        redirect ? router.push(redirect) : router.push("/play");
```
p2004a: nit: let's use normal if condition for readability or flip to `router.push(redirect ? redirect : '/play')`

src/renderer/App.vue
```diff
@@ -123,12 +128,29 @@ window.barNavigation.onNavigateTo((target: string) => {
...
+    const isAuthed = Boolean(toValue(me.isAuthenticated));
+    if (!isAuthed) {
+        const resolved = router.resolve(rememberedPath);
+        const requiresAuth = resolved.matched?.some((r) => Boolean(r.meta?.onlineOnly)) ?? Boolean(resolved.meta?.onlineOnly);
+
+        if (requiresAuth) {
+            const defaultPath = typeof defaultRedirect === "string" ? defaultRedirect : router.resolve(defaultRedirect).fullPath;
+            return { path: defaultPath };
+        }
     }
```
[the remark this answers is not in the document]
robclarked: I had to add this auth check to prevent an issue when navigating between top level menu routes. It would prevent a user from navigating without logging in if the last visited route was online only.

To see the bug / test it:
- Remove the !isAuthed guard
- Start the app
- Click Play Offline
- navigate to an onlineOnly submenu item like 'matchmaking'
- Log in? modal appears. dismiss / play offline
- Navigate to top level route line 'NEWS'
- Navigate back to 'PLAY' route
- Log in? modal will appear and prevent navigation

The guard essentially prevents this by redirecting them to the default redirect for the route instead of the remembered submenu route.

I've relooked at it and improved the implementation, it should now be cleaner and clearer.

src/renderer/views/index.vue
robclarked: I simplified it entirely to just: `router.push(redirect || "/play");`

src/renderer/components/navbar/NavBar.vue
robclarked: I put friends, direct messages, and player count behind devMode,

robclarked:
Thankyou! I've put those the online functionality behind dev mode, and improved those other bits.

Issue #424, opened 2025-06-19
Setup I18n automated workflow


webdevsavvy:
Now that basic support for i18n has been added we need a way to add an automated workflow for i18n.

This issue aims to track the setup for this automated workflow and its relevant documentation.

**NOTES**

- This node package [https://github.com/Spittal/vue-i18n-extract](https://github.com/Spittal/vue-i18n-extract) reports unused translation keys in code, and can be helpful for the automated workflow.

webdevsavvy:
@p2004a I've edited the description of this issue to clarify what my intentions where for this issue.

PR #521, opened 2025-12-01
Update lang/README.md to add contribution instructions + fix a few typos

Wabuo:
The title says it all ...

### DIFF
--- a/lang/README.md
+++ b/lang/README.md
@@ -7,3 +7,31 @@ SPDX-License-Identifier: MIT
 # What is this directory?
 
 This directory and its contents are a placeholder for the future git submodule that will manage translation files for both this repository and the main game repository, which will itself be managed through Transifex.
+
+### Do not edit any files in the `lang` folder manually, except for this READEME.md
+
+All other files are managed automatically!
+
+## How to contribute:
+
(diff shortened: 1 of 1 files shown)


lang/README.md
```diff
@@ -6,4 +6,32 @@ SPDX-License-Identifier: MIT
 
 # What is this directory?
 
-This directory and it's contents are a placeholder for the future git submodule that will manage translation files for both this repository and the main game repository, that will it self be managed through transifax.
+This directory and its contents are a placeholder for the future git submodule that will manage translation files for both this repository and the main game repository, which will itself be managed through Transifex.
+
+### Do not edit any files in the `lang` folder manually, except for this READEME.md
+
+All other files are managed automatically!
```
p2004a: No, they are not. They *will* be in the future, as written in the paragraph that exits at the top and tracked in https://github.com/beyond-all-reason/bar-lobby/issues/424

And once this directory will be a submodule, this readme file won't even exist, so this text won't exist.

The current process is described in https://github.com/beyond-all-reason/bar-lobby/wiki/Internationalization and it says "To create a new translation string, you want to first edit the ./lang/en/lobby.json to include the new strings"

PR #522, opened 2025-12-08
Don't update engine version after downloads

### LINKED ISSUE #515: Finishing a map download resets engine selection
https://github.com/user-attachments/assets/75752fd5-ef0d-4db5-8f84-1c728ce77e1a

As you can see in the video, as soon as a map download finishes the selected engine gets reset to default.

JasonReed-2:
Fixes #515. Refreshing the available engine versions should be decoupled from selecting the default engine version.


p2004a (approved):
Thank you!

PR #529, opened 2026-01-02
issue 202: mimic border with dropshadow to avoid rerender loops that cause sound spam

### LINKED ISSUE #202: Hovering on the edge of controls spams their hover event and sound

gdoteof:
this appears to fix #202 .. by working around bug at some level of the stack outside this repo.  

another fix that does a worse job of maintaining current look and feel is to remove the border changes for the hover event.  

another fix that leaves the current css but has potential bad side effects is to delay returning `onMouseEnter` by dozens of ms.  also, somewhat frustratingly, simply adding a console.log command after the  `audioApi.play` *also* fixes this.  basically any delay and we don't get the loop.  i can't say what the true root cause is, but i am confident it has something to do with rerender of the border re-triggering the mouse enter event

video evidence was too large to embed directly

https://youtu.be/GnFkbTo9wl4

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,11 +14,11 @@ If you haven't already, join our Discord at https://discord.gg/beyond-all-reason
 
 Please feel free to ask questions and seek help from the community in the channel. Often, a fellow contributor can provide a quick answer!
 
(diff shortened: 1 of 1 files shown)


bcdrme:
~It does fix the issue. Even though the style is not 1:1 with what we had before, I think it's acceptable.~\

EDIT: Nevermind it doesn't fix the issue. I can easily repro here:
image

PR #531, opened 2026-01-05
Offline Play

SirPranceAlot:
# Offline Mode & Map Scanning Improvements

## Summary

This PR introduces offline mode support for map fetching and adds map scanning functionality to improve the user experience when working without an internet connection. The changes ensure that users can still access and manage their locally installed maps while offline.

### New Features

- __Offline mode for map fetching__: Maps service now checks `isOnline()` to determine whether to fetch maps from online sources or use local data
- __Map scanning functionality__: Added `scanMaps` IPC command and exposed it through the preload API for scanning locally installed maps
- __Play Offline option__: Users can now click "Play Offline" in the login confirmation modal to proceed without authentication

### Technical Details

Why change localhost to the ip? 
Because localhost won't resolve if the user disables their network.

- __Protected Asset Paths__:

  - Defined as `readonly string[]` in `app.ts`
  - Maps: `[WRITE_DATA_PATH/maps, ASSETS_PATH/maps]`
  - Games: `[WRITE_DATA_PATH/games, ASSETS_PATH/games]`

- __Implementation Characteristics__:

  - Immutable path definitions (`readonly`)
  - Separate writeable vs. protected asset paths
  - ASSETS_PATH is treated as read-only content
  - WRITE_DATA_PATH allows user modifications

- __Integration Points__:

  - Map scanning only modifies WRITE_DATA_PATH
  - Game downloads save to WRITE_DATA_PATH/games
  - ASSETS_PATH remains untouched during operations

image

image

image

- __Account Service__ (`account.service.ts`)

  - Provides `isOnline()` for connectivity checks

- __Game Launch__ (`game.ts`)

  - Modifies battle scripts for offline mode
  - Uses `SPRING_DATADIR=ASSETS_PATH` for protected assets

- __Map Service__ (`maps.service.ts`)

  - Returns empty arrays when offline
  - Provides `scanMaps()` IPC for local detection
  - Maintains separate live/non-live map lists

- __Replay Worker__ (`parse-replay-worker.ts`)

  - Special handling for Z_BUF_ERROR
  - Treats as non-fatal error
  - Marks replays as potentially corrupted

Error Handling Flow (Z_BUF_ERROR)

image

__. Offline Mode Impact__

- Battle scripts modified locally
- DNS-based connectivity checks
- Protected assets remain immutable
- Local scanning replaces remote fetching
- Special error handling for offline edge cases

## Testing 

Tested locally with the network off. The debug msgs seen have been removed.

https://github.com/user-attachments/assets/44e1ddc1-cbd6-4d71-88ba-5e6c5ce8adf4

## Related Issues

- Enables offline gameplay functionality
- Improves map management for users with limited connectivity

### DIFF
--- a/vendor/map-parser/ultrasimple-map-parser.ts
+++ b/vendor/map-parser/ultrasimple-map-parser.ts
@@ -60,8 +60,26 @@ export class UltraSimpleMapParser {
 
     protected async parseMapInfo(mapinfo: string): Promise<MapInfo> {
         const parsedMapInfo = parse(mapinfo, { encodingMode: "x-user-defined", comments: false });
-        const rootObj = parsedMapInfo.body[0] as LocalStatement;
-        const rootTable = rootObj.init.find((block) => block.type === "TableConstructorExpression") as TableConstructorExpression;
+
+        if (!parsedMapInfo.body || parsedMapInfo.body.length === 0) {
+            console.warn("Map info is empty or malformed.");
+            return {} as MapInfo;
+        }
+
+        const rootObj = parsedMapInfo.body[0];
+
+        if (rootObj.type !== "LocalStatement" || !rootObj.init || rootObj.init.length === 0) {
+            console.warn("Map info root object is not a LocalStatement or is missing init property.");
+            return {} as MapInfo;
+        }
+
+        const rootTable = rootObj.init.find((block) => block.type === "TableConstructorExpression");
+
+        if (!rootTable || rootTable.type !== "TableConstructorExpression") {
+            console.warn("Map info root table is not a TableConstructorExpression.");
+            return {} as MapInfo;
+        }
+
         const obj = this.parseMapInfoFields(rootTable.fields);
         return obj as MapInfo;
     }
--- a/src/main/config/app.ts
+++ b/src/main/config/app.ts
@@ -91,9 +91,6 @@ export const STATE_PATH = path.resolve(process.env.BAR_STATE_PATH || defaultLoca
 // locations, not only our own code.
 app.setPath("userData", STATE_PATH);
 
-console.log(`ASSETS_PATH: ${ASSETS_PATH}`);
-console.log(`STATE_PATH: ${STATE_PATH}`);
-
 export const CONFIG_PATH = path.join(STATE_PATH, "config");
 export const LOGS_PATH = path.join(STATE_PATH, "logs");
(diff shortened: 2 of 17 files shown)


vendor/map-parser/ultrasimple-map-parser.ts
```diff
@@ -60,8 +60,26 @@ export class UltraSimpleMapParser {
 
     protected async parseMapInfo(mapinfo: string): Promise<MapInfo> {
         const parsedMapInfo = parse(mapinfo, { encodingMode: "x-user-defined", comments: false });
-        const rootObj = parsedMapInfo.body[0] as LocalStatement;
-        const rootTable = rootObj.init.find((block) => block.type === "TableConstructorExpression") as TableConstructorExpression;
+
```
p2004a: This looks like it should be it's own PR with it's own description. All the `as MapInfo` type casts are wrong.

p2004a (commented):
For now, only very surface level review.

This PR has a significant design flaw: it makes incorrect assumption about network connectivity:
- it is binary, either or on off
- the single detection mechanism based on DNS is sufficient
- that it can be just automatically determined at all
- that it's constant over time of the update
- that we don't have outages in our infra we want to protect against with this mode

Those are not correct in _majority_ of cases. There is a Disord thread with a more extensive discussion https://discord.com/channels/549281623154229250/1401696418909126666/1401696422650314862 I've linked it now also from issue, sorry it was missed :(.

Lobby needs to be resilient against every single full and partial outage of any of the services that are being contacted during update. Either:
- A way for user to skip certain steps allowing to depend on "last already fetched" when e.g. one endpoint is problematic. That's the approach taken by our launcher: some steps are optional other are skippable, it works well in 99% of cases but there are some unfortunate edge cases sometimes when state of assets is not in sync (e.g. user can't see games, because engine download failed, and the skipped it and they are confused about it)
- A way to trigger *by user* to switch to full offline mode at any point during the update progress, persistent button always present, and also popping up when a particular step fails or times out (offline being binary on/off is a bit weird in case only e.g. a single endpoint during update fails, it blocks fully progress to online mode, and that particular update endpoint isn't actually *critical* (e.g. official map list))

PS. I get an unhealthy AI smell from this PR, especially from description.

Hectate:
Closing this as #572 supersedes this work and has been merged.

PR #560, opened 2026-03-14
Clarify readme for development

geekingfrog:
By far the number one problem of devs trying to run the dev version of the lobby is about the correct node.js version. This is an (optimistic) attempt to fix that.


p2004a:
https://docs.npmjs.com/cli/v8/using-npm/config#engine-strict

PR #561, opened 2026-03-14
Update profile

Rimek86:
grafik

I've updated the profile page a bit to include the information already available in the Tachyon Protocol, and at the same time, I've improved the design a little. (As a first small step here.)

**Updates:**
- If the own profile is displayed a additional label "[My profile]" is visible.
- User ID is now displayed.
- Status is now translated with en.json and has an status icon (Green circle: Menu / Green dot: Lobby / Yellow dot: Playing / Grey dot: Offline).
- If user is no clan member a "Find clan" button is visible. If the user is a clan member the tag and name is displayed.
- If the user has special roles assigned they are now displayed.
- The rating is displayed.
- Layout updated that labels and values are now in rows and cols.
- **BUGFIX:** Now the data of a friend are visible and not "Unknown user" is displayed.


src/renderer/views/profile/[userId].vue
```diff
@@ -11,14 +11,48 @@ SPDX-License-Identifier: MIT
     <div class="view">
         <Panel class="profile-container" v-if="user">
             <div class="profile-header">
-                <img ref="logo" class="avatar" src="/src/renderer/assets/images/BARLogoFull.png" />
+                <!--<img ref="logo" class="avatar" src="/src/renderer/assets/images/BARLogoFull.png" />-->
```
Comment (geekingfrog) [inline src/renderer/views/profile/[userId].vue]: just remove the line instead of commenting.

src/renderer/views/profile/[userId].vue
```diff
@@ -11,14 +11,48 @@ SPDX-License-Identifier: MIT
...
                 <div class="profile-user-info">
                     <h2 class="flex-row gap-lg">
                         <Flag :countryCode="user.countryCode" style="width: 50px" />
                         {{ user.displayName }}
+                        <span v-if="'isMe' in user && user.isMe === 1" class="my-profile-note"
```
Comment (geekingfrog) [inline src/renderer/views/profile/[userId].vue]: I don't like having `isMe in user` everywhere. You should be able to enforce this property is always there. But also, `{}.foo === undefined`, so you don't even need to check for the property to exist if you're going to check it's equal to 1 right after.

src/renderer/views/profile/[userId].vue
```diff
@@ -34,14 +68,44 @@ import Panel from "@renderer/components/common/Panel.vue";
...
+        .join(", ");
+}
+
 const props = defineProps<{
     userId: string;
 }>();
 
-const user = useDexieLiveQueryWithDeps([() => props.userId], () => {
-    return db.users.get(props.userId);
+const user = useDexieLiveQueryWithDeps([() => props.userId], async () => {
```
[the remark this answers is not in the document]
Comment (Rimek86) [reply src/renderer/views/profile/[userId].vue]: Retval is just an auxiliary variable that’s supposed to always contain the user. It’s retrieved only once from the dexie-db and, for a different profile (a friend’s, for example), via a single request. The linter said I also need to account for null... But maybe I did something else wrong?

src/renderer/views/profile/[userId].vue
```diff
@@ -34,14 +68,44 @@ import Panel from "@renderer/components/common/Panel.vue";
 import { useDexieLiveQueryWithDeps } from "@renderer/composables/useDexieLiveQuery";
 import { db } from "@renderer/store/db";
 import { useTypedI18n } from "@renderer/i18n";
+import { fetchUserInfo } from "@renderer/store/users.store";
 const { t } = useTypedI18n();
 
+type UserRole = "contributor" | "admin" | "moderator" | "tournament_winner" | "tournament_caster";
```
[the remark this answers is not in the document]
Comment (Rimek86) [reply src/renderer/views/profile/[userId].vue]: You are right. I'll fix it.

Rimek86:
@geekingfrog I think i have solved everything.

src/renderer/store/users.store.ts
```diff
@@ -23,24 +23,38 @@ export function initUsersStore() {
...
+                    username: user.username ?? "Unknown User",
+                    displayName: user.displayName ?? "Unknown User",
+                    clanBaseData: user.clanBaseData ?? null,
                     partyId: null,
-                    countryCode: "??",
-                    status: "offline",
+                    countryCode: user.countryCode ?? "??",
+                    status: user.status ?? "offline",
+                    roles: user.roles ?? [],
+                    rating: user.rating ?? undefined,
```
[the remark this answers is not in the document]
Rimek86: Now i know what i have to do...

src/renderer/store/db.ts
```diff
@@ -72,9 +72,20 @@ db.version(1).stores({
     users: "userId, username, countryCode, status, displayName, clanId, partyId, scopes, isMe",
 });
 
-db.version(2).stores({
-    users: "userId, username, countryCode, status, displayName, partyId, scopes, isMe",
-});
+db.version(2)
```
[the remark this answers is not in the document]
Rimek86: Was also in Master. Have no idea why it is implemented...

src/main/model/user.ts
```diff
@@ -17,11 +17,11 @@ export type User = {
...
     countryCode: string;
     status: "offline" | "menu" | "playing" | "lobby";
-    rating?: { value: number };
-    roles?: ("contributor" | "admin" | "moderator" | "tournament_winner" | "tournament_caster")[];
+    rating?: { value: number } | null;
+    roles?: ReadonlyArray<"contributor" | "admin" | "moderator" | "tournament_winner" | "tournament_caster">;
 
     // Is the user me?
-    isMe?: 0 | 1;
+    isMe: 0 | 1;
```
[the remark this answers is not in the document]
geekingfrog: Agree, a boolean would be better, but also, this can be done in a separate PR

Rimek86 (commented):
> It is to ensure data consistency when the schema evolve. When we update the code/client and the local db is using a previous schema. I think we should keep that even if we can be a bit loose with it for now, and consider only one previous version. 

In that case, I'll switch to version 2, which is already on the master branch.

Rimek86 (commented):
> Seems there's still a typechecker problem 

I'm not quite sure why yet. I ran “npm run typecheck:node” and “npm run typecheck:web” locally without any errors. What did I do wrong? Do I need to run a different command locally to test it first?

I'll fix it, of course. :-)

Rimek86:
> Clicking "approve launching workflow" only to see red again starts getting tired 😛

Also that i think i have everything done... Don't understand why the local check on my client passes... (I'd appreciate any tips on how to prevent this. )

Rimek86:
Don't understand why the results differ

Rimek86:
@p2004a @geekingfrog I merge the upstream master with the current taychon version. Now everything works fine. :-)

src/renderer/store/db.ts
```diff
@@ -72,6 +72,25 @@ db.version(1).stores({
...
+    .upgrade(async (tx) => {
+        await tx
+            .table("users")
+            .toCollection()
+            .modify((user) => {
+                if (typeof user.isMe === "undefined") {
+                    user.isMe = 0;
+                }
+            });
+    });
```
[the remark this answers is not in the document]
p2004a: not fixed

src/renderer/store/users.store.ts
```diff
@@ -23,24 +23,51 @@ export function initUsersStore() {
                 console.warn("Received user/updated event with no userId, skipping update.");
                 return;
             }
-            const updated = await db.users.update(user.userId, { ...user });
+
```
p2004a, line 28: @geekingfrog 

With the changes in protocol, this doesn't look correct to me and I believe with `clanBaseData` added to protocol, the semantic in protocol is no longer clear.

Initially the user object was quite flat, and the protocol says "User objects should be full when first sent, then only updates gets sent." we didn't upgrade the text with the method we use for lobby (that was designed later): json merge patch. So right now protocol allow only setting values, they can never be deleted. I don't think it matches the user object type in protocol anymore.

As for the implementation here, if the update message doesn't have `user.clanBaseData` set at all, because it didn't change on server, this code will drop this field from the user incorrectly.

src/renderer/store/db.ts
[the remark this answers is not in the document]
Rimek86: Please see last commit. It's already fixed...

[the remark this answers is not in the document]
p2004a: >  Also, you can check against undefined without going through the type x === undefined works just as well.

src/renderer/store/users.store.ts
Rimek86: I don't quite understand the problem yet: If a user doesn't have any clan data because they aren't in a clan, then it makes sense that the data would be deleted from the local database...

p2004a: When only `status` field changed for user, server sends update with only status set, and current code won't just update `status` in the database, but also purge `clanBaseData`.

geekingfrog: > So right now protocol allow only setting values, they can never be deleted. I don't think it matches the user object type in protocol anymore.

Looking at the [user object](https://github.com/beyond-all-reason/tachyon/blob/master/src/schema/definitions/user.ts), I think this is still alright as the only bit of data that can really be removed is the clan, which is marked as nullable.
Although the integration of `clanUpdateableBaseData` as is means that indeed, we can't send partial update for the clan, which isn't ideal and should be fixed.

The implementation in the user store is indeed buggy then, that's a good catch.
Because the key being undefined/not present in the object is not the same as being null, but `x ? ...` doesn't make the difference.

So yeah, a player going into a lobby would then wipe its clan data from the DB, oops.

src/renderer/store/users.store.ts
```diff
@@ -23,24 +23,51 @@ export function initUsersStore() {
...
 
             if (updated === 0) {
                 // No records updated, so user doesn't exist - create new user
                 db.users.add({
                     userId: user.userId,
-                    username: "Unknown User",
-                    displayName: "Unknown User",
-                    clanId: null,
+                    username: user.username ?? "Unknown User",
+                    displayName: user.displayName ?? "Unknown User",
```
geekingfrog, line 47: I think there's a problem there. If you don't have anything in the DB, but you are getting a partial update, then you can't really create the correct record. So either you need to check that the event you got has all the data, that is, it's the initial event and use that. Or, if it's a partial update, you can't use the event and would need to `fetchUserInfo` to then create the data.

I am not sure how that could happen, but creating a user with these default seems just wrong.

src/renderer/store/users.store.ts
```diff
@@ -23,24 +23,51 @@ export function initUsersStore() {
...
+                    ...user,
                 });
             }
         });
     });
 
     usersStore.isInitialized = true;
 }
+
+export async function fetchUserInfo(userId: string) {
```
geekingfrog: not sure if there are convention in ts/this repo, but shouldn't this be annotated with the return type?

src/renderer/store/users.store.ts
p2004a: > If you don't have anything in the DB, (...)

The original code just assumes that it can happen and fills in some blanks, no harm done if server functions correctly, but we can make it cleaner.

> Or, if it's a partial update, you can't use the event and would need to fetchUserInfo to then create the data.

Not really. Recall the issues with ordering of events/reponses. This would happen here too.

src/renderer/store/users.store.ts
p2004a: I think it would be clean if we made it json merge path, it makes semantic very, very clear.

src/renderer/store/db.ts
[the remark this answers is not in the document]
Rimek86: Fixed in next commit...

src/renderer/store/users.store.ts
Rimek86: I think the original author programmed defensively here. I don't think it's wrong to do it that way. It only covers the case where there is inconsistent or incomplete data. It should never happen...

@p2004a How we can make it cleaner? (My goal wasn't to imrpove the users.store.ts My goal was to improve the profiles page here. :-) )

You once asked me not to change too much at once. That's why I wanted to stick to it and not touch everything directly with a PR. Can we possibly make this an issue that will then be changed later?

src/renderer/store/users.store.ts
Rimek86: I didn't receive any warning. But I still changed it in the next commit.

src/renderer/store/users.store.ts
Rimek86: Will change it in next commit...

src/renderer/store/users.store.ts
Rimek86: @p2004a Hi, i was absent in the last weeks. But now i thin i find time again to continue. What's about my question so that i can continoue here...

p2004a: It needs to be changed in protocol to be merge patch too, including all the implications of that: drop all the allowed nulls, it's reserved to patch semantic. (overall, having a optional field that can also take null is weird)

> My goal wasn't to imrpove the users.store.ts My goal was to improve the profiles page here. :-)

I mean, piling up more mess on the top is not correct strategy, this is not deep refactoring, you are changing this code.

All this code can be simplified to

```ts
const toUpdate = await db.users.bulkGet(event.users.map((u) => u.userId));
const updatedUsers = event.users.map((u, i) => applyPatch(toUpdate[i] || {}, u));
await db.users.bulkPut(updatedUsers);
```

Hectate:
Closing as stale. There is likely to be changes to the UI regardless, so the result would not survive until 1.0 anyway.

PR #567, opened 2026-03-27
Initial Tachyon-based chat implementation

Hectate:
Chat/Messaging components in the client were non-functional placeholders or contained outdated assumptions about the API that would be used for this feature. This PR removes all placeholders, enables a working direct message feature, and prepares the ``chat.store`` for future development (parties, lobbies).

## Changes:
* Removed several files not required for Tachyon-based chat
* Removed the placeholder "bottom chat" component because it is already known that it will not be present in future UI designs
* Full re-write of ``chat.store.ts`` to match the conventions of other Tachyon-based stores
* Updated Messaging and other components to utilize the new ``chat.store``

Note, these direct messages are by design not expected to stay around between client sessions, nor after closing a direct message "channel". Should some form of persistence to the data be desired, it can be added in the future. Doing so would require changes to how the Messages component functions (to allow searching/showing/hiding chat histories) that are not needed for this initial PR.
Additionally, this PR does not contemplate the concept of "chat rooms" outside of Parties, Lobbies, and Direct Messages. Should those be requested in the future it could be extended to such channels, but again are not needed at this time for this initial PR.

### AI / LLM usage statement:
This PR was created exclusively by a human.


src/renderer/store/chat.store.ts
```diff
@@ -3,131 +3,116 @@
...
+        chatStore.partyChat.push({ ...data, seen: false, isMe: false });
+    }
+}
+
+export function insertSelfMessage(targetUserId: UserId, message: string) {
+    subsManager.attach(targetUserId, chatSymbol);
+    const data: MessagingReceivedEventData = {
+        message: message,
+        source: { type: "player", userId: me.userId },
+        timestamp: Date.now(),
```
Hectate: This will give us milliseconds. I do not know what the timestamp provided from the server will be, so this may need to be modified.

NortySpock: Unix timestamp in microseconds, counting from 1970-01-01 


 https://github.com/beyond-all-reason/tachyon/blob/19beb9e08041b1d03db6786d667f1a00f2c3e051/schema/messaging/received/event.json#L61  says   "$ref": "../../definitions/unixTime.json", which reads 


```
{
    "$id": "https://schema.beyondallreason.dev/tachyon/definitions/unixTime.json",
    "title": "UnixTime",
    "description": "Unix timestamp in microseconds",
    "type": "integer",
    "examples": [1705432698000000]
}
```

https://github.com/beyond-all-reason/tachyon/blob/master/schema/definitions/unixTime.json

Hectate: Ah thanks, yes I couldn’t remember where I saw that.

src/renderer/store/chat.store.ts
```diff
@@ -3,131 +3,116 @@
...
-export interface ChatMessage {
-    userId: string;
-    userName: string;
-    text: string;
-    timestamp: number;
+function requestSend(data: MessagingSendRequestData) {
+    try {
+        const response = window.tachyon.request("messaging/send", data);
+        console.log("Tachyon messaging/send:", response);
+        // Upon success, we add our own message to the body if it's a "player" type
```
geekingfrog: this doesn't explain *why* this is done.

src/renderer/model/message.ts
```diff
@@ -0,0 +1,10 @@
+// SPDX-FileCopyrightText: 2025 The BAR Lobby Authors
+//
+// SPDX-License-Identifier: MIT
+
+import { MessagingReceivedEventData } from "tachyon-protocol/types";
+
+export type Message = MessagingReceivedEventData & {
+    seen: boolean;
+    isMe: boolean; // Required so we can distinguish messages we add from ourselves to the array.
```
geekingfrog: I am not convinced by the usefulness of this flag since you can recover the same data by comparing the `userId` of the message with your own.

src/renderer/components/navbar/Messages.vue
```diff
@@ -91,10 +93,22 @@ const emits = defineEmits<{
     (event: "update:modelValue", open: boolean): void;
 }>();
 
+const displayNames = computedAsync(async () => {
+    const users: Map<UserId, string> = new Map();
+    for (const id of chatStore.userChats.keys()) {
+        const cached = await getUserByID(id);
```
geekingfrog: looking at that function, it queries the local db. Is there any guarantee that the user is in there?
Also, this is doing a bunch of async call without any dependencies between them, `Promise.all` would be better suited there.

Maybe that's out of scope for this PR though. I think the user store should be able to be smarter about managing user data, with subscription and db abstracted away.

src/renderer/components/navbar/Messages.vue
```diff
@@ -21,10 +21,10 @@ SPDX-License-Identifier: MIT
                         <div
                             v-for="(message, i) in messages"
                             :key="i"
-                            v-in-view.once="() => (message.read = true)"
-                            :class="['message', { fromMe: message.senderUserId === me.userId }]"
+                            v-in-view.once="() => (message.seen = true)"
+                            :class="['message', { fromMe: message.isMe }]"
```
geekingfrog: there is a `message.userId` present so I think that can still be used instead of this `isMe`

src/renderer/store/chat.store.ts
```diff
@@ -3,131 +3,109 @@
...
-        const selectedChatRoom = chatStore.chatRooms.find((room) => room.id === id);
+function insertSelfMessage(source: MessagingReceivedEventData["source"], message: string) {
+    const data: MessagingReceivedEventData = {
+        message: message,
+        source: source,
+        timestamp: Date.now() * 1000,
+        marker: "",
+    };
+    onMessagingReceivedEvent(data, true);
+}
```
geekingfrog, line 128: I find the logic between `onMessagingReceivedEvent` and `insertSelfMessage` backward. Modifying the internal store should not manually call an event handler supposed to be invoked when receiving a message from the server.

I think a more natural way would be to have something along the lines of:

```typescript
function onMessagingReceivedEvent(data) {
  console.log("...")
  insertMessage(data)
}

function insertSelfMessage(source, message) {
  const data = ...
  insertMessage(data)
}

// don't export this function, it's for internal use.
function insertMessage(data) {
  // do the thing that are currently in onMessagingReceivedEvent
}

```

src/renderer/components/navbar/Messages.vue
```diff
@@ -118,41 +132,24 @@ function focusTextbox(el: HTMLElement) {
...
 }
 
-function close(userId: number) {
-    console.log("close", userId);
-    // api.session.directMessages.delete(userId);
+function close(userId: string) {
+    // If we want user chat history to persist, then the interface will need to be changed.
+    // Until that time, we simply purge the chat entirely. It would not persists on restart
+    // of the client anyway, so we do not need to extend that at this time.
+    chatStore.userChats.delete(userId);
```
geekingfrog: there should be an exported function to do that, similar to `clearPartyChat` and `clearLobbyChat`

src/renderer/model/message.ts
Hectate: ~~You could, yes, but the flag makes the UI-side much easier to work with since the information is right there.~~

Edit: Will remove this flag as per other review comments; it's not so challenging to use the user ID to check so the boolean is of minimal benefit.

src/renderer/components/navbar/Messages.vue
Hectate: If the user is not present, we will get ``undefined`` and the name will become a generic "User 123" string (subject to the translation string just below this code).

I see what you mean about ``Promise.all``. I'll take a look at what can be done with this function.

src/renderer/components/navbar/Messages.vue
Hectate: Yeah, as I noted earlier I felt it was cleaner to have a simple boolean instead, but it's entirely possible to go the former route instead and remove the flag.

src/renderer/store/chat.store.ts
Hectate: Reasonable. I had a somewhat different thing at first and then I realized that the self-messages could be treated functionally as coming from the server as well with only minimal tweaks. That said, I do like the idea you present, and it's easier to follow in the code, so I'll explore that direction.

src/renderer/components/navbar/Messages.vue
Hectate: Makes sense. Will fix.

Hectate:
Testing note: Clicking the "message" icon from the Friend list does not open the Messages TabView to that friend (if there are multiple tabs). That tab is present and clickable, but I would have preferred that this button also open the Panel to that particular tab. I tried quite a few different things and I'm convinced that the issue is a PrimeVue issue (perhaps one resolved already since we are on an outdated version) or related to how we set up the component initially. Since I was unable to make it work the way I desired, I've submitted the PR with this minor inconvenience.

I would welcome attempts on how to fix this.

src/renderer/store/users.store.ts
```diff
@@ -12,6 +12,14 @@ export const usersStore: {
     isInitialized: false,
 });
 
+export async function getUserById(id: string) {
+    return await db.users.get(id);
+}
+
+export async function getUsersByIds(users: string[]) {
```
geekingfrog: nitpick: that should be `userIds: string[]`

src/renderer/store/tachyon.store.ts
```diff
@@ -82,6 +83,11 @@ export async function initTachyonStore() {
         if (matchmakingStore.isInitialized) {
             matchmaking.sendListRequest();
         }
+
+        // Subscribe to direct messages when connected
```
geekingfrog: small correction, this subscribe to *all* chat messages. Without it, party and lobbies are also disabled iirc

src/renderer/store/chat.store.ts
```diff
@@ -3,131 +3,121 @@
...
 
-        if (selectedChatRoom) {
-            chatStore.selectedChatRoom = selectedChatRoom;
-            chatStore.selectedChatRoom.unreadMessages = 0;
+function insertMessage(data: MessagingReceivedEventData, self?: boolean) {
+    const targetUser: UserId = data.source.userId;
+    if (self) {
+        data.source.userId = me.userId;
+    }
```
geekingfrog: I am not a big fan of the `self` boolean, especially there. I think this bit should be inside `insertSelfMessage`, so that this `insertMessage` function doesn't have to deal with any of these shenanigans.

src/renderer/store/chat.store.ts
```diff
@@ -3,131 +3,121 @@
...
+        if (data.target.type === "player" && data.target.userId !== me.userId) {
+            // We attach for DMs only, because lobby/party will handle their own.
+            subsManager.attach(data.target.userId, chatSymbol);
+            insertSelfMessage({ type: "player", userId: data.target.userId }, data.message);
+        }
+        if (data.target.type === "party") {
+            insertSelfMessage({ type: "party", partyId: "", userId: me.userId }, data.message);
+        }
+        if (data.target.type === "lobby") {
+            insertSelfMessage({ type: "lobby", lobbyId: "", userId: me.userId }, data.message);
```
geekingfrog: empty `lobbyId` and `partyId` seems a bit sus. Annoyingly, looking at the `SendRequestData` it doesn't hold any actual ID. So I think this function should populate these ID based on what's in the lobby and party stores.
Although it doesn't do anything right now, I feel it's going to bite us in the future to have these invalid ids.

geekingfrog (commented):
some small nitpick but otherwise should be pretty close. Just the question about bogus ids

src/renderer/store/tachyon.store.ts
Hectate: Indeed that is correct, messages currently are all or none. I'll modify the comment. Which reminds me, we'll eventually be asked to do things like block users (could be client-side), set status visible/invisible (needs server requests), etc.

src/renderer/store/chat.store.ts
Hectate: I had considered this oddity as well, and came to the conclusion that the IDs didn't matter because we only allow users to be in a single party and/or lobby at a time.

src/renderer/store/users.store.ts
Hectate: Agreed.

src/renderer/store/chat.store.ts
Hectate: The issue revolves around the fact that a message "from ourselves" would automatically get inserted into a chat history keyed by our own ID (e.g. a chat with ourselves). Additionally, we have to handle ``party`` and ``lobby`` types that have different structures.

But I have one other idea; I'll update to something like ``insertMessage(destination:userId, data:...)``. This way both ``onMessagingReceivedEvent`` and ``insertSelfMessage`` will have clean ``data`` output for it, and the message will be stored in the correct user history still.

Hectate: OK, so I went a similar but different approach. Let me know what you think.

geekingfrog (approved):

Hectate:
@p2004a Do you have any thoughts?

src/renderer/store/users.store.ts
```diff
@@ -12,6 +12,14 @@ export const usersStore: {
     isInitialized: false,
 });
 
+export async function getUserById(userId: string) {
```
p2004a: dead code

src/renderer/store/users.store.ts
```diff
@@ -12,6 +12,14 @@ export const usersStore: {
...
+    return await db.users.get(userId);
+}
+
+export async function getUsersByIds(userIds: string[]) {
```
p2004a: function like this seems unnecessary, across the codebase, in components, `db.users` is queried directly from components.

src/renderer/store/chat.store.ts
```diff
@@ -3,131 +3,129 @@
...
+
+const chatDestinations = {
+    lobby: chatStore.lobbyChat,
+    party: chatStore.partyChat,
+    player: chatStore.userChats,
+};
+
+function requestSend(data: MessagingSendRequestData) {
+    try {
+        const response = window.tachyon.request("messaging/send", data);
```
p2004a: this misses `await`, code below will be run unconditionally.

src/renderer/store/chat.store.ts
```diff
@@ -3,131 +3,129 @@
...
 }
 
-export interface ChatMessage {
-    userId: string;
-    userName: string;
-    text: string;
-    timestamp: number;
+function requestSubscribeReceived(data?: MessagingSubscribeReceivedRequestData) {
+    try {
+        const response = window.tachyon.request("messaging/subscribeReceived", data ?? {});
```
p2004a: no `await`

src/renderer/store/chat.store.ts
```diff
@@ -3,131 +3,129 @@
...
-    },
+function setupMessageArray(userId: UserId) {
+    if (!chatDestinations["player"].get(userId)) {
+        chatDestinations["player"].set(userId, []);
+    }
+}
+
+// We want to start with an empty chat array if we join a new lobby/party chat,
+// so we give the UI a simple way to purge the old data just before joining.
+function clearLobbyChat() {
```
p2004a, line 133: dead code

src/renderer/store/chat.store.ts
```diff
@@ -3,131 +3,129 @@
...
+    chatStore.lobbyChat.length = 0;
+}
+
+function clearPartyChat() {
```
p2004a, line 140: dead code

src/renderer/store/chat.store.ts
```diff
@@ -3,131 +3,129 @@
...
-    Lobby = "lobby",
+function onMessagingReceivedEvent(data: MessagingReceivedEventData) {
+    console.log("Tachyon event: messaging/received:", data);
+    subsManager.attach(data.source.userId, chatSymbol);
+    if (data.source.type === "player") {
+        setupMessageArray(data.source.userId);
+        insertMessage(data, chatDestinations["player"].get(data.source.userId)!);
+    } else {
+        insertMessage(data, chatDestinations[data.source.type]);
+    }
```
p2004a: This logic should be fully contained in `insertMessage`, `setupMessageArray` shouldn't exist at all. It will be then a single function that takes `MessagingReceivedEventData` and puts it in correct place, reused between `insertSelfMessage` and this function.

src/renderer/store/chat.store.ts
```diff
@@ -3,131 +3,129 @@
...
+    party: chatStore.partyChat,
+    player: chatStore.userChats,
+};
+
+function requestSend(data: MessagingSendRequestData) {
+    try {
+        const response = window.tachyon.request("messaging/send", data);
+        console.log("Tachyon messaging/send:", response);
+        // We add our own message to the history because we do not get it from the server, unless it is targeted at ourselves.
+        if (data.target.type === "player" && data.target.userId !== me.userId) {
```
p2004a: This logic, except the `subsManager.attach(data.target.userId, chatSymbol);`, should be contained in `insertSelfMessage`.

src/renderer/components/navbar/NavBar.vue
```diff
@@ -164,14 +165,12 @@ function handleFriendsClick() {
 }
 
 const messagesUnread = computed(() => {
```
p2004a: nit:
```ts
const messagesUnread = computed(() => chatStore.userChats.values().some((m) => m.at(-1)?.seen === false));
```

src/renderer/components/navbar/Messages.vue
```diff
@@ -118,41 +138,31 @@ function focusTextbox(el: HTMLElement) {
...
+    // Until that time, we simply purge the chat entirely. It would not persists on restart
+    // of the client anyway, so we do not need to extend that at this time.
+    chat.clearUserChat(userId);
+}
+
+function hasUnseenMessage(messages: Message[]) {
+    // The most recent message should always be !seen when new
+    // So there's no need to loop, we just check the final message in the array
+    if (messages.length === 0) return false;
+    return !messages.at(-1)!.seen;
```
p2004a: nit: just `return messages.at(-1)?.seen === false` no need for length check

src/renderer/components/navbar/Friend.vue
```diff
@@ -155,9 +156,10 @@ async function viewProfile() {
 
 const toggleMessages = inject<Ref<((open?: boolean, userId?: string) => void) | undefined>>("toggleMessages")!;
 function sendMessage() {
-    // if (!api.session.directMessages.has(props.user.userId)) {
-    //     api.session.directMessages.set(props.user.userId, []);
-    // }
+    // If the userChat Map does not have this user yet, we populate it empty so it can open to that tab
+    if (!chatStore.userChats.get(props.userId)) {
```
p2004a: doesn't seem like this logic belongs here, I would expect ui instead to resolve to resolve to empty list if not set.

src/renderer/components/navbar/Messages.vue
```diff
@@ -91,24 +98,37 @@ const emits = defineEmits<{
     (event: "update:modelValue", open: boolean): void;
 }>();
 
+const displayNames = ref(new Map<UserId, string>());
```
p2004a: Is this intentional that it's not being recomputed when more display names are resolved and get to the database?

Hectate: I was having a lot of trouble with the ``displayNames`` being a computed property as I initially set it up. My solution ended up being to use the ``watch()`` (line 103) to update the ``displayNames`` ref anytime the ``chatStore.userChats`` updated.

src/renderer/store/chat.store.ts
Hectate: Both ``clearLobbyChat`` and ``clearPartyChat`` will be needed in the future. It is true that we do not have parties nor lobbies fully ready, so there's no place to use them at this time. However, it seemed sensible to include them now so that the next stage of working on either of those features would have the chat store set up and ready-to-go for it.

I can pull them out but I anticipate that we will need them both "soon".

src/renderer/store/users.store.ts
Hectate: This and ``getUsersByIds`` were convenience functions, nothing more. I can remove both if you feel they're redundant.

src/renderer/components/navbar/Messages.vue
p2004a: But as far as I understand the logic you aren't updating `displayNames` when a new entry is put in `db.users`, because you aren't using reactive view on the dixie database but abstraction over simple getter.

src/renderer/components/navbar/Friend.vue
Hectate: This is a result of how ``Messages.vue`` is designed. Line 10 shows where the loop is implemented directly on the Map of user chats.
```ts
<TabPanel v-for="[userId, messages] in chatStore.userChats" :key="userId">
```
As a result, if a user is not in the Map at all, there will be no tab created for that user. The client will not have access to a text entry field nor the "send" button for the targeted user. So I put the check/create right before the switch.

If you prefer for this to live inside of ``Messages.vue`` instead, I could move the check/create into it during the ``toggleMessages.value = (open...`` (lines 124 to 133). I could see that giving us the benefit of being able to to this from other places in the client easily (e.g. right-click on a user in a Lobby/Party and initiate a DM from a context menu).

src/renderer/components/navbar/Messages.vue
Hectate: I see what you mean. I will work on that issue.

src/renderer/store/chat.store.ts
p2004a: Ok, let's keep it.

src/renderer/components/navbar/Friend.vue
p2004a: Okey, I see what you mean. Let's instead export function `openChat` of whatever is the appropriate name from the chat store. Maybe the comment could be also adjusted a bit, mention directly in e.g. chat store that existence of entries control the presence of ui elements?

Hectate: After doing some digging into the current setup, and thinking about to implement it, I have come to believe that adding an ``openChat`` to the store is the wrong approach. We should ideally have the ``chat.store`` only be responsible for the retrieval, storing, and updating of data, and only allow the UI to decide how to present the data or trigger requests to update it.

So the initial concern is that we have a UI element that is directly modifying the data. I'm going to break that out into a ``addEmptyChat()`` which will separate the concerns. It will also ensure that if anything changes in the future about how we access and modify the chat data we can have a centralized place (``chat.store``) where we are touching it directly. Meanwhile, we will leave the UI to manage itself as it sees fit or changes in the future.

I did add a comment about the chat entries controlling the UI, but ultimately that's the UI's decision on how to handle the data as-stored anyway. This is true everywhere in the client anyway - for example the number of users in a Lobby or the number of Lobbies in the overall list.

p2004a (approved):
Pushed one small cleanup commit, LGTM, thank you and sorry for long wait time!

PR #570, opened 2026-04-23
Add polygon startbox preview support

burnhamrobertp:
### Work done

Renders polygon startbox regions on the map preview for maps with polygon configs (e.g., Cirolata 1.03). Companion to [beyond-all-reason/Beyond-All-Reason#7513](https://github.com/beyond-all-reason/Beyond-All-Reason/pull/7513) and [beyond-all-reason/BYAR-Chobby#1184](https://github.com/beyond-all-reason/BYAR-Chobby/pull/1184).

- Loads polygon configs from game archive (modside priority) or map archive (.sd7), matching the game-side priority order in `startbox_utilities.lua`
- Mapside configs use full Lua evaluation via fengari (Lua 5.3 in JS) with mock Spring/Game globals — handles conditionals, variable references, and computed values
- SVG polygon overlays on `MapBattlePreview` with team labels
- Generates engine startrects from polygon bounding boxes for game launch
- Caches configs by (mapName, gameVersion) with negative-cache and invalidation on map add/delete
- Startbox preset selector remains functional — selecting a rectangular preset overrides polygons, "Default Boxes" restores them
- Backwards compatible — maps without polygon configs behave exactly as before

#### Setup

Requires a game build from [beyond-all-reason/Beyond-All-Reason#7513](https://github.com/beyond-all-reason/Beyond-All-Reason/pull/7513) and a map with polygon startbox config (e.g., Cirolata 1.03 or Onyx Cauldron).

#### Test steps

- Load chosen map in skirmish. Polygon startboxes render on map preview with team labels.
- Select a rectangular preset from the startbox dropdown. Polygons clear, rectangles appear.
- Select "Default Boxes". Polygons restore.
- Load a rectangular-only map. Existing behavior unchanged.
- Start a skirmish game on Cirolata 1.03. Game loads with correct startrects.

### Related PRs

- Core Game - https://github.com/beyond-all-reason/Beyond-All-Reason/pull/7513
- Chobby Lobby - https://github.com/beyond-all-reason/BYAR-Chobby/pull/1184

### AI / LLM usage statement

Developed with GitHub Copilot (Claude). Used for implementation, debugging, and codebase research. All code was reviewed, tested in the lobby across multiple maps and configurations, and verified by the contributor.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,6 +24,7 @@
                 "date-fns": "^4.1.0",
                 "dexie": "^4.0.11",
                 "dompurify": "^3.2.6",
+                "fengari": "^0.1.5",
                 "flag-icons": "^7.5.0",
                 "glob-promise": "^6.0.7",
                 "howler": "^2.2.4",
@@ -7720,6 +7721,17 @@
                 "pend": "~1.2.0"
             }
         },
+        "node_modules/fengari": {
+            "version": "0.1.5",
+            "resolved": "https://registry.npmjs.org/fengari/-/fengari-0.1.5.tgz",
+            "integrity": "sha512-0DS4Nn4rV8qyFlQCpKK8brT61EUtswynrpfFTcgLErcilBIBskSMQ86fO2WVuybr14ywyKdRjv91FiRZwnEuvQ==",
+            "license": "MIT",
+            "dependencies": {
+                "readline-sync": "^1.4.10",
+                "sprintf-js": "^1.1.3",
+                "tmp": "^0.2.5"
+            }
+        },
         "node_modules/figures": {
             "version": "6.1.0",
             "resolved": "https://registry.npmjs.org/figures/-/figures-6.1.0.tgz",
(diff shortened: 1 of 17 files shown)


geekingfrog:
I haven't fully read the changes, but you are evaluating lua embedded in the map files to get the startbox positions???
That seems very very wrong, this type of data should exist elsewhere, in a data format.

There is already the logic to set startboxes for the maps, that doesn't require lua execution afaik.

burnhamrobertp:
> I haven't fully read the changes, but you are evaluating lua embedded in the map files to get the startbox positions??? That seems very very wrong, this type of data should exist elsewhere, in a data format.
> 
> There is already the logic to set startboxes for the maps, that doesn't require lua execution afaik.

This is true; and we could remove support for it (it would simplify all 3 PRs), but at some cost. It allows for some pretty powerful things, like changing the default starting boxes based on game type (or other LUA-accessible properties). Including it (and thus, having to add lua-parsing to this repo) was a compromise as supporting the broadest collection of the existing functionality as seemed practical.

If I recall correctly, there are a couple dozen maps that have polygon startboxes as part of the map archive, even presently. These are all done in a `mapconfig/map_startboxes.lua` so _some_ lua parsing is inevitable to some degree (even aside from startobxes, the lua-script parsing problem is kind of everywhere).

burnhamrobertp:
There is the data in rowy, which mapmakers are expected to put in. Strictly speaking there isn't anything lost if mappers provide this information there, and that data is then made available for the lobby's use (rather than parsing the lua).

The point I made above, though is still relevant. If we go this ^ route for startboxes, that's fine and should work? but what about the next thing after that? and the one after that? I fear eventually the avoidance of parsing lua from the map files is going to be difficult to avoid completely

burnhamrobertp:
On the assumption that this (related) PR https://github.com/beyond-all-reason/Beyond-All-Reason/pull/7516 is approved (I can't see a reason why it shouldn't; these are codepaths that have been dead for years), I'm going to modify this pull request to **remove** the modconfig support.

p2004a:
1. The current plan was to load startboxes from rowy, like we currently do, and pass via modoption like we do for startpos. Loading startboxes from modoption is effectively required to support changing startboxes dynamically, per game. We also do not want to have to modify map file to change startboxes.
2. I'm not interested at this stage of any dynamic, per game mode features for startboxes. This is a whole YANGI. Those ideas depend on way way more thing that simply do not exist.
3. I'm not going to merge the proposed way to evaluate Lua, even if we will want to evaluate Lua, which I'm considering because current parsing of code and extraction is extremaly fragile, the proposed way is not the solution and I will always require schema validation for all data extracted from Lua.

Issue #291, opened 2025-01-19
Dynamic configuration of lobby


p2004a:
There are many runtime parameters that are currently statically hardcoded in [`src/main/config`](https://github.com/beyond-all-reason/bar-lobby/tree/7d56252c3d583acb2a95694433ca4c4b075f0dc2/src/main). Changing any of them will require re-release of new version of lobby which is very suboptimal as they might change way more often then lobby codebase.

The values being hardcoded in code also might make it also harder for local development, or connecting to different stacks of infrastructure with the same lobby (e.g. prod and integration teiserver), so it would be good to be able to override those configurations without rebuild.

The issue here is to implement mechanism similar to the one in https://github.com/beyond-all-reason/spring-launcher:
- Dynamic configuration via JSON file
- Updates of the configuration from remote url
- Ability to override config via command line flag

In addition:
- Configuration needs to be described by JSON schema

keithharvey:
Working on this.

PR #572, opened 2026-04-27
Add offline resilience to startup pipeline

### LINKED ISSUE #333: Offline single player support
Once the game is fully installed once, it should be possible to play single player fully offline.

Right now the update/download functions are critical and block launching lobby. There might be also more things that break once the lobby is opened, so need just be tested.

burnhamrobertp:
Closes #333, related to #531

### Work done

Makes the startup pipeline resilient to network failures so the lobby doesn't crash or hang when offline. Once content is installed from a previous session, single player works without a connection.

This intentionally avoids the global `isOnline` flag approach from #531 -- per p2004a's review feedback, connectivity is per-service, not binary. Each network operation handles its own failure individually instead.

#### Auth resilience

`initMeStore()` runs before the UI mounts, so there's no way to show a retry dialog if auth renewal fails. Wrapped the `login()` call in a try/catch -- on failure the app continues unauthenticated and the user can retry from the UI later.

On the main process side, `auth.service.ts` was wiping credentials on any login error including transient network failures. Now it distinguishes network errors (`TypeError` from fetch) from actual auth rejections -- only wipes the refresh token when the server explicitly rejects it, so a temporary outage doesn't permanently log you out.

#### Maps store fallback

Wrapped `fetchAllMaps()` in `maps.store.ts` with a try/catch that falls back to IndexedDB cached map data from previous sessions. The store still initializes -- users see their previously-cached map list instead of a crash.

#### Fetch timeout

`fetchAllMaps()` in `maps.service.ts` had no timeout -- on a broken connection it would hang for the OS TCP timeout (minutes). Added an `AbortController` with a 15s timeout so it fails fast and the error propagates to the caller.

#### Map image worker fix

The worker in `map-image-worker.ts` was catching fetch errors but never posting failure messages back to the main thread. Promises in the `jobs` Map would hang forever. Fixed the worker to post `{ imageSource, error }` on failure, check `response.ok`, and use a 10s fetch timeout. On the main thread side, wired up the `worker.on("error")` and `worker.on("exit")` handlers (they were commented out) to reject all pending promises on fatal crashes.

#### Step-based InitialSetup

Reworked `InitialSetup.vue` from a sequential fire-and-forget pipeline into a step-based system. Each step (engine, game, maps, auto-update) is wrapped in a retry loop -- on failure, the user sees **Retry** and **Skip** buttons. Skip is gated on whether local content already exists (can't skip engine download if no engine is installed). A persistent **Play Offline** button is always visible to skip all remaining network steps at once.

Uses `Promise.race` with an abort signal so Play Offline works even if a step is mid-flight -- it stops blocking immediately instead of waiting for the in-flight request to finish or time out.

Also fixed a pre-existing issue where the map download check used `db.maps.count()` (all cached metadata) instead of filtering for actually installed maps.

#### Local testing

- Tested with network adapter disabled -- lobby loads, preloader completes with cached data, InitialSetup shows retry/skip UI on failures, Play Offline drops into the lobby
- All existing tests pass (28/28), typecheck clean, lint clean, prettier clean

### AI / LLM usage statement

Developed with GitHub Copilot CLI (Claude). Used during codebase research, implementation, and testing. All code was reviewed, directed, and verified by the contributor.


burnhamrobertp:
This PR does add localization strings. I'm not quite sure how to add them to Transifex myself (or if I even could)

geekingfrog:
Can you add at least a screenshot of the retry thingy when offline please?
I do not trust the "local testing" part when written by llm.
Then I'll have a look

burnhamrobertp:
Edit: Nevermind, was thinking of other work that had other UI changes. Yep I'll add the screenshots here, shortly.

Network disruption during optional step (maps or auto-update)
image

Network disruption during required step
image

Hectate:
I'll take a look at the code and do some testing soon. Meanwhile, some thoughts from what I've noticed otherwise:

- If we have a way to know if this is actually "Initial Setup" or simply "Updating" then we should ideally call it that on the screen, that way the "skip" option is exclusive to updating, not setup.
- The message "Step failed. You can retry or skip this step." for a _required_ step should say something like "Step failed. You can retry or play offline." since those are the options visible.

One consideration is that some game updates require an engine update also, so the two are irrevocably tied together and skipping one should skip the other as well. We could presume that the user will not be able to get either in this situation as evidenced by the failure, but it's a case to protect against.

Ultimately, I would like to make available a larger download package for offline users that contains client, game, engine, and a basic collection of maps (e.g. any required for scenarios at minimum); possibly in a portable app format (no install required). That way people with no internet access can still be handed a USB with the game and play some version of it. But that's an aside to this PR and not a priority at the moment.

Hectate:
###  Test 1: Launch with no internet (existing assets) ✅ 
Disabled network adapter and launched into a client that had existing assets. It reported that it was unable to connect to authenticate and I was able to enter offline mode.

### Test 2: First time launch with no internet ❌ 
Disabled network adapter and launched into a client with no existing state or assets. It reported "Something went wrong." and allowed me to choose "Retry" or "Play Offline". Upon choosing to play offline and then attempting to launch a Skirmish, the client crashed with error ``PromiseRejectionEvent {isTrusted: true, reason: Error: failed to access game version``.

We need to communicate that internet is required for initial setup and offline play until we can otherwise offer it upon an initial launch.

### Test 3: Network failure during download ✅ / ❌  
Launched a client with no existing state or assets. Mid-download the network adapter was disabled. The UI showed the _game_ download stopped at a partially complete percentage.. The download stopped as expected, although nothing clearly indicated that something was wrong. As in Test 2, I could still continue to play offline. I noted that there appeared to be no timeout for the game download to fail.
A second launch of the same client, with the network adapter re-enabled, was successfully able to download the files, etc.

### Test 4: Network interruption during initial download ❌ 
Launched a client with no existing state or assets. Mid-download the network adapter was disabled. The UI showed the _map_ download as failed and offered to retry. I re-enabled the network adapter and waited for it to connect, then tried to Retry, which advised me it failed. I repeatedly clicked Retry and eventually it was successful (?) and put me in the main menu. Immediately upon reaching the main menu, I noticed that the Notifications API had a toast disappearing that indicated that the _game_ had failed to download. See screenshot of console log below:
image
Following this, I was able to confirm that the game did not successfully download, because launching the skirmish screen crashed the client in the same way as Test 2 did.

Hectate:
### Test 5: Network degradation upon initial launch❓ 
I used Clumsy 0.3 to limit my bandwidth to 1 KB/s and launched a client with no prior state or assets. Engine and game downloads were slow, but reasonable. Upon reaching _map_ downloads, however, it immediately registered as failed and offered to Retry, Skip, or Play Offline. The first attempt I did a Retry and it seemed to be successful. The second attempt I did Skip instead and it also seemed to be okay.
The reason this isn't a full pass is because it shouldn't have an initial failure state on the optional map downloads - presumably just because it just took too long to download the game and engine? Screenshot of the console below:
image
Notably, the "Red Comet" map did appear in the Library > Maps view despite this error.

burnhamrobertp:
Thanks for the thorough tests and the breakdown of behavior, I've missed something in my testing strategy to not have encountered them myself; I'll run back through it and take care of them.

burnhamrobertp:
Test 4 I couldn't reproduce but is actually caused by the game failing to download and then the subsequent errors result from that, I believe. Fundamentally its the same issue, with the added map loading error on top

Test 5  isn't really solved; the list of map data during preload is not trivially small (950kb) and there was a 15s timeout. Raising it would hold up users; since this is during preload it blocks further initialization progress.

Hectate:
> Test 4 I couldn't reproduce but is actually caused by the game failing to download and then the subsequent errors result from that, I believe. Fundamentally its the same issue, with the added map loading error on top

Yes, if we protect the client from poor behavior if the game did not download successfully, that will resolve the issue there. I'm not sure why the map download failed multiple retries, but the initial failure is probably the same as Test 5; the timeout expired while the connection was interrupted.

> Test 5 isn't really solved; the list of map data during preload is not trivially small (950kb) and there was a 15s timeout. Raising it would hold up users; since this is during preload it blocks further initialization progress.

The issue this test raises is that immediately upon the map step starting it seems to register as failing already. I haven't checked the code; is the 15s timeout starting in the background while the game and engine are still downloading? If these steps are not running concurrently then the timeout should not begin until we reach the map step.

burnhamrobertp:
> The issue this test raises is that immediately upon the map step starting it seems to register as failing already. I haven't checked the code; is the 15s timeout starting in the background while the game and engine are still downloading? If these steps are not running concurrently then the timeout should not begin until we reach the map step.

There are two things covered by "downloading maps" that aren't well differentiated (either colloquially or in the interface):

* Download validated map list json file (950kb)
* Download the 2 default maps

The validated map list json (plus all the other stuff that happens during preload, such as fonts, indexDb init, map thumbnail downloads, replay directory scanning) starts almost immediately after you get into the client, even while the placeholder intro is fading in, then fading out. I believe this is the one you're referring to and seeing in test 5, though on my system I believe its failing before the intro fades out and we transition to the loading screen (which tracks and visually represents the loading status of engine, game data, downloading of 2 default maps

p2004a:
I've not read all conversation and so on, just something that caught my eye:

> starts almost immediately after you get into the client, even while the placeholder intro is fading in, then fading out

+1

https://github.com/beyond-all-reason/bar-lobby/blob/28e8900d53bab60437623633dbfd8150305e7492/src/renderer/components/misc/Preloader.vue#L35

is a problem, the `Preloader` should be 100% offline and not do any online work. The fetching of maps metadata should be part of the initial setup step that has all the other updates. It's also impacting my work on https://github.com/beyond-all-reason/bar-lobby/issues/291 because I would like to be able to fetch config update in initial setup *before* other steps run/be able to retrigger when there is update. I've looked very briefly into how to move that step into initial setup, but there are some weird/incorrect dependencies in the lobby and it needed some more work that I didn't get to...

burnhamrobertp:
This is existing behavior that wasn't touched in the scope of this work. Did we want to include its correction as part of this? Or remove the stub of an error-wrapping that I added into the preloader, on the assumption that we're going to move all async behaviors out of preloader?

> the Preloader should be 100% offline and not do any online work. The fetching of maps metadata should be part of the initial setup step that has all the other updates.

Should this be its own ticket?

Hectate:
> > the Preloader should be 100% offline and not do any online work. The fetching of maps metadata should be part of the initial setup step that has all the other updates.
> 
> Should this be its own ticket?

Personally, I'm fine with it being it's own issue or sub-issue to be resolved separately.

p2004a:
> Personally, I'm fine with it being it's own issue or sub-issue to be resolved separately.

+1

burnhamrobertp:
Were there any other changes or feedback that I had missed? I didn't have any further changes planned for this branch.

Hectate:
I'll rerun some tests this week and see if it's time to look at the code itself.

Hectate:
Apologies for this, but I just merged #575 which overlaps with this one somewhat. The ``Preloader.vue`` work has now been wrapped into ``IntialSetup.vue``. I still want to get this in next if we can resolve the new conflicts.

burnhamrobertp:
I'll take a look tomorrow (later today?) but it may not be until the  evening before I get to it. If you want to get to it sooner you are welcome to make changes to the pr as you see fit.

src/renderer/components/misc/InitialSetup.vue
```diff
@@ -31,45 +44,249 @@ const emit = defineEmits<{
...
+        resolveStepError("skip");
+        resolveStepError = null;
+    }
+}
+
+const AUTO_RETRY_ATTEMPTS = 3;
+const AUTO_RETRY_DELAY_MS = 2000;
+const STALL_THRESHOLD_MS = 10_000;
+
+function delay(ms: number): Promise<void> {
```
p2004a: use delay from `import { delay } from "$/jaz-ts-utils/delay";`

src/main/content/maps/map-content.ts
```diff
@@ -46,6 +46,9 @@ export class MapContentAPI extends PrDownloaderAPI<string, MapData> {
             }
         }
         log.debug("Scanning for maps");
+        // TODO: Verify file checksums (md5 from CDN /find API or maps-metadata) before considering
+        // a map installed. Partially downloaded or corrupted .sd7 files currently pass this scan
+        // if their header is parseable, leading to broken maps appearing as installed.
```
p2004a: This sounds wrong. Source?

1. Hashing all maps on load is too slow, not acceptable
2. pr-downloader does ~safe file operations, no partial files written with sd7 extension, hash verification.

src/main/content/maps/map-image-worker.ts
```diff
@@ -11,15 +11,25 @@ if (isMainThread) {
 
     const narrowedParentPort = parentPort;
 
+    const FETCH_IMAGE_TIMEOUT_MS = 10_000;
+
     // listen to messages from the main thread
     narrowedParentPort.on("message", async (imageSource: string) => {
+        const controller = new AbortController();
+        const timeout = setTimeout(() => controller.abort(), FETCH_IMAGE_TIMEOUT_MS);
```
p2004a: https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static

src/main/services/maps.service.ts
```diff
@@ -12,8 +12,19 @@ async function init() {
     await mapContentAPI.init();
 }
 
+const FETCH_MAPS_TIMEOUT_MS = 15_000;
+
 async function fetchAllMaps(): Promise<[MapData[], MapDownloadData[]]> {
-    const maps = await fetch("https://maps-metadata.beyondallreason.dev/latest/lobby_maps.validated.json");
+    const controller = new AbortController();
+    const timeout = setTimeout(() => controller.abort(), FETCH_MAPS_TIMEOUT_MS);
```
p2004a: https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static

src/main/services/auth.service.ts
```diff
@@ -19,7 +19,13 @@ function registerIpcHandlers() {
             startTokenRenewer((expiresIn / 2) * 1000);
         } catch (error) {
             log.error("Error during login");
-            accountService.wipe();
+            // TypeError indicates a network-level failure (fetch failed, DNS, timeout).
```
p2004a: That sounds wrong in comparison to what `TypeError` is supposed to mean https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError

If we throw that type in that cases that's likely instead a bug in our code.

src/renderer/components/misc/InitialSetup.vue
```diff
@@ -31,45 +44,249 @@ const emit = defineEmits<{
...
-onMounted(async () => {
-    console.debug("Initial setup");
-    if (!enginesStore.selectedEngineVersion || enginesStore.selectedEngineVersion.installed === false) {
-        state.value = "engine";
-        text.value = t("lobby.components.misc.initialSetup.downloadingEngine");
-        await window.engine.downloadEngine();
-        text.value = t("lobby.components.misc.initialSetup.installingEngine");
+let resolveStepError: ((action: "retry" | "skip") => void) | null = null;
+let abortSetup = false;
+let resolveAbort: (() => void) | null = null;
```
p2004a: why not abort controller but custom?

src/main/content/maps/map-content.ts
burnhamrobertp: I'll see if I still have the map in question, but I had somehow ended up with one in this state; where it read like a valid map but was otherwise corrupted. Practically speaking it may not be a real concern because I can't rule out the possibility that I was the one that corrupted it from my work in other repos or on other tooling. Rather than duplicate files everywhere I symlink my main maps directory from a central location.

src/main/services/auth.service.ts
burnhamrobertp: Many instances of `TypeError` from a fetch could indicate a bug in our code, but they can be raised from network connectivity issues as well, at least based on the fetch spec. I've not done exhaustive testing myself recently enough to be confident in stating the circumstances where it actually results in a TypeError; there are quite a few ways different browsers, devices, and network conditions can trigger different codepaths in modern browsers.

https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch#exceptions

I'll clean up the comment at the least, and there may be more followup changes to handle different types of TypeError exceptions in different ways.

burnhamrobertp: Actually the more I think on it the more likely it seems like this is splitting hairs and the spirit of your original comment is correct. Absent a more complete implementation keeping the existing behavior seems most appropriate.

src/renderer/components/misc/InitialSetup.vue
```diff
@@ -125,17 +213,83 @@ const preloadSteps: [string, () => Promise<unknown>][] = [
...
+        if (abortSignal.aborted) return "aborted";
+        console.error(`Setup step "${stage.label}" failed:`, error);
+        return "failed";
+    }
+}
+
+async function runStage(stage: DownloadStage): Promise<boolean> {
+    currentStage = stage;
+    let cameFromRetry = false;
+    while (!abortSignal.aborted) {
```
Hectate: The UI is freezing up during Retries that are failing, and I believe it's because so much is happening in this ``while``. Do we have an approach to this can can work asynchronously instead of locking it up?

Hectate:
My tests all passed. I think there may be an issue with the default map name, because pr-downloader does not seem to like it, but everything else seems good other than the UI freeze mentioned above.

## Tests
* Online startup with no assets ✅ 
* Online startup with existing assets ✅ 
* Offline startup with no assets ✅ 
* Offline startup with existing assets ✅ 
* Online startup after ALT+F4 during download in progress ✅ 
* Temporary network failure during initial startup ✅

Hectate (changes requested):

src/renderer/components/misc/InitialSetup.vue
burnhamrobertp: Are you meaning that the UI is actually becoming non-responsive? I'm not seeing this locally when I run the app with my networking disabled (in various ways).

Do you have reproduction steps you can provide, by chance? 

I do see the screen get into this sort of state where its definitely not making progress, but I can click on "Play offline" without issue

image

Hectate: I’ll have to check again later. The Loader stops spinning, which made me think it was frozen, but maybe there is something else happening.

Hectate: OK, I found out what happened. A static icon was added in #575 instead of the PrimeVue ``<ProgressSpinner />`` so we just have a non-spinning Icon 🤣 

We can fix that separately. Thanks!

Hectate (approved):

PR #574, opened 2026-05-05
Allow polygon start boxes

burnhamrobertp:
### Work done

Adds polygon and Catmull-Rom spline startbox rendering to the battle map preview. Maps that ship 3+ vertex polygons in their `startboxesSet` (via the maps-metadata schema PR) now render their actual region shape as an SVG overlay instead of collapsing to a bounding-box rectangle.

Reworks the previously-closed [#570](https://github.com/beyond-all-reason/bar-lobby/pull/570) from `master`. That earlier attempt loaded polygon configs directly from `.sd7` archives via Fengari Lua evaluation.

- New `src/renderer/utils/spline-tessellation.ts` — TypeScript port of `common/lib_spline.lua` (game-side), `LuaMenu/configs/gameConfig/byar/lib_spline.lua` (Chobby), and the in-line tessellator in the Rowy editor. Identical inputs produce vertex-identical output across all four consumers, so the lobby preview, the Chobby preview, and the in-game render of the same startbox can never visually disagree.
- `src/main/content/maps/map-metadata.ts` — added optional `strength?: number` to `StartBoxPoly` for spline anchors. Mirrors the schema in maps-metadata#615.
- `src/main/content/maps/box-utils.ts` — `spadsPointsToLTRBPercent` accepts N-point polygons (computes bounding box across all vertices). Removed dead `pointsToXYWH` helper that had been unused since introduction in `92f0f66`.
- `src/renderer/utils/start-boxes.ts` — `spadsBoxToStartBox` accepts N-point polygons; the existing min/max math already handled N points correctly, only the length-2 throw blocked it.
- `src/renderer/components/maps/MapBattlePreview.vue` — SVG polygon overlay with `pointer-events: none`, `viewBox="0 0 200 200"` matching the schema's coordinate space. When the active preset has any polygon-shaped startboxes, the rect-div drag handles are suppressed (read-only) — matching the BYAR-Chobby#1184 behaviour AntlerForce approved. Switching to a custom preset re-enables drag/resize.

#### Setup
Until [beyond-all-reason/maps-metadata#615](https://github.com/beyond-all-reason/maps-metadata/pull/615) merges and Rowy starts populating polygon vertex data, the only way to drive this end-to-end through the live data pipeline is via a custom build of maps-metadata that emits a `lobby_maps.validated.json` with 3+ point polys. Once that PR merges, polygon and spline data will flow through automatically.

For local testing today, manually edit a map's entry in your local `lobby_maps.validated.json` cache to insert 3+ point polys (with optional `strength` fields).

### Related PRs
- Core Game - https://github.com/beyond-all-reason/Beyond-All-Reason/pull/7513
- Chobby Lobby - https://github.com/beyond-all-reason/BYAR-Chobby/pull/1184
- Rowy Fork - https://github.com/p2004a/rowy/pull/1
- Maps-Metadata - https://github.com/beyond-all-reason/maps-metadata/pull/615

### AI / LLM usage statement
Assisted by GitHub Copilot (Claude Opus 4.7). Used for PR authoring, refactoring, and test scaffolding.

### DIFF
--- a/src/renderer/components/maps/MapBattlePreview.vue
+++ b/src/renderer/components/maps/MapBattlePreview.vue
@@ -39,7 +39,8 @@ SPDX-License-Identifier: MIT
                     :d="shape.path"
                     fill="rgba(255, 255, 255, 0.08)"
                     stroke="rgba(255, 255, 255, 0.9)"
-                    stroke-width="0.6"
+                    stroke-dasharray="6, 6"
+                    stroke-width="1.5"
                     vector-effect="non-scaling-stroke"
                 />
             </svg>
--- a/src/renderer/components/maps/MapBattlePreview.vue
+++ b/src/renderer/components/maps/MapBattlePreview.vue
@@ -8,9 +8,41 @@ SPDX-License-Identifier: MIT
     <div class="map-container">
         <div v-if="battleStore.battleOptions.map" class="map" :style="aspectRatioDrivenStyle">
             <img loading="lazy" :src="mapTextureUrl" />
-            <div v-if="battleStore.battleOptions.mapOptions.startPosType === StartPosType.Boxes && boxes" class="boxes">
+            <!--
+            When the active preset has polygon-shaped startboxes (3+ vertex
+            ring), render a read-only SVG overlay only — the rect-div drag
+            handles are suppressed. The user "breaks free" into custom rect
+            edit mode by picking a different preset in the map options modal.
+            This matches the BYAR-Chobby#1184 behaviour reviewers approved.
+            For pure-rectangle presets (or custom mode), the existing rect
+            divs render with full drag/resize affordance unchanged.
+            -->
+            <div
+                v-if="battleStore.battleOptions.mapOptions.startPosType === StartPosType.Boxes && boxes && !polygonPresetActive"
+                class="boxes"
+            >
                 <MapBattlePreviewStartBox v-for="(box, i) in boxes" v-startBox="box" :key="`box${i}`" :id="i" :box="box" />
             </div>
+            <svg
+                v-if="
+                    battleStore.battleOptions.mapOptions.startPosType === StartPosType.Boxes &&
+                    polygonPresetActive &&
+                    polygonOverlays.length > 0
+                "
+                class="polygon-overlay"
+                viewBox="0 0 200 200"
+                preserveAspectRatio="none"
+            >
+                <path
+                    v-for="shape in polygonOverlays"
+                    :key="`poly${shape.index}`"
+                    :d="shape.path"
+                    fill="rgba(255, 255, 255, 0.08)"
+                    stroke="rgba(255, 255, 255, 0.9)"
+                    stroke-width="0.6"
+                    vector-effect="non-scaling-stroke"
+                />
+            </svg>
             <div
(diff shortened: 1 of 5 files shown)


geekingfrog:
I don't have a  lobby_maps.validated.json file on my setup, where is it supposed to be and how do I get it?

Also, do you actually read the PR description when you create it? I find it's largely useless, especially the "summary of changes" file by file, that's what the diffs are for.
There are some useful info after, but it's so easy to glaze over it because of all the slop before.
That also gives off a very strong vibe of "I can't be arsed with summarizing my changes for reviewing, gl hf suckers"

burnhamrobertp:
> I don't have a lobby_maps.validated.json file on my setup, where is it supposed to be and how do I get it?

It should download automatically as part of the preload; the code is in `maps.service.ts` ; I don't believe it gets downloaded to disk, though, its probably in the `IndexedDB`

edit: I just double-checked, you won't see it in the network panel because the fetch occurs via the IPC (which runs via node). You can verify that its occurred, however, by checking Application -> Storage -> IndexDB -> BarLobby -> Maps

image

---

> Also, do you actually read the PR description when you create it? I find it's largely useless, especially the "summary of changes" file by file, that's what the diffs are for. There are some useful info after, but it's so easy to glaze over it because of all the slop before. That also gives off a very strong vibe of "I can't be arsed with summarizing my changes for reviewing, gl hf suckers"

So yes, both when I have the agent summarize for me on a PR I'm putting in, but also yes when I look at others' work. Its where I first found I liked the pattern. In this instance the summary of changes is probably a bit too long and I should've trimmed it up, given that the changes are only `+222 -23`. Generally one of the things I'm doing when reviewing is looking at how all the changes fit together to achieve the stated changed that the PR is for, and that's easier to do up front on the PR description so that the review isn't starting from naught. Sometimes I'll do it by commit, other times by changed file, or (if its simple enough) just via a few sentences (if the description wasn't enough).

I understand the notion you're referring to; certainly I've seen the usage of AI tooling to be along that vein, but its not how I think of it.

src/renderer/components/maps/MapBattlePreview.vue
```diff
@@ -8,9 +8,41 @@ SPDX-License-Identifier: MIT
...
+            >
+                <path
+                    v-for="shape in polygonOverlays"
+                    :key="`poly${shape.index}`"
+                    :d="shape.path"
+                    fill="rgba(255, 255, 255, 0.08)"
+                    stroke="rgba(255, 255, 255, 0.9)"
+                    stroke-width="0.6"
+                    vector-effect="non-scaling-stroke"
+                />
```
Hectate, line 45: Just to make the visual a little more consistent, will you do the following for this section?
* ``stroke-dasharray="6, 6"``
* ``stroke-width="1.5"``

Example of what this looks like, a lot closer to the rectangles' CSS appearance.
Image

Hectate: I had the code so I went ahead and sent in the commit. 🤷

Hectate (commented):
Looks good from this point, but we will have to do a lot of validation once the rest of the pipeline is working properly.

Hectate (approved):

PR #575, opened 2026-05-11
Improved progress bar while downloading

### LINKED ISSUE #415: Better progress indicators for downloading/extracing etc game
The progress indicator for download of engine/map/game are very barebones, just percentage. It would be useful to provide more information just like launcher does.

Moreover, currently the game *extracting* from data.7z doesn't have any indicator, I'm not sure we can have progress there, but at least better message and some spinner would be good.

BElluu:
Close #415 

The progress during download and extraction has been refined. During the download, we now show the downloaded and remaining MB, the download percentage, and the remaining time. A loader spinner is present during extraction.

Initial downloading game

image

Downloading map

image


src/renderer/components/misc/InitialSetup.vue
```diff
@@ -33,20 +45,103 @@ const emit = defineEmits<{
...
+function buildProgressText(download: DownloadInfo): string {
+    const now = Date.now();
+    const key = download.name;
+    const prev = speedTracker.get(key);
+    let speed = 0;
+    if (prev) {
+        const dt = (now - prev.prevTime) / 1000;
+        if (dt > 0.25) {
+            const raw = (download.currentBytes - prev.prevBytes) / dt;
+            speed = raw > 0 ? prev.speed * 0.7 + raw * 0.3 : prev.speed * 0.9;
```
Hectate: Was there any particular reason for these float values in particular? From my reading, you're giving a stronger weight to past speed, and applying a decay to past speed if there was no progress made since the last update (which is every quarter second it seems?). Is this just a matter of what feels right in the UI?

src/renderer/components/misc/InitialSetup.vue
```diff
@@ -33,20 +45,103 @@ const emit = defineEmits<{
...
+}
+const speedTracker = new Map<string, SpeedEntry>();
+
+const progressDetail = computed(() => {
+    const download = currentDownload.value;
+    if (!download || download.currentBytes === 0) return t("lobby.navbar.downloads.starting");
+    return buildProgressText(download);
+});
+
+function buildProgressText(download: DownloadInfo): string {
```
Hectate: This function should either return an object with the values used to generate the string, or return a string that's already been internationalized. We can use [Interpolated Strings](https://github.com/beyond-all-reason/bar-lobby/wiki/Internationalization:-translation-strings-in-bar%E2%80%90lobby#interpolated-strings) combined with the values to make sure that however this gets displayed it will be correct in all forms.

src/renderer/components/navbar/Downloads.vue
```diff
@@ -59,25 +60,79 @@ const toggleMessages = inject<Ref<(open?: boolean, userId?: number) => void>>("t
...
+function formatSpeed(bps: number): string {
+    if (bps >= 1024 * 1024) return `${(bps / (1024 * 1024)).toFixed(1)} MB/s`;
+    if (bps >= 1024) return `${(bps / 1024).toFixed(1)} KB/s`;
+    return `${bps.toFixed(0)} B/s`;
+}
+
+function formatEta(seconds: number): string {
+    if (seconds >= 3600) return `${Math.floor(seconds / 3600)}h ${Math.floor((seconds % 3600) / 60)}m`;
+    if (seconds >= 60) return `${Math.floor(seconds / 60)}m ${Math.floor(seconds % 60)}s`;
+    return `${Math.floor(seconds)}s`;
```
Hectate: These functions are duplicated between ``IntialSetup.vue`` and ``Downloads.vue`` files. If this is going to be re-used then we should extract the functions out of both and make them part of the ``download.store.ts`` instead. This will allow us to import them in the .vue files for use and have only one place to touch them.

Hectate (changes requested):
See comments. I'll also do some testing on the PR as written and reply here if anything catches my attention.

BElluu:
Thank you for review. I will look at this :)

src/renderer/components/misc/InitialSetup.vue
BElluu: These are standard Exponential Moving Average (EMA) parameters. 0.3 / 0.7 is an EMA with alpha = 0.3 — it smooths out speed readings so the UI doesn't jump erratically on every tick. The 0.9 decay factor handles stalls gracefully: instead of dropping to zero immediately when no bytes arrive, the displayed speed fades out gradually.

I've extracted these into named constants (EMA_ALPHA = 0.3, STALL_DECAY = 0.9) in the new `useDownloadProgress `composable to make the intent clearer.

src/renderer/components/misc/InitialSetup.vue
BElluu: Fixed

src/renderer/components/navbar/Downloads.vue
BElluu: Agreed on extraction. I went with a new composable `src/renderer/composables/useDownloadProgress.ts` rather than `download.store.ts`

BElluu:
I think after fixes and improvements, that look's much better. Just take my branch and run lobby :)

src/renderer/assets/languages/en.json
Hectate, line 1: Changes to this file aren't reflected in any change to ``lang\en\lobby.json``, which means they'll be lost the next time we run the generate script. See the[ i18n page](https://github.com/beyond-all-reason/bar-lobby/wiki/Internationalization:-translation-strings-in-bar%E2%80%90lobby#creating-new-translation-strings) for more details.

Hectate (changes requested):
I haven’t had a chance to do testing but I’m going to drop this in so you can update the i18n stuff without waiting on me.

src/renderer/assets/languages/en.json
BElluu: Thank you :) I didn't know about it. Fixed!

Hectate (approved):
Looks good from here. Thanks!

PR #576, opened 2026-05-12
Calculate checksum for engine, game and map files

### LINKED ISSUE #317: Trigger engine archive checksumming
Engine implemented a feature in https://github.com/beyond-all-reason/spring/issues/1903 that allows to trigger creating a checksum of archive before game starts.

This will be important to speed up game loading times and make sure that players don't time out when trying to join game.

BElluu:
The changes introduce checksum calculation immediately after downloading the engine, game or map.

Closed #317


src/main/content/engine/engine-content.ts
```diff
@@ -124,6 +125,36 @@ export class EngineContentAPI extends AbstractContentAPI<string, EngineVersion>
...
+                if (code !== 0) log.warn(`calcChecksum exited with code ${code} for: ${archiveName}`);
+                resolve();
+            });
+            proc.on("error", (err) => {
+                log.warn(`calcChecksum failed for ${archiveName}: ${err}`);
+                resolve();
+            });
+        });
+    }
+
```
Hectate: If this is all going to be shared between engine, games, and maps, then it should be taken out of ``engine-content.ts`` entirely. I recommend that it become it's own utility file inside ``src/main/utils/checksums.ts``.

Hectate (changes requested):

src/main/content/engine/engine-content.ts
BElluu: You are right. Fixed :)

src/main/utils/checksums.ts
```diff
@@ -0,0 +1,40 @@
...
+    log.debug(`Calculating checksum for: ${archiveName}`);
+    return new Promise<void>((resolve) => {
+        const proc = spawn(binaryName, args, {
+            cwd: enginePath,
+            stdio: "pipe",
+            env: { ...process.env, SPRING_DATADIR: ASSETS_PATH },
+        });
+        proc.on("exit", (code) => {
+            if (code !== 0) log.warn(`calcChecksum exited with code ${code} for: ${archiveName}`);
+            resolve();
```
Azzurite, line 35: Should it really resolve successfully when it exited with a non-0 exit code? Just asking, might be fine.

src/main/utils/checksums.ts
```diff
@@ -0,0 +1,40 @@
...
+        });
+        proc.on("error", (err) => {
+            log.warn(`calcChecksum failed for ${archiveName}: ${err}`);
+            resolve();
```
Azzurite, line 39: Should it really resolve successfully when it encountered an error? Just asking, might be fine.

src/main/content/game/game-content.ts
```diff
@@ -35,6 +37,13 @@ export class GameContentAPI extends PrDownloaderAPI<string, GameVersion> {
         await this.initLookupTables();
         await this.scanPackagesDir();
         await this.scanLocalGames();
+
+        engineContentAPI.onDownloadComplete.add((downloadInfo) => {
+            for (const gameVersion of this.availableVersions.keys()) {
+                calcChecksum(downloadInfo.name, gameVersion).catch(() => {});
```
Azzurite: I don't know what the convention has been so far, but generally, I would say an empty block almost always needs a comment explaining why it is empty, and in this case specifically because we're just ignoring errors basically.

If you want to implement later what to do when the checksum calculation fails, I would add a `// TODO` at least.

Right now, the checksum function never throws an exception. So if `calcChecksum` should be like this and never fail, then you should also not silently `catch` it here, and remove the `catch` block altogether.

(Repeat this comment for the other places where this was done)

src/main/utils/checksums.ts
```diff
@@ -0,0 +1,40 @@
...
+const log = logger("checksums.ts");
+
+let checksumQueue: Promise<void> = Promise.resolve();
+
+export function calcChecksum(engineVersion: string, archiveName: string): Promise<void> {
+    checksumQueue = checksumQueue.then(() => runChecksumProcess(engineVersion, archiveName));
+    return checksumQueue;
+}
+
+function runChecksumProcess(engineVersion: string, archiveName: string): Promise<void> {
```
Azzurite, line 19: I feel like this needs a comment explaining that by triggering this, the engine creates a cache file that speeds up loading later.

Otherwise, just by reading this code, this is not clear at all. Of course it's possible to find out if you go into --calc-checksum engine code or already know this, but otherwise you're out of luck.

It just looked weird to me as someone new looking at this code, and I did actually have to spend the time to look at the linked issue and stuff to understand why there's nothing "done" with the result of the invocation.

BElluu: Added comment :)

src/main/content/game/game-content.ts
BElluu: Removed all unused catches.

src/main/utils/checksums.ts
BElluu: Yea this is ok.  This is background operation, we just log errors like warns because we won't abort next calculating checksums operations. `resolve()` just inform  `Promise<void>` we finished and can go forward. 

Alternative is `reject(err)` when we want use `.catch()` but we wont. We intentionally use `resolve()` even on error, because we want subsequent checksums in the queue to continue to count, regardless of whether the previous one was successful.

src/main/utils/checksums.ts
BElluu: as above

Hectate (approved):
Passed testing, and definitely speeds up load times for newly downloaded assets. Thanks!

PR #577, opened 2026-05-14
Allow set custom path location for assets

### LINKED ISSUE #387: Allow customization of install directories

BElluu:
Closed #387 
I added the option for players to choose where BAR stores game files - the engine, maps and game data. On the first launch, the app now asks where you want everything downloaded. You can also change the location later in settings, with three options:

- move everything to the new folder and delete the old one
- copy everything and keep the original as a backup
- just point the app at a new folder without touching any files

Under the hood, I reworked how the app tracks the assets path so it can be changed at runtime without a full restart - only the renderer reloads. I aslo added proper input validation to prevent selecting dangerous paths, error handling so the UI doesn't get stuck if something goes wrong during a transfer, and progress reporting so you can see how the file copy is going.  

You can see how it works here - the video is too large to link to github ;(

[Video](https://drive.google.com/file/d/1NKgHb2nKIeDBEdFfnzmksczD8vRL8__-/view)

### AI / LLM usage statement:
Discovery how application downloading assets and how uses specific paths. In additional, AI helped me to do security check, to prevent path traversal etc. Used Claude Code.


p2004a:
Related issue https://github.com/beyond-all-reason/bar-lobby/issues/387

BElluu:
@Hectate looks like something went wrong with merge conflicts ? :) In this case we change static fields for functions
```
export const getEnginePath = () => path.join(ASSETS_PATH, "engine");
export const getPackagePath = () => path.join(ASSETS_PATH, "packages");
export const getPoolPath = () => path.join(ASSETS_PATH, "pool");
export const getRapidIndexPath = () => path.join(ASSETS_PATH, "rapid");
export const getMapsPaths = (): readonly string[] => [path.join(WRITE_DATA_PATH, "maps"), path.join(ASSETS_PATH, "maps")];
export const getGamePaths = (): readonly string[] => [path.join(WRITE_DATA_PATH, "games"), path.join(ASSETS_PATH, "games")];
```

Should I look at this or are you already messing with it?

Hectate:
> @Hectate looks like something went wrong with merge conflicts ? :) In this case we change static fields for functions
> 
> ```
> export const getEnginePath = () => path.join(ASSETS_PATH, "engine");
> export const getPackagePath = () => path.join(ASSETS_PATH, "packages");
> export const getPoolPath = () => path.join(ASSETS_PATH, "pool");
> export const getRapidIndexPath = () => path.join(ASSETS_PATH, "rapid");
> export const getMapsPaths = (): readonly string[] => [path.join(WRITE_DATA_PATH, "maps"), path.join(ASSETS_PATH, "maps")];
> export const getGamePaths = (): readonly string[] => [path.join(WRITE_DATA_PATH, "games"), path.join(ASSETS_PATH, "games")];
> ```
> 
> Should I look at this or are you already messing with it?

The checksum file didn't account for these changes, so they're still pointing to constants that don't exist. It appears to be a simple fix if you want to throw it in. I've already done it myself on the version I'm testing.

Hectate:
So that resolves the typecheck error, but we will have a deeper issue; the checksum paths will be wrong.

For example, here is a cached map:
```
		{
			name = "quicksilver_remake_1.24.sd7",
			path = [[D:\git\bar-lobby\assets\maps\]],
			modified = "1780694916",
			checksum = "d5591dd3f1f412ee2d3ed41853c359f7f05330d7d50331bda69b77b2f7ab5208ca11605b047f699a6a669b2c40674ad01881dc354ef21ae5311421dc597c31aa",
			archivedata = {
				author = "IceXuick, original from Cheesecan",
				autoshowmetal = true,
				description = "Quicksilver V1 Remake | BAR Certified",
				extractorradius = 120.000000,
				gravity = 120.000000,
				mapfile = "maps/QSR.smf",
				maphardness = 100.000000,
				maxmetal = 1.300000,
				modtype = 3,
				name = "Quicksilver Remake 1.24",
				name_pure = "Quicksilver Remake",
				notdeformable = false,
				shortname = "QSR",
				tidalstrength = 10.000000,
				version = "1.24",
				voidwater = false,
			},
		},
```
This is located at ``state\data\cache\ArchiveCache20.lua`` and we can see that it refers to an ``asset`` path that we can now move in this PR. Moving files will negate any benefits gained from the checksum. I think we would have to purge and regenerate any checksums, perhaps?

BElluu:
I think the checksum should be recalculated when changing the location. Regardless, changing the asset location is the most important thing right from the start. If someone wants to change the location in the settings, they should be aware that it will take a while longer due to the checksum recalculation. It's still better than waiting when joining a match :)

src/renderer/components/navbar/Settings.vue
```diff
@@ -45,17 +45,66 @@ SPDX-License-Identifier: MIT
...
+
+            <template v-if="pendingAssetsPath && pendingAssetsPath !== currentAssetsPath">
+                <div></div>
+                <div class="pending-section">
+                    <Textbox :model-value="pendingAssetsPath" readonly class="path-input" />
+                    <div v-if="isTransferring" class="copy-progress">
+                        <div class="progress-bar">
+                            <div class="progress-fill" :style="{ width: copyProgressPercent + '%' }"></div>
+                        </div>
+                        <small>{{ t("lobby.navbar.settings.transferringFiles") }} {{ copyProgressPercent }}%</small>
```
Hectate: This is the wrong approach to building a translated string. Some languages may need this arranged differently to read properly. Use an[ interpolated string](https://github.com/beyond-all-reason/bar-lobby/wiki/Internationalization:-translation-strings-in-bar%E2%80%90lobby#interpolated-strings) instead of two separate strings/values.

src/main/services/paths.service.ts
```diff
@@ -0,0 +1,129 @@
...
+    await settingsService.updateSettings({ assetsPath: newAssetsPath });
+    setAssetsPath(newAssetsPath);
+    await engineService.reinit();
+    await mapsService.reinit();
+    await gameService.reinit();
+}
+
+function registerIpcHandlers(webContents: BarIpcWebContents) {
+    ipcMain.handle("paths:selectFolder", async () => {
+        const { canceled, filePaths } = await dialog.showOpenDialog({
```
Hectate, line 95: Small thing; if we add a ``defaultPath`` then Electron will try open the file explorer to that path instead of finding some default directory for the explorer.

```ts
const { canceled, filePaths } = await dialog.showOpenDialog({ defaultPath: getAssetsPath(), properties: ["openDirectory"] });
```

Hectate (changes requested):
Pretty close, just a couple of minor things to handle.

BElluu:
I fixed things about strings. Additionally I found two bugs
1. Wrong assets path in settings. This was because the path was from the application's initialization stage and the settings didn't recognize that we had changed it (first run). Now this state is refreshed.
2. Sometimes I got long (20 sec) black screen in app, after change assets path. Application did full re-render. I changed it to refresh states and looks much better, because app is ready after change assets path instantly :)

I tested it and checksums and paths in ArchiveCache20.lua are correct after change path.

p2004a:
Have you ensured expected behavior, even just prevent doing assets move, when there is ongoing download/download started?

BElluu:
> Have you ensured expected behavior, even just prevent doing assets move, when there is ongoing download/download started?

You are right. I did some improvements.
1. User cannot browser path and apply change path while downloading is in progress - buttons are disabled
2. User cannot start download while changing path is in progress. - buttons are disabled

image

Hectate (approved):

Hectate:
Thanks!

Issue #38, opened 2022-08-28
Ingame settings editor


Jazcash:

Hectate:
Portions of the client will require non-public development to determine requirements, scope, visual design, and user experience. Because public contributions can not be accepted until that is complete, this issue is being closed to ensure volunteers do not spend time on pull requests that cannot be merged. When and where possible, new issues will be created with adequate information for volunteers to contribute to the final requirements.

PR #578, opened 2026-05-15
Set correct springsettings for engine

### LINKED ISSUE #410: Set correct `springsettings.cfg` for engine
Currently launcher has a mechanism https://github.com/beyond-all-reason/spring-launcher/blob/master/src/springsettings.js where it applies default engine settings before staring up the engine. The value of current setting we have is in https://github.com/beyond-all-reason/BYAR-Chobby/blob/master/dist_cfg/config.json in `default_springsettings` key.

This issue is to set up proper default engine settings for BAR in new lobby too. It would be also awesome if we actually audited what from current spring settings is still good with the help of game and engine devs, but it can be done as a separate task to make scope of this one smaller.

This is _related_ to https://github.com/beyond-all-reason/bar-lobby/issues/38 some in-game settings are actually engine settings, other are configuring the game.

BElluu:
Closed #410

### DIFF
--- a/src/main/game/springsettings.ts
+++ b/src/main/game/springsettings.ts
@@ -1,0 +1,150 @@
+// SPDX-FileCopyrightText: 2026 The BAR Lobby Authors
+//
+// SPDX-License-Identifier: MIT
+
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
+import { WRITE_DATA_PATH } from "@main/config/app";
+import { logger } from "@main/utils/logger";
+
+const log = logger("main/game/springsettings.ts");
+
+const SPRINGSETTINGS_PATH = path.join(WRITE_DATA_PATH, "springsettings.cfg");
+
+// Default engine settings for BAR, sourced from:
+// https://github.com/beyond-all-reason/BYAR-Chobby/blob/master/dist_cfg/config.json
+const BAR_SPRINGSETTINGS_DEFAULTS: Record<string, string | number> = {
+    AdvUnitShading: 1,
+    AdvMapShading: 1,
+    AdvSky: 0,
+    AllowDeferredMapRendering: 1,
+    AllowDeferredModelRendering: 1,
+    BuildWarnings: 0,
+    BumpWaterBlurReflection: 1,
+    BumpWaterDepthBits: 16,
+    BumpWaterShoreWaves: 1,
+    BumpWaterTexSizeReflection: 1024,
+    CamTimeExponent: 4.0,
+    CamTimeFactor: 1.0,
+    CrossAlpha: 0,
+    CubeTexSpecularExponent: 100,
+    CubeTexGenerateMipMaps: 1,
+    CubeTexSizeReflection: 1024,
+    DisplayDebugPrefixConsole: 0,
+    DualScreenMiniMapOnLeft: 1,
+    FPSFOV: 90,
+    FeatureDrawDistance: 10000,
+    FeatureFadeDistance: 10000,
+    ForceDisableShaders: 0,
+    Fullscreen: 0,
+    FullscreenEdgeMove: 1,
+    GrassDetail: 0,
+    GroundDecals: 2,
+    GroundDetail: 200,
+    GroundScarAlphaFade: 1,
+    HangTimeout: 30,
+    HardwareCursor: 1,
+    InitialNetworkTimeout: 0,
+    LODScale: 1.0,
+    LODScaleReflection: 1.0,
+    LODScaleRefraction: 1.0,
+    LODScaleShadow: 1.0,
+    LuaGarbageCollectionMemLoadMult: 2.0,
+    LogFlush: 0,
+    LuaShaders: 1,
+    LoadingMT: 0,
+    MaxParticles: 20000,
+    MaxSounds: 256,
+    MaximumTransmissionUnit: 0,
+    MaxTextureAtlasSizeX: 8192,
+    MaxTextureAtlasSizeZ: 8192,
+    MiddleClickScrollSpeed: -0.005,
+    MinimapOnLeft: 1,
+    MouseDragScrollThreshold: 0.3,
+    MoveWarnings: 0,
+    MSAA: 1,
+    MSAALevel: 4,
+    NormalMapping: 1,
+    ReconnectTimeout: 0,
+    Roam: 1,
+    ReflectiveWater: 3,
+    ScrollWheelSpeed: -20,
+    Shadows: 1,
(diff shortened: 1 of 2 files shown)


Hectate:
Before I take a look at the code, quick question for @p2004a is whether or not this was intended to be hard-coded into the client like this? I feel like the ideal approach would be that the game/engine comes with it's own set of default settings that are used based on the versions being launched. Of course, an initial pass might be to do this hard-coding and then evaluate a way to extract the preferred default values from the game/engine once it's available (with these as a fallback).

src/main/game/springsettings.ts
```diff
@@ -0,0 +1,150 @@
...
+            }
+        }
+        if (changed) {
+            writeSettings(settings, SPRINGSETTINGS_PATH);
+            log.info("Applied default springsettings");
+        }
+    } catch (err) {
+        log.error(`Failed to apply default springsettings: ${err}`);
+    }
+}
```
Hectate, line 150: There's no reason for us to have spring settings with their own separate methods to read/write to disk. These can directly use ``src/main/services/settings.service.ts`` to handle the OS side of things.
Honestly, this entire file can go down to just an exported object that is grabbed only when the settings need to be set to the default. During the ``src/renderer/store/settings.store.ts > initSettingsStore()`` we should just retrieve the default values if the settings are not already saved for the user.
Later we can add a UI button that "resets to default" and restores from this list also.

Hectate (changes requested):

src/main/game/springsettings.ts
p2004a: I'm not sure I understand this comment. Spring settings have to be stored in concrete format dictated by engine in concrete location, so some of those functions are unavoidable.

Can you clarify please?

p2004a:
> Before I take a look at the code, quick question for @p2004a is whether or not this was intended to be hard-coded into the client like this?

I thought it would be part of the dynamic configuration, but that doesn't exist so it can be ok I guess.

> I feel like the ideal approach would be that the game/engine comes with it's own set of default settings that are used based on the versions being launched.

That's a good idea, I've somehow not considered that just because how it was handled in the old system, but the old system had a constraint that lobby doesn't have (chobby runs *in* engine, so it couldn't apply engine settings that were required before engine start) 

> Of course, an initial pass might be to do this hard-coding and then evaluate a way to extract the preferred default values from the game/engine once it's available (with these as a fallback).

Yeah, that's related to "It would be also awesome if we actually audited what from current spring settings is still good with the help of game and engine devs, but it can be done as a separate task to make scope of this one smaller." I wrote in the issue. I preferred to have something, to have settings ok.

At the same though, look like people are watching replays with engine without it just fine, so maybe it's not really required to be done quickly.

src/main/game/springsettings.ts
Hectate: The functions above (``readSettings``, ``writeSettings``, and ``applyDefaultSpringsettings``) are all doing custom reading and writing to/from disk. This is already a solved problem with the ``settings.service.ts`` handling the act of putting them into JSON or reading them from JSON when we need to save settings.
So what I'm saying I think should happen here:
1. Take the existing ``BAR_SPRINGSETTINGS_DEFAULTS`` object and associated key/value pairs and make them an export
2. Remove the functions noted above.
3. Edit ``settings.service.ts`` to include a ``springSettings`` object.
4. If upon initialization there is nothing in ``springSettings`` we pull the default settings from ``BAR_SPRINGSETTINGS_DEFAULTS`` and apply them.
5. Later, we add a "Reset to Defaults" UI button that also reads ``BAR_SPRINGSETTINGS_DEFAULTS`` and applies them.

p2004a: `springsettings.cfg` is a file that's being read directly by engine, in a very specific format. There must be a dedicated piece of code in lobby that is able to write and potentially read `springsettings.cfg`.

What you are describing though has relation here and to https://github.com/beyond-all-reason/bar-lobby/issues/38. This is for when we would have in-lobby settings editor, and then it's likely that it should go via `settings.service.ts` but dedicated functions to serialize engine settings to `springsettings.cfg` will have to still exist.

p2004a: Why "then it's *likely* that it should" -> the interface how in-lobby and in-game settings editing will work, and interact with each other afaik isn't yet fully defined. `settings.service.ts` might remain for lobby settings only, and game/engine settings might need a separate service. I don't know yet.

Hectate: Gotcha. I'll mark this resolved and re-review from that perspective.

Hectate:
I did some quick testing and I couldn't tell if the default settings were being applied or not, or if we were just saving a file to disk and ignoring it. I'll do some further testing when I have an opportunity, but right now I think that's the only question to be answered that's going to hold this back.

BElluu:
I've tested the changes and **they work correctly**.

How it works:
`applyDefaultSpringsettings()` runs before Spring launches and writes BAR defaults only for keys that are missing from `springsettings.cfg`. This means it sets sensible defaults for new installations without overriding any settings the user has already changed.
Spring confirms it loaded our values at startup - you can see them listed under **<User Config>** in the engine log `state/data/infolog.txt`.

How to test:
1. Delete or rename your `springsettings.cfg` (or just remove the `ScrollWheelSpeed` line from it)
2. Launch lobby and start the game
3. Open in-game settings - scroll speed should show **20** - the absolute value of our default **-20**
4. To confirm the value acctually changed, try change `ScrollWheelSpeed` in code e.g **-40**, delete the key from `springsettings.cfg` and relauch lobby and game - you will see 40 in game settings.

If your `springsettings.cfg` already have `ScrollWheelSpeed`, it won't be overwrite.

Additional is something what I saw during tests. in-game settings panel doesn't refrect all changes from springsettings.cfg.
If I'm not wrong, graphics settings is separate Lua widget that store own state. e.g when I changed key shadows, it was changing in `springsettings.cfg` but in game settings I saw Shadows = Medium all the time. I think we need to separate task for research and fix that.

Hectate:
> Additional is something what I saw during tests. in-game settings panel doesn't refrect all changes from springsettings.cfg. If I'm not wrong, graphics settings is separate Lua widget that store own state. e.g when I changed key shadows, it was changing in `springsettings.cfg` but in game settings I saw Shadows = Medium all the time. I think we need to separate task for research and fix that.

That could be why I didn't find that my changes took effect. I definitely agree that we would need to look into this further if that is the case. I'll do some testing now as you indicated. Thanks.

Hectate (approved):

PR #584, opened 2026-05-25
Override extension for replay files

### LINKED ISSUE #538: Set up file association for replay files
Right now we do not automatically associate the replay file extension with lobby. People have to do it manually themselves.

The association should work reliably across platforms (Windows and Linux), which includes when game is installed - currently the only option - as the per-user, not per-machine install on Windows (we might not want to ever support per-machine installs as outlined in #395). [I hope that won't require manually touching registry.](https://learn.microsoft.com/en-us/windows/win32/shell/app-registration#using-the-applications-subkey)

Associating `sdfz` is also a bit problematic because all Recoil games share that extension. *Maybe* that's not that such a huge issue: people can always change association manually, but it's worth considering. Opened https://github.com/beyond-all-reason/RecoilEngine/issues/2784 in engine for alternate extension support.

BElluu:
Closed #538 

Waiting for changes from:
1. #578 
2. [RecoilEngine #2975](https://github.com/beyond-all-reason/RecoilEngine/pull/2975)
Changes in engine merged


Hectate:
[Per the electron-builder documentation,](https://www.electron.build/docs/api/app-builder-lib.interface.fileassociation/) we need `perMachine` to be `true` to associate files on Windows. I thought we had that already, though, which is why that seems odd to me. I could be wrong.

I presume I can just rename a normal replay to have the new extension to test? That is to say; internally there is no difference in the replays, it's only the extension that has changed?

BElluu:
Yea that's only extension name. You can change extension for any replay and test it.

Hectate (approved):

Hectate:
This seems good. The replays won't start using the new extension until the next engine release, I presume, but that's not going to stop the existing ones from working and it'll automatically pick up the new replays once they start being made. Thanks.

PR #589, opened 2026-06-05
Tachyon party support added

### LINKED ISSUE #592: Add party endpoint for Tachyon requests and events

Hectate:
Creates a backend ``party.store.ts`` to handle party-related Tachyon requests and events. None of this is exposed in the UI at this time.

### AI / LLM usage statement:
No AI was used in any fashion during the creation of this PR.


Hectate:
Screenshot of console history of two clients showing that party actions are functional between each. Also tested were ``party/declineInvite``, ``party/cancelInvite``, and ``party/kickMember``.
image

src/renderer/store/party.store.ts
```diff
@@ -0,0 +1,217 @@
...
+import { notificationsApi } from "@renderer/api/notifications";
+import { Party } from "@renderer/model/party";
+import { subsManager } from "@renderer/store/users.store";
+
+const partySymbol = Symbol("party.store");
+
+export enum PlayersPartyState {
+    None = "None",
+    InvitedOnly = "InvitedOnly",
+    JoinedOnly = "Joined",
```
BElluu: That should not be `JoinedOnly = "JoinedOnly"` ?

src/renderer/store/party.store.ts
```diff
@@ -0,0 +1,217 @@
...
+    partyStore.parties.set(data.party.id, { ...data.party, seen: false });
+    parsePartyData();
+}
+
+function onRemovedEvent(data: PartyRemovedEventData) {
+    console.log("Tachyon: party/removed:", data);
+    // Note that "party/removed" includes cancelled or expired invitations in addition to being kicked/leaving.
+    partyStore.parties.delete(data.partyId);
+    parsePartyData();
+    if (partyStore.activeParty?.id == data.partyId) {
```
BElluu: use `===` instead `==`

src/renderer/store/party.store.ts
```diff
@@ -0,0 +1,217 @@
...
+        partyStore.activeParty = null;
+    }
+}
+
+function onUpdatedEvent(data: PartyUpdatedEventData) {
+    console.log("Tachyon: party/updated:", data);
```
BElluu, line 164: `console.log` is ok for development, but not for production. Maybe change all calls to `console.debug` and `console.error` ?

src/renderer/store/party.store.ts
```diff
@@ -0,0 +1,217 @@
...
+    const bools = {
+        joined: false,
+        invited: false,
+    };
+    const users: Array<UserId> = [];
+    partyStore.parties.forEach((party) => {
+        for (const member of party.members) {
+            if (member.userId === me.userId) {
+                bools.joined = true;
+                partyStore.activeParty = party;
```
BElluu, line 182: I'm not sure, but you should not reset `partyStore.activeParty` on the beginning of fuction?

src/renderer/store/party.store.ts
Hectate: If you look across the various stores that use Tachyon requests/responses/events you'll see this pattern of `console.log()` for the data. This is because we're frequently using the client to test and validate the protocol/server, so we pretty much always want to see the data at this stage.

All these will need touching regardless later. Not just to remove extraneous logging, but also to implement i18n into notifications, etc.

src/renderer/store/party.store.ts
Hectate: No. This is due to the way the Tachyon protocol provides party updates to both members and invitees. Receiving a party update is not a guarantee that you are getting an update about the user's active party - they could have been invited to a new party.

As a result, we throw the data into `parsePartyData()` and if the user's own ID is in the `members` property of said party, we update the `partyStore.activeParty` property to be that party (line 188).

Hectate: Disregard my prior comment. You are correct about the reset. I was thinking it through and I realized that while we have the `requestLeave` and `onRemovedEvent`, we cannot guarantee ordering of updates either. I've fixed that now.

Hectate:
Holding this until https://github.com/beyond-all-reason/teiserver/pull/1274 is merged or rejected so that I know which way to handle ``party/create``.

BElluu:
> Holding this until [beyond-all-reason/teiserver#1274](https://github.com/beyond-all-reason/teiserver/pull/1274) is merged or rejected so that I know which way to handle `party/create`.

Merged

Hectate:
> > Holding this until [beyond-all-reason/teiserver#1274](https://github.com/beyond-all-reason/teiserver/pull/1274) is merged or rejected so that I know which way to handle `party/create`.
> 
> Merged

Yep! Just pushed the change and was waiting for the checks to pass. All done now.

PR #590, opened 2026-06-07
Matchmaking assets downloading

### LINKED ISSUE #526: Ensure assets are ready before queuing for matchmaking
When doing a `matchmaking/list` [request](https://github.com/beyond-all-reason/tachyon/blob/master/docs/schema/matchmaking.md#list) each queue has a list of engine version, game version, and maps.

The client should ensure all these assets are downloaded and ready before actually joining the queue.

This is to avoid the client receiving a `battle/start` request to join a match with missing assets, leading to runtime errors at startup.

Hectate:
Matchmaking now uses DownloadContentButton to acquire required assets prior to allowing the user to join a queue.
Added a modal to display the required assets for the selected queue, along with other information about the queue.

It is acceptable to navigate to other areas of the client and use them while downloads are in progress.
Note that until `pr-downloader` handles concurrent downloads, navigating to _other downloads_ and triggering them concurrently in the client may result in individual download failures. If this happens, the failure is graceful and recognizes that the asset is missing. A fresh download of the requested asset can be triggered to resolve this afterward.

### Video Demonstration (no audio)
https://youtu.be/e7-l0sqd9BM

### AI / LLM usage statement:
No AI was used in the creation of this PR.


src/renderer/store/matchmaking.store.ts
```diff
@@ -120,28 +122,59 @@ async function sendListRequest() {
...
 }
 
-async function checkIfAnyMapsAreNeeded(maps: { springName: string }[]): Promise<boolean> {
-    if (maps.length == 0) return false;
+/**
+ * Refreshes the downloadsRequired arrays upon demand, if we expect that things have changed.
+ */
+function triggerAssetsRefresh() {
+    for (const queue of matchmakingStore.playlists) {
+        setRequiredAssetsArrays(queue.id, queue.engines, queue.games, queue.maps);
```
BElluu: Should `await`

src/renderer/store/matchmaking.store.ts
```diff
@@ -120,28 +122,59 @@ async function sendListRequest() {
...
-    return false;
+    const queueEngines = new Set(engines.map((e) => e.version));
+    const installedEngines = new Set(enginesStore.availableEngineVersions.filter((e) => e.installed).map((e) => e.id));
+    const diffEngines = queueEngines.difference(installedEngines);
+    matchmakingStore.downloadsRequired[queue].engines.push(...diffEngines);
+    const queueGames = new Set(games.map((g) => g.springName));
+    const installedGames = new Set(gameStore.availableGameVersions.keys());
+    const diffGames = queueGames.difference(installedGames);
+    matchmakingStore.downloadsRequired[queue].games.push(...diffGames);
+    return;
```
BElluu: unnecessary

src/renderer/views/play/matchmaking.vue
```diff
@@ -82,24 +96,72 @@ SPDX-License-Identifier: MIT
...
+
+const anyDownloadsAreRequired = computed(() => {
+    if (
+        matchmakingStore.downloadsRequired[matchmakingStore.selectedQueue].maps.length > 0 ||
+        matchmakingStore.downloadsRequired[matchmakingStore.selectedQueue].engines.length > 0 ||
+        matchmakingStore.downloadsRequired[matchmakingStore.selectedQueue].games.length > 0
+    )
+        return true;
+    else return false;
+});
```
BElluu: maybe like that?

``` 
const anyDownloadsAreRequired = computed(() => {
         const d = downloadsRequired.value;
         return d.maps.lengiht > 0  || d.engines.length > 0 || d.games.lenght > 0;
});
```

Hectate: `matchmakingStore.downloadsRequired`` has a key that uses the name of the queue to distinguish which downloads are required for which queue (game, engine, map list can all be unique between queues).

This computed property `anyDownloadsAreRequired`` changes based on the queue that is currently selected to modify the visibility of the DownloadContentButton, since a user may have all the assets for one queue but not another. As such, we can't simply check if _any_ queue needs downloads, or else we would prevent users from joining queues where they have all the assets simply because another queue does not.

I will, however, rename the computed property to `areDownloadsRequiredForSelected` be more clear as to what it is presenting, as that seems to have influenced your suggestion.

PR #594, opened 2026-06-12
Implemented Taskbar/Dock progress bar

### LINKED ISSUE #131: Make app icon show download progress

Hectate:
Closes #131 

Electron permits updating a "progress bar" attached to the main icon in the taskbar or dock. This implements the use of that feature for all downloads.

Tested and working on Windows 11. I don't have immediate access to a Linux machine that can run this so some confirmation there would be appreciated. Note comments in the [Electron documentation](https://www.electronjs.org/docs/latest/api/browser-window#winsetprogressbarprogress-options).

> On Linux platform, only supports Unity desktop environment, you need to specify the *.desktop file name to desktopName field in package.json. By default, it will assume {app.name}.desktop.

### AI / LLM usage statement:
`qwen2.5-coder:1.5b-base` was used as an Autocomplete/Inline Suggestions aid. All suggestions were reviewed and approved individually.

### DIFF
--- a/src/main/services/downloads.service.ts
+++ b/src/main/services/downloads.service.ts
@@ -3,10 +3,13 @@
 // SPDX-License-Identifier: MIT
 
 import { autoUpdaterAPI } from "@main/content/auto-updater";
+import { DownloadInfo } from "@main/content/downloads";
 import { engineContentAPI } from "@main/content/engine/engine-content";
 import { gameContentAPI } from "@main/content/game/game-content";
 import { mapContentAPI } from "@main/content/maps/map-content";
 import { BarIpcWebContents } from "@main/typed-ipc";
+
+const downloads: Map<string, DownloadInfo> = new Map();
 
 function registerIpcHandlers(webContents: BarIpcWebContents) {
     engineContentAPI.onDownloadStart.add((downloadInfo) => {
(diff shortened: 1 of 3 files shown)


src/main/services/downloads.service.ts
```diff
@@ -59,8 +62,96 @@ function registerIpcHandlers(webContents: BarIpcWebContents) {
     });
 }
 
+function registerProgressHandler(mainWindow: Electron.CrossProcessExports.BrowserWindow) {
+    engineContentAPI.onDownloadStart.add((downloadInfo) => {
```
BElluu: I think w should little refactor this registration, because we have around 70 lines of code for registering. Maybe like that? 

```TypeScript
const downloads: Map<string, DownloadInfo> = new Map()
const allDownloaders: Downloader[] = [engineContentAPI, gameContentAPI, mapContentAPI, autoUpdaterAPI];

function registerProgressHandler(mainWindow:Electron.BrowserWindow) {
  for (const api of allDownloaders) {
       api.onDownloadStart.add((d) => { downloads.set(d.name, d); updateProgressBar(mainWindow); });
       api.onDownloadComplete.add((d) => { downloads.set(d.name, d); updateProgressBar(mainWindow); });
       api.onDownloadProgress.add((d) => { downloads.set(d.name, d); updateProgressBar(mainWindow); });
       api.onDownloadFail.add((d) => { downloads.set(d.name, d); updateProgressBar(mainWindow); });
   }
}
```

src/main/services/downloads.service.ts
```diff
@@ -3,11 +3,14 @@
...
 import { autoUpdaterAPI } from "@main/content/auto-updater";
+import { DownloadInfo } from "@main/content/downloads";
 import { engineContentAPI } from "@main/content/engine/engine-content";
 import { gameContentAPI } from "@main/content/game/game-content";
 import { mapContentAPI } from "@main/content/maps/map-content";
 import { BarIpcWebContents } from "@main/typed-ipc";
 
+const downloads: Map<string, DownloadInfo> = new Map();
+
 function registerIpcHandlers(webContents: BarIpcWebContents) {
```
BElluu, line 12: Same as below. Can be refactor too :)

Azzurite:
> On Linux platform, only supports Unity desktop environment

This basically means no one uses it :D you have to specifically install the `Ubuntu Unity` Linux distribution for that, which probably almost no one has. The default Ubuntu window manager is `GNOME`, not `Unity`. So I don't think it's particularly important to test that, since also it's just testing Electron behavior, which Electron devs should be testing, not us.

src/main/services/downloads.service.ts
Hectate: Unlike the other, this one is a bit tricky because we need to maintain type-safety with the name of the events in `BarIpcWebContents`. We can't make the strings dynamic (e.g. ``webContents.send(`downloads:${api}:${event}`)`` because there's no way to guarantee that the interpolated string will match a valid event string.

I experimented with an object to map the API to the commands and then loop, but then you have separate string literals and the problem repeats.

In short, refactoring this section doesn't seem to gain us much, but could introduce points of failure instead.

PR #597, opened 2026-06-15
Config Service Added

### LINKED ISSUE #291: Dynamic configuration of lobby

### LINKED ISSUE #293: Add minimal Github Actions workflow to push new version of configuration
Push part will depend on the API to be exposed by https://github.com/beyond-all-reason/bar-lobby/issues/292

### LINKED ISSUE #596: Create a file for client configuration
We now have a Github action to push new configurations.
The client will need to regularly fetch the config file and use it. Since this file does not exist, we have to define it first.

The steps to complete this are:

- [x] Determine which variables, settings, etc. are to be included in this config file
- [x] Create a schema to validate configurations against.
- [x] Create a default, local configuration file for the client to load and apply
- [x] Add a process to check for updated config files and fetch it
- [x] Replace the local config file if appropriate.
- [x] Refresh configuration with the new file.
- [X] Create a generate script that writes a config file for upload use
- [x] Allow command line arguments to override the configuration file

Hectate:
Closes #596 

# Client
As per the parent issue, this will allow us to reconfigure some values without requiring an entire release of the client. Upon initialization, the client will attempt to handle 4 possible sources of config information. They will be applied in the following order, with each one having the ability to overwrite values from any prior:
1. Local file `state/config/config.json` values.
2. Fetch `https://lobby-config.beyondallreason.dev/config.json`.
3. A local config file at a path provided as a flag, e.g. `--config="path/to/config.json"`

Absent anything else, the default value in the schema will be used.

For use in `main` there is `configService`, and for `renderer` there is `configStore`, which is read-only and utilizes the former as the source of data. At runtime, config values are written to `state\config\config.json` and accessed as noted in (1) above.

A complete config file can be generated using ``npm run generate-default-config``, which will read the Schema and assign the default values into matching JSON output.

# CI
PR also updates the checks (local and CI) to validate that the `remoteConfig/config.json` is valid JSON that matches the config update Schema.

Upon a push to `master`, the contents located in `remoteConfig/config.json` will be sent to the config serving endpoint for use by live clients as per (2) above.

### AI / LLM usage statement:
Local `qwen2.5-coder` and later `qwen3-coder`, as well as Co-Pilot were used as an Autocomplete/Inline Suggestions tool. All suggestions were individually approved by me.


Hectate:
@p2004a I believe this is ready for some examination.

I am unfamiliar with setting up Github Actions, so if you want it to auto-generate a fresh `config.json` feel free to hook up the script or add your own commits to the PR for that. If you just want to point it to the `config.json` file, then that works too and we can manually generate one and merge the commit if needed.

src/renderer/store/config.store.ts
```diff
@@ -0,0 +1,29 @@
+// SPDX-FileCopyrightText: 2026 The BAR Lobby Authors
+//
+// SPDX-License-Identifier: MIT
+
+import { Config } from "@main/services/config.service";
```
p2004a: nit: `import type`

auxiliary/config-serving/json/config.json
p2004a, line 1: Why check in into repo and why in this location?

src/main/content/game/game-content.ts
```diff
@@ -156,7 +156,7 @@ export class GameContentAPI extends PrDownloaderAPI<string, GameVersion> {
      * Downloads the actual game files, will update to latest if no specific gameVersion is specified
      * @param gameVersion e.g. "Beyond All Reason test-16289-b154c3d"
      */
-    public async downloadGame(gameVersion = `${contentSources.rapid.game}:test`) {
+    public async downloadGame(gameVersion = `${configService.getConfig().rapidGame}:test`) {
```
p2004a: At this point I guess dropping the default entirely and making sure all caller picks correct one makes more sense.

src/main/json/model/config.ts
```diff
@@ -0,0 +1,65 @@
...
+//
+// SPDX-License-Identifier: MIT
+
+// Changes to this config schema's defaults will require a new config version to be deployed.
+// You can generate a new one from defaults with the command: npm run generate-default-config
+
+import { Type } from "@sinclair/typebox";
+
+export const configSchema = Type.Object({
+    // Note, configUrl must match the expected location or else Github Actions will not upload a new one.
```
p2004a: The whole BAR_SPRINGSETTINGS_DEFAULTS should be in config too IMHO.

src/main/json/model/config.ts
```diff
@@ -0,0 +1,65 @@
...
+    configUrl: Type.String({ default: "https://lobby-config.beyondallreason.dev/config.json" }),
+    rapidHost: Type.String({ default: "repos-cdn.beyondallreason.dev" }),
+    rapidGame: Type.String({ default: "byar" }),
+    gameGithubOwner: Type.String({ default: "beyond-all-reason" }),
```
p2004a: I don't see gameGithubOwner to be used anywhere. Let's drop this and all other properties that are unused.

src/main/json/model/config.ts
```diff
@@ -0,0 +1,65 @@
...
+    gameGithubRepo: Type.String({ default: "Beyond-All-Reason" }),
+    engineGithubOwner: Type.String({ default: "beyond-all-reason" }),
+    engineGithubRepo: Type.String({ default: "spring" }),
+    engineReleaseUrl: Type.String({ default: "https://files-cdn.beyondallreason.dev/find" }),
+    defaultMaps: Type.Array(Type.String(), { default: ["Quicksilver Remake 1.24", "Red Comet Remake 1.8"] }),
+    latestGameVersion: Type.String({ default: "byar:test" }),
+    defaultEngineVersion: Type.String({ default: "2025.06.21" }),
+    cdnUrl: Type.String({ default: "https://pool-init.beyondallreason.dev" }),
```
p2004a: too generic name, use e.g. `initialPoolDataUrl`

src/main/json/model/config.ts
```diff
@@ -0,0 +1,65 @@
...
+    prdRapidUseStreamer: Type.String({ default: "false" }),
+    prdRapidRepoMaster: Type.String({ default: "https://repos-cdn.beyondallreason.dev/repos.gz" }),
+    prdHttpSearchUrl: Type.String({ default: "https://files-cdn.beyondallreason.dev/find" }),
+    logUploadUrl: Type.String({ default: "https://log.beyondallreason.dev/" }),
+    lobbyServer: Type.String({ default: "wss://server4.beyondallreason.info" }),
```
p2004a: why we have both lobby server and default lobby servers?

src/main/json/model/config.ts
```diff
@@ -0,0 +1,65 @@
...
+    mapsMetadataUrl: Type.String({ default: "https://maps-metadata.beyondallreason.dev/latest/lobby_maps.validated.json" }),
+    newsRssUrl: Type.String({ default: "https://www.beyondallreason.info/news/rss.xml" }),
+    devlogRssUrl: Type.String({ default: "https://www.beyondallreason.info/microblogs/rss.xml" }),
+    replayServiceUrl: Type.String({ default: "https://bar-rts.com/replays" }),
+    newsServiceUrl: Type.String({ default: "https://www.beyondallreason.info/news" }),
```
p2004a: aThis is used only in whitelist, it doesn't have semantic meaning. Instead, it should be e.g. array `allowedUrlLinks` or something like that.

src/main/json/model/config.ts
```diff
@@ -0,0 +1,65 @@
...
+    defaultServers: Type.Array(Type.String(), {
+        default: ["wss://server4.beyondallreason.info", "wss://server5.beyondallreason.info", "wss://lobby-server-dev.beyondallreason.dev", "ws://localhost:4000"],
+    }),
+    replayDownloadUrl: Type.String({ default: "https://bar-rts.com/replays" }),
+});
+
+export const envVarsMap = {
```
p2004a: This is _very_ error-prone, even right now:
- `mapDownloadUrl` below doesn't exist in the object
- `defaultServers` is in schema, doesn't exist here

I'm not sure if instead of doing this with env vars, it's not easier to take a `--config` flag that just takes a path to a full config used instead of the main one. The env vars don't work well for any array variables etc.

Is there some important use cases I'm missing that need env vars?

src/main/services/config.service.ts
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Fetch the latest configuration from the remote URL and update the local config store.
+ * Note that env vars will be used to override the config values if they are set, including remote config values.
+ */
+async function fetchConfig() {
+    try {
+        const response = await fetch(getConfig().configUrl);
+        const data = await response.json();
+        if (!Value.Check(configSchema, data)) {
+            throw Error("Fetched config does not match schema");
```
p2004a: Let's print more error details via [`Value.Errors`](https://sinclairzx81.github.io/typebox/#/docs/value/errors)

src/main/main.ts
```diff
@@ -131,6 +132,8 @@ app.whenReady().then(async () => {
         });
     });
     // Initialize services
+    // Config is fetched first because it contains URLs and other values that other services may depend on.
+    await configService.init();
```
p2004a, line 129: What is behavior when the config fetching is slow? I believe application will look like it hang.

src/main/services/config.service.ts
```diff
@@ -0,0 +1,79 @@
...
+    return await configStore.update(data);
+}
+
+/**
+ * Fetch the latest configuration from the remote URL and update the local config store.
+ * Note that env vars will be used to override the config values if they are set, including remote config values.
+ */
+async function fetchConfig() {
+    try {
+        const response = await fetch(getConfig().configUrl);
```
p2004a, line 63: you don't check response status

auxiliary/config-serving/json/config.json
Hectate: Location was arbitrary, but it was adjacent to the config-serving files which seemed appropriate. We can change to whatever you prefer.

As far as committing the file itself, I was unsure if you preferred the "generate locally and commit after making changes that affect it" or if you just would trigger the script in the workflow and grab the file from there on every run. We have the option to go either route you feel is suitable. I do prefer the automated route, although the vast majority of commits will likely not touch the config.

src/main/json/model/config.ts
Hectate: I debated that myself. The decision to exclude them came down to two reasons:
1. It would add a significant number of entries to the config
2. Valid/default settings could vary depending on the engine version

The second in particular is why I think the default engine settings should be included with either the game or engine files, rather than the client configuration.

src/main/json/model/config.ts
Hectate: The `lobbyServer` is the currently selected lobby server. It can be overwritten with any other string desired but we have a sensible default provided. Changing it only happens with dev-mode enabled. Obviously important for the config because it allows us to redirect clients automatically to a new lobby server address at client-launch.

The `defaultServers` list is the pre-known lobby servers (which can be selected from or added to via the settings' schema `customServerList`), which are for the convenience of developers. We have the default `server4...` included so that it appears in the list as a choice regardless of whatever else may be the current server in `lobbyServer`. Less important for the config, but including it can allow us to modify the default list at-will as well (in case we add a `server6...` etc.).

src/main/json/model/config.ts
Hectate: I made env vars for each because #291 indicated "Ability to override config via command line flag" as a requirement, and it was the first approach that came to mind. Your suggestion of reading a file is fine too. We could treat it as "all values optional" so the user only has to provide the ones they wish to override.

Note that the `./state/config/config.json` file itself can be directly edited after the first-run as well to change these when the client loads the local config.

Some values may be mildly convenient for development as env vars, but I'm not worried if we don't take that approach for any of them.

src/main/main.ts
Hectate: Good call, I'll take a look for that specifically. I knew we needed more testing anyway because currently the fetched config has always been invalid anyway.

Unfortunately because so many of these values are required for future initializations, we do need to have something to stop the client from proceeding. Not to say that we can't try to have something that says "fetching config" or similar while that happens, if the renderer's window has been created yet.

src/main/content/game/game-content.ts
Hectate: Fair point. Also, with all the updates we've been able to do elsewhere (e.g. `DownloadContentButton` enables downloads for game/engine/map assets combined) I feel like this probably isn't needed often. Maybe only during initial setup. I'll look.

src/main/json/model/config.ts
Hectate: I noticed as well but was unsure for their purpose. I assume it was something that has been supplanted in the past. I'll double-check all of them.

src/main/services/config.service.ts
Hectate: Right you are. Fixed and tested w/ ``"configUrl": "https://www.google.com/config.json",``

```
[22:22:22.294] ERROR: main - 2026-06-28T02:22:21.983Z - config.service.ts - Error fetching config: Failed to fetch config: 404 Not Found
```

src/main/main.ts
Hectate: OK, so because of the placement of this, the main window is not yet created. Adding a manual delay causes the window to not appear since the initializations have not completed.

I tested moving the window creation before initializations (without any other changes) and it just failed to load properly because the IPC handlers were not able to hook up to the services yet since they were not ready.

So I think perhaps the "proper" way to do this is to create a mini-window "splash screen" specifically for "Getting stuff ready..." that does no setup work, but just exists to make sure that the user knows the client is doing things. Once we get to the point where the main window is going to get created, we can kill that first window. This is common among Electron applications too - Discord uses a setup like this.

src/main/json/model/config.ts
p2004a: You shouldn't encode all possible engine settings in schema, engine settings are just `Record<string,string|number>`.

Including in engine is not a good idea because this is not BAR's engine. Engine doesn't know what are settings for BAR game.

Including them in game means that we need a new game release to update them, I'm not sure it's desirable in the future.

In the future we might want to do more then just "every user has the same exact config".

src/main/json/model/config.ts
p2004a: The override wouldn't be stored in config but in settings, so the `lobbyServer` here is effectively just "main default server", we could just use first one from the list for that purpose.

src/main/json/model/config.ts
Hectate: You're right, I wasn't considering that I didn't have to strictly define them per-entry but can just allow an undefined set of them to be included.

I do think that the game files should come with defaults, however, but that's a different topic. Even if they were provided, there's nothing to say we might not choose to override them this way still.

Also, when stating engine/game it was just my shorthand for "sourced from either as appropriate" - certainly game is the right place for probably everything we are looking at.

src/main/json/model/config.ts
Hectate: Fair enough. I'll move it back to settings and add some code to pull the first entry when setting up the `settings.service` initially.

src/main/json/model/config.ts
p2004a: Yeah, by "override config via command line flag" I had command line flag that we can point at config just like we have in spring launcher in mind 😛.

Modifying config can be acceptable too, that's true.

For env vars though, this isn't good implementation anyway because it's error prone: it needs to be automatically computed from schema properties and again it will be annoying to deal with nested properties objects, arrays etc.

Hectate: Considering that this is largely dev-focused for someone to want to change it anyway, worst case we can just pull a clean `config.json` from the service and tweak it if it has to be done first-launch. I'll switch it around.

src/main/main.ts
p2004a: Sooo, like a launcher for client that doesn't have launcher that does this single thing and no other updates.

In spring launcher we do it differently: we reload initialization pieces that are impacted by config change when the config changed.

Spring-launcher is overall optimized for speed more then new lobby, some steps are done concurrently to minimize time user has to wait.

Hectate: Not even a launcher really (not at this stage anyway), just a window with some life to users know something is happening.

It’s in the PR now if you want to see it in this form.

p2004a:
By the way, why the current empty config is invalid?

Hectate:
> By the way, why the current empty config is invalid?

None of the items in the schema are optional, so their absence is causing errors.

p2004a:
This sounds off given that we have defaults in schema.

In the future:
- Schema will need to be part of release 
- In CI we will verify that new config version works correctly with old releases
- Because we will need it to work with old releases, we might have to keep some fields set that do not exist in the schema anymore
- For fields that existed in the schema, but the defaults are correct, we do not need to keep them in config when we drop the property from schema

src/main/json/model/config.ts
Hectate: Got it working with a config path. I found out that if using it in the start script you have to pass the flag with a pair of double dashes so electron-forge passes the config along to electron itself. For example:

```
npm run start -- -- --config="D:\git\bar-lobby\auxiliary\config-serving\json\config.json"
```

src/main/content/game/game-content.ts
Hectate: Cleaned up the `game-content.ts` file a little bit.

I was also able to determine that the default is never needed because of prior checks, etc. Instead, we _are_ already pulling in the default version when doing setup, etc. So yeah, this is taken care of already - even during initial setup.

Hectate:
OK, I think this is actually ready now. Given the sensitive nature of this, I think the next step before merge is going to be a test. That said, I believe it's working correct from my manual tests (see https://github.com/Hectate/bar-lobby/actions/runs/29551022116/job/87793331862 for an example dry run).

Hectate:
OK @p2004a I think this is got everything covered that you requested and has tests on the config generation script.

Hectate:
@p2004a I'm going to pull the SplashWindow stuff out, but it does no work - appears early in the launch, is auto-closed once the main window is spawned - so its a zero impact change other than readjusting tests. I wanted you to know in case you happened to look at this before I had made those changes.

During my testing of the 0.16.0 release I realized that the only reason I had been seeing the SplashWindow was because the dev startup is slower (presumably the live reloading server startup/hookup). I did a handful of testing and could not get the client to hang with the config fetching; I'm assuming it's because the file is so small that the fetch is fast. Even when fully offline it doesn't hold us up because the fetch fails immediately. Possibly if the config file grows large it could be an issue, but since we have auto-updates in the client that will provide updated defaults _that way_ also, we can likely keep the config small constantly.

.github/workflows/config-serving-deploy.yml
```diff
@@ -16,18 +22,47 @@ jobs:
...
+            IS_DRY_RUN: ${{ github.event.inputs.dry_run == 'true' }}
         steps:
-            - name: Push config
+            - name: Checkout Git repository
+              uses: actions/checkout@v6
+
+            - name: Setup node
+              uses: actions/setup-node@v6
+              with:
+                  node-version: 22.18.0
```
p2004a: Do this instead so we don't hardcode node version everywhere https://github.com/beyond-all-reason/bar-lobby/blob/ae829e034848938099c35794facc5579c21c1e72/.github/workflows/build.yml#L32

.github/workflows/config-serving-test.yml
```diff
@@ -24,7 +24,7 @@ jobs:
             - name: Setup node
               uses: actions/setup-node@v6
               with:
-                  node-version: 20
+                  node-version: 22.18.0
```
p2004a: Not really, `auxiliary/config-serving` is it's own application. There is no reason why it's node version should be the same with main application.

It is a mistake I've not provided correct version in that directory, but that's unrelated change to this pr.

tools/generate-diff-config.ts
```diff
@@ -0,0 +1,73 @@
+// SPDX-FileCopyrightText: 2026 The BAR Lobby Authors
```
p2004a: I'm against this approach. This means that config we store on server side is not versioned by anything, it's ephemeral. The version stored on server does not exist anywhere in source, it's not possible to do any rollback etc.

I assume you've done this because of my comment

> Because we will need it to work with old releases, we might have to keep some fields set that do not exist in the schema anymore

But it was for "in the future" and if you wanted to actually do it right now, there is no other way then to check in the config into the source and validate that against the schema. That also means dropping `tools/generate-default-config.ts` and so on.

p2004a (commented):
> I'm going to pull the SplashWindow stuff out, but it does no work

I don't follow, you want to remove it from this PR because config fetching for you is quick? It can take multiple seconds with retries etc, you just happen to have a good connection.

We don't have to do it in this PR though, it can be done later. I believe there is a better and cleaner way to do it, that doesn't involve the splash window, we just happened to have a very messy initialization order.

Hectate:
> I don't follow, you want to remove it from this PR because config fetching for you is quick? It can take multiple seconds with retries etc, you just happen to have a good connection.
> 
> We don't have to do it in this PR though, it can be done later. I believe there is a better and cleaner way to do it, that doesn't involve the splash window, we just happened to have a very messy initialization order.

What I simply mean is that it added no value so we can instead, as you noted, address the initialization issue later with a better approach.

tools/generate-diff-config.ts
Hectate: So if I understand you, what you want is for the config to be a committed file that gets uploaded as-is after validation. It'll be up to the reviewer to make sure they're satisfied with the config contents before pushing it to `master`, assuming it passes schema tests. Is that accurate?

If so, that means we really just need:
1. A validation script
2. A consistent place to store it
3. Modifications to your action script that takes, validates, and uses it

p2004a: Yes

Hectate:
Updated:
* SplashWindow removed.
* `/remoteConfig/config.json` committed
* `/tools/validate-config.ts` added to `npm run checks` and CI tests
* CI Config Deploy updated to use the committed file

src/main/json/model/config.ts
```diff
@@ -0,0 +1,130 @@
...
+    mapsMetadataUrl: Type.String({ default: "https://maps-metadata.beyondallreason.dev/latest/lobby_maps.validated.json" }),
+    newsRssUrl: Type.String({ default: "https://www.beyondallreason.info/news/rss.xml" }),
+    devlogRssUrl: Type.String({ default: "https://www.beyondallreason.info/microblogs/rss.xml" }),
+    allowedUrlLinks: Type.Array(Type.String(), { default: ["https://bar-rts.com/replays", "https://www.beyondallreason.info/news"] }),
+    replayServiceUrl: Type.String({ default: "https://bar-rts.com/replays" }),
+    onlineReplaysApiUrl: Type.String({ default: "https://api.bar-rts.com/replays" }),
+    defaultServers: Type.Array(Type.String(), {
+        default: ["wss://server4.beyondallreason.info", "wss://server5.beyondallreason.info", "wss://lobby-server-dev.beyondallreason.dev", "ws://localhost:4000"],
+    }),
+    replayDownloadUrl: Type.String({ default: "https://bar-rts.com/replays" }),
```
p2004a: Dead property, drop please.

src/main/game/springsettings.ts
```diff
@@ -7,102 +7,13 @@ import * as os from "os";
...
-    UseNetMessageSmoothingBuffer: 0,
-    NetworkLossFactor: 2,
-    MaxDynamicModelLights: 0,
-    LinkOutgoingBandwidth: 262144,
-    LinkIncomingSustainedBandwidth: 1048576,
-    LinkIncomingPeakBandwidth: 1048576,
-    LinkIncomingMaxPacketRate: 2048,
-    DemoFileExtension: "barreplay",
-};
+const BAR_SPRINGSETTINGS_DEFAULTS = configService.getConfig().springSettings;
```
p2004a: This gets evaluated only at import time, before config service is actually initialized. Instead use `configService.getConfig().springSettings` directly below.

src/main/content/replays/online-replays.ts
```diff
@@ -4,10 +4,11 @@
 
 import { IpcResult } from "@main/typed-ipc";
 import { logger } from "@main/utils/logger";
+import { configService } from "@main/services/config.service";
 
 const log = logger("online-replays.ts");
 
-export const ONLINE_REPLAYS_API_URL = "https://api.bar-rts.com/replays";
+export const ONLINE_REPLAYS_API_URL = configService.getConfig().onlineReplaysApiUrl;
```
p2004a: Similar to engine, this is read and assigned as constant during module initiliazation, before config service can be ready. Instead use `configService.getConfig().onlineReplaysApiUrl` below directly.

src/renderer/views/debug/script-launcher.vue
```diff
@@ -39,10 +38,10 @@ const script = ref(`[game] {
     mapname=Red Comet Remake 1.8;
```
p2004a: nit: use first of default maps here?

src/main/services/config.service.ts
```diff
@@ -0,0 +1,97 @@
...
+
+const configStore = new FileStore<typeof configSchema>(path.join(CONFIG_PATH, "config.json"), configSchema);
+
+async function init() {
+    await configStore.init();
+    await fetchConfig();
+    await checkForConfigOverride();
+}
+
+async function checkForConfigOverride() {
```
p2004a, line 25:
```ts
import { parseArgs } from "node:util";

async function checkForConfigOverride() {
    const parsedArgs = parseArgs({
        args: process.argv.slice(1),
        options: { config: { type: "string" } },
        strict: false,
    });
    if (parsedArgs.values.config) {
        const configPath = path.resolve(process.cwd(), parsedArgs.values.config.toString());
```

src/main/services/config.service.ts
```diff
@@ -0,0 +1,97 @@
+// SPDX-FileCopyrightText: 2026 The BAR Lobby Authors
```
p2004a, line 1: Across file, and overall, better `throw new Error` instead of `throw Error`.

If you used custom error class, `throw MyError` would not work.

src/renderer/store/config.store.ts
```diff
@@ -0,0 +1,29 @@
...
+    if (configStore.isInitialized) {
+        console.warn("Config store is already initialized, skipping initialization.");
+        return;
+    }
+    const currentConfig = await window.config.getConfig();
+    Object.assign(configStore, currentConfig);
+    configStore.isInitialized = true;
+    watch(
+        configStore,
+        () => {
```
p2004a: I don't believe I understand what is the scenario in which we would ever want to update config directly from renderer, so this function, and all the plumbing via IPC? Config is static.

p2004a (commented):
Additional places where we have still hardcoded properties:

- ReportUserModal.vue:177 `const REPLAY_SERVICE_URL = "https://bar-rts.com/replays";`
- game-content.ts:150: `if (version === "byar:test") {`

src/renderer/store/config.store.ts
Hectate: I don't anticipate using it in a release, but thought that it could be helpful in development by letting us change something over without having to relaunch the client just to pull in new config settings. I agree that config should be relatively static.
I'll remove it so it's less code to maintain. If we end up needing it, we'll put it back in.

PR #598, opened 2026-06-15
Move maps online work to download stage

### LINKED ISSUE #583: InitialSetup UI component should not do online work
Source: https://github.com/beyond-all-reason/bar-lobby/pull/572#issuecomment-4359255855
> ... ``Preloader`` should be 100% offline and not do any online work. The fetching of maps metadata should be part of the initial setup step that has all the other updates.

Edit: Since ``Preloader.vue`` has been removed, see ``InitialSetup.vue`` instead.

Related issue:
This is a blocker on #291

BElluu:
Closed #583

### DIFF
--- a/src/renderer/store/maps.store.ts
+++ b/src/renderer/store/maps.store.ts
@@ -63,8 +63,8 @@ async function init() {
     window.downloads.onDownloadMapComplete((download) => {
         mapsStore.availableMapNames.add(download.name);
     });
-    let liveMaps = await db.maps.toArray();
-    let nonLiveMaps = await db.nonLiveMaps.toArray();
+    const liveMaps = await db.maps.toArray();
+    const nonLiveMaps = await db.nonLiveMaps.toArray();
 
     for (const map of liveMaps) {
         if (map.isInstalled) mapsStore.availableMapNames.add(map.springName);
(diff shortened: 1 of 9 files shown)


src/renderer/store/maps.store.ts
```diff
@@ -63,17 +63,46 @@ async function init() {
...
     try {
-        [liveMaps, nonLiveMaps] = await window.maps.fetchAllMaps();
+        const installedOnDisk = new Set(await window.maps.getInstalledMapNames());
+        mapsStore.availableMapNames = installedOnDisk;
+        await db.maps.toCollection().modify((map) => {
+            map.isInstalled = installedOnDisk.has(map.springName);
+        });
+        await db.nonLiveMaps.toCollection().modify((map) => {
+            map.isInstalled = installedOnDisk.has(map.springName);
+        });
```
Hectate: It seems like you could just call your new function `refreshMapsStore()` here to avoid duplicating the code.

Hectate (changes requested):
Looks good except for the one comment. Thanks!

src/renderer/store/maps.store.ts
BElluu: Yea, I forgot about it :D Thx!

Hectate (approved):

PR #604, opened 2026-06-22
Party UI components added

### LINKED ISSUE #390: Notify player when joined matchmaking through a party
This depends on party support (https://github.com/beyond-all-reason/bar-lobby/issues/389).
When a member of a party joins the matchmaking system, all other members also join the same queues. They are notified with the message [matchmaking/queuesJoined](https://github.com/beyond-all-reason/tachyon/blob/master/docs/schema/matchmaking.md#queuesjoined) and the client should update its state accordingly.

### LINKED ISSUE #593: Add Party features UI to the client

Hectate:
Closes #593 #390

Adds a Party view, as well as a little "popout" that indicates to the user that there is an active party and/or pending invite(s). The only way to create a party is to either invite a friend from NavBar > Friends, or use the Debug Sidebar's "Create Party" button. Clicking the "popout" will direct the user to the Party view.

The Party view has two tabs; active party and invites.

## Active Party
The current members and invitees are listed, along with "Kick Member" and "Cancel Invite" buttons. Party chat is present. Current matchmaking queues are presented along with the "Join/Download Assets" button. There is no filtering at the moment to hide queues that the party would not fit into.

## Invites
A simple panel that will list all current invites along with "Accept" and "Decline" buttons. Since we cannot know which member issued the invitation, all current members are identified in the party "description".


solventak:
Big change.  is this ready for review?

Hectate:
> Big change. is this ready for review?

Yes. For a larger PR, I have historically let it sit for a bit. It gives me a chance to review it with fresh eyes as well as anyone interested in looking at it. If it's "Draft" then it's there so people can see, test, comment, or contribute as they wish while understanding that it's not final.

I wasn't in a hurry on this either because I was watching [the `checkAssets` request](https://github.com/beyond-all-reason/tachyon/pull/146) implementation; but that can't be merged unless it's part of the `tachyon-protocol` NPM package or else `npm install` will not include that change.

PR #611, opened 2026-06-27
Perf: Debounce search input, throttle MaxPlayersFilter

Christopher-Hayes:
## Description

Adds debounce and throttle to the map list to reduce re-renders of the UI to improve performance.

This affects specifically the SearchBox, I wouldn't do this for certain text fields like login fields, but for Search debouncing should be fine.

Video walk-through: https://customer-hrn0pvgcql3sn5ug.cloudflarestream.com/66b8c7ce931ccc8f7a4b9fad5da8669a/watch (too big for GitHub)

## Throttle vs Debounce

I use throttle for the `<Range>` to show updates as the user is dragging the slider. I use debounce for the SearchBox because map list re-renders / DB re-queries will lag the user typing. Debounce means the list only updates on a 200ms pause in typing, but typing is not lagged at all.

## Technicalities

Note that something that threw me off as someone who spends more time in React than Vue is that useThrottleFn has a `trailing` default of `false` while, watchThrottled as a `trailing` default of `true`. Trailing means the latest update is always fired. The different defaults threw me off initially, so for clarity I included `trailing: true` in the `watchThrottled` options, even though that is the default value.

`@vueuse` is used for the debouncing and throttling, this library is already used elsewhere in the project.

### AI / LLM usage statement:

VS Code auto-complete: **GitHub Copilot**
LLM code generation: **Minimal/none**
LLM consultation: **Consulted Claude**

### DIFF
--- a/src/renderer/components/controls/SearchBox.vue
+++ b/src/renderer/components/controls/SearchBox.vue
@@ -90,9 +90,10 @@ export default defineComponent({
         const showClearIcon = computed(() => !!(props.clearIcon && props.modelValue.length > 0));
         const debouncedEmit = useDebounceFn((value: string) => emit("update:modelValue", value), 200);
 
-        function clear() {
-            emit("update:modelValue", "");
-        }
+function clear() {
+    debouncedEmit.cancel();
+    emit("update:modelValue", "");
+}
 
         function onInput(e: Event) {
             debouncedEmit((e.target as HTMLInputElement).value);
--- a/src/renderer/components/controls/SearchBox.vue
+++ b/src/renderer/components/controls/SearchBox.vue
@@ -88,7 +88,7 @@ export default defineComponent({
             return res;
         });
         const showClearIcon = computed(() => !!(props.clearIcon && props.modelValue.length > 0));
-        const debouncedEmit = useDebounceFn((value: string) => emit("update:modelValue", value), 200);
+        const debouncedEmit = useDebounceFn((value: string) => emit("update:modelValue", value), 200) as ReturnType<typeof useDebounceFn> & { cancel: () => void };
 
         function clear() {
             debouncedEmit.cancel();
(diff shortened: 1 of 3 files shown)


Christopher-Hayes:
**Update:**

Copilot PR Review suggested cancelling the debounce on `clear()`. This is a reasonable suggestion, but bar-lobby is on `@vueuse/core` v13 and `.cancel()` is not added until v14.

I added a TODO: note about upgrading since wider implications of upgrading to v14 may be out-of-scope for this PR. We're debouncing after 200ms, so it's unlikely `clear()` would be called on SearchBox within 200ms of the user typing. Therefore, not of real concern.

`.cancel()` added to `@vueuse/core` in v14: https://github.com/vueuse/vueuse/commit/4aa1803a0e61b9aeabbbe48cc033a589be331177

Hectate:
Looks like we have some issue with the `cancel` option. I pressed the "X" in the search bar to clear it and got a fatal error that `Uncaught TypeError: debouncedEmit.cancel is not a function`.

Hectate (changes requested):
Looks good if we can fix the the "cancel" issue.

Christopher-Hayes:
> Looks like we have some issue with the `cancel` option. I pressed the "X" in the search bar to clear it and got a fatal error that `Uncaught TypeError: debouncedEmit.cancel is not a function`.

Sorry about that, I swear I commented out the cancel with a note when I pushed the revert commit.

I have it commented out with a note to uncomment when `@vueuse/core` is upgraded to v14. It's not that important here, but if we use debounce more, it's good practice to avoid input bugs.

I didn't touch the `@vueuse/core` version because I saw there's at least 1 other PR that is upgrading dependencies, so didn't want to conflict with that.

Hectate (approved):

PR #612, opened 2026-06-27
fix: Debug logs are not shown.

Christopher-Hayes:
## Description

Unless I'm missing something, debug logs are never logged despite being used throughout the project.

For example:

https://github.com/beyond-all-reason/bar-lobby/blob/9373667ad95bb46e861ba4114a658a45f86412f6/src/main/main-window.ts#L79

is never logged.

## Cause

It seems related to the project using `pino.multistream`. Multistream sends logs to multiple locations. Pino's default level of `info` overrides any and all `debug` or `trace` logs _when using multistream._

https://github.com/beyond-all-reason/bar-lobby/blob/9373667ad95bb46e861ba4114a658a45f86412f6/src/main/utils/logger.ts#L41

The Pino docs state:

> In order for multistream to work, the log level must be set to the lowest level used in the streams array. Default is info.

_From [pinojs/pino API docs](https://github.com/pinojs/pino/blob/main/docs/api.md#pinomultistreamstreamsarray-opts--multistreamres)_

## The fix

This PR adds an ENV variable to set what that global level should be. The default is `info` which means by default logging works the same. But, developers can add `BAR_LOG_LEVEL=debug` or `BAR_LOG_LEVEL=trace` to show lower level logs.

## DX

We could even add scripts to make that easier, like:

```bash
npm run start:debug
```

```bash
npm run start:trace
```

But, I don't want to bloat the dev scripts unless contributors really want it.

### AI / LLM usage statement:

VS Code auto-complete: GitHub Copilot
LLM code generation: None
LLM consultation: Used Claude to investigate why Pino was tossing debug logs.

### DIFF
--- a/src/main/config/app.ts
+++ b/src/main/config/app.ts
@@ -105,6 +105,7 @@ console.log(`STATE_PATH: ${STATE_PATH}`);
 
 export const CONFIG_PATH = path.join(STATE_PATH, "config");
 export const LOGS_PATH = path.join(STATE_PATH, "logs");
+export const LOG_LEVEL = process.env.BAR_LOG_LEVEL || "info";
 
 // We will point engine at ASSETS_PATH as a base data directory to only read
 // data from, and at WRITE_DATA_PATH as data directory it can write to.
--- a/src/main/config/app.ts
+++ b/src/main/config/app.ts
@@ -109,14 +109,11 @@ export const CONFIG_PATH = path.join(STATE_PATH, "config");
 
 // Logging configuration
 export const LOGS_PATH = path.join(STATE_PATH, "logs");
-const LogLevelSchema = Type.Union([Type.String({ pattern: "^(fatal|error|warn|info|debug|trace|silent)$" }), Type.Integer({ minimum: 10, maximum: 60, multipleOf: 10 })]);
-const isNumberLogLevel = process.env.BAR_LOG_LEVEL !== undefined && process.env.BAR_LOG_LEVEL.trim() !== "" && !Number.isNaN(Number(process.env.BAR_LOG_LEVEL));
-const formattedLogLevel: Static<typeof LogLevelSchema> = isNumberLogLevel ? Number(process.env.BAR_LOG_LEVEL) : (process.env.BAR_LOG_LEVEL ?? "info").toLowerCase();
+const LogLevelSchema = Type.String({ pattern: "^(fatal|error|warn|info|debug|trace|silent)$" });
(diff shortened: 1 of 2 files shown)


src/main/config/app.ts
```diff
@@ -105,6 +105,7 @@ console.log(`STATE_PATH: ${STATE_PATH}`);
 
 export const CONFIG_PATH = path.join(STATE_PATH, "config");
 export const LOGS_PATH = path.join(STATE_PATH, "logs");
+export const LOG_LEVEL = process.env.BAR_LOG_LEVEL || "info";
```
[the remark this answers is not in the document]
Christopher-Hayes: Really just affects developers, runtime type-checking seems a bit much for a dev option.

src/main/utils/logger.ts
```diff
@@ -38,7 +38,15 @@ const logStream = PinoPretty({
     }),
 });
 
-const parentLogger = pino({}, pino.multistream([stdStream, logStream]));
+const parentLogger = pino(
```
p2004a, line 41: what is behavior when log level doesn't exist because user made typo?

Christopher-Hayes: I'd have to test, but this would only affect developers unless we would want users to set env vars for logging purposes.

Christopher-Hayes: Okay so it throws an error on load and does not open. For a feature we only expect developers to use, I think this is pretty okay because I'm trying to be mindful of introducing complexity when a simple error does the job. But, I'll look at a simple way of throwing an error without preventing the app from starting.

image

Christopher-Hayes:
I pushed a commit to add type-checking to log level env parsing. This uses `typebox` which is already used on the project for runtime checking.

Pino supports either string values or numbers, so the type-checking reflects that. Defaults to "info" like before.

```bash
# Using an incorrect value like so:
$ BAR_LOG_LEVEL=wrongvalue npm run start
# Will show an error, but won't stop the client from launching:
Invalid BAR_LOG_LEVEL: wrongvalue. Using default log level: info. Valid values are: fatal, error, warn, info, debug, trace, silent, or one of the pino level numbers (10, 20, 30, 40, 50, 60).
```

Got a little complicated with string vs number types to avoid Pino errors. But, previously if you used uppercase `DEBUG` rather than `debug` it would break, so this overall improves the usability.

Christopher-Hayes:
**Update** - just realized Pino only supports number log levels via a different option, `levelVal`, which we do not use. So, type checking number levels is not needed, I've removed number type-checking and updated the warning text.

Hectate (approved):

PR #618, opened 2026-06-28
Improvement management of logs

### LINKED ISSUE #610: Something wrong with log file management
No repro steps, but one uploaded folder looked like this:

Image

Note: this is used by moderator that double clicks etc replay files to open them. Speculation: Maybe there is some race or something when there is the new lobby instance started but old one is still running so it's opened in the second one? Or something like that?

BElluu:
Closed #610 

The problem with multiple weird logs what that, the app too early created instance of logger. Single-instance lock was too late.
These changes prevent empty second-instance logs,  nested `most_recent` files, stale ZIP archives after upload, and upload race conditions.

AI was used to discover some pieces of code, better understanding problem (talking with friend :) ) and write tests.

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -4,157 +4,43 @@
 
 import { app } from "electron";
 
+type SecondInstanceHandler = (argv: string[], workingDirectory: string) => void;
+
+const pendingSecondInstances: Array<{
+    argv: string[];
+    workingDirectory: string;
+}> = [];
+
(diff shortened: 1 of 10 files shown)


p2004a:
This is *a lot* of unrelated refactoring pieces. Please make it more reviewable, even just splitting changes per-commit in PR.

Hectate:
This PR has been supplanted by merged work. Closing.

PR #623, opened 2026-07-02
Resolve todo by implementing OAuth2 metadata caching

PrivacyIsARight:


src/main/oauth2/oauth2.ts
```diff
@@ -171,19 +286,38 @@ export async function renewAccessToken(): Promise<TokenResponse> {
...
+                accountService.saveRefreshToken(value.refreshToken);
+                accountService.saveToken(value.token);
+                log.info("Saved new tokens.");
+            })
+            .catch((error) => {
+                log.error(`Failed to renew access token in background loop: ${error}`);
+
+                // The errors persistent, stop any background attempts
+                if (error instanceof OAuth2Error && !error.isTransient) {
+                    stopTokenRenewer();
```
p2004a: You stop renever here, and now what? What happens next? Why is this ok?

src/main/oauth2/oauth2.ts
```diff
@@ -12,18 +12,104 @@ import { stringify } from "node:querystring";
...
+            if (currentFetchId === fetchId) {
+                activeFetchPromise = null;
+            }
+            throw error;
+        });
+
+    return activeFetchPromise;
+}
+
+export function invalidateAuthorizationServerMetadataCache(): void {
```
p2004a, line 114: Why this exists? Why are we invalidating cache?

src/main/oauth2/oauth2.ts
```diff
@@ -12,18 +12,104 @@ import { stringify } from "node:querystring";
...
+    cachedMetadata = null;
+    metadataExpiresAt = 0;
+    activeFetchPromise = null;
+    fetchId++; // Prevents any currently in-flight fetch from populating the cache
+    log.debug("OAuth2 metadata cache was manually invalidated");
+}
+
+function getCacheTtlMs(response: Response): number {
```
p2004a, line 122: Rolling this stuff ourselves doesn't have a lot of value. We will likely want to introduce caching in any other place in lobby, we start to copy around all this logic. You have here error code handling, retries and unnecessary request deduplication.

For caching for example there are libraries that handle this logic properly like for example `http-cache-semantics` but overall handling of different aspects of http requests is handled by libraries.

For new lobby, given we build on node, we can use `undici` which is developed by Node.jd team, features from undici over time get available in node etc (becasue node's global fetch is from bundled undici), so for now, it might be a good fit.

We do something like:

```js
import { interceptors, fetch, Agent } from "undici";

const dispatcher = new Agent().compose(interceptors.cache());

(...)

await fetch(getOAuthWellKnownURL(), {dispatcher});
```

and that pretty much resolves that specific todo.

p2004a: This is LLM generated comment response, isn't it?

p2004a: This does not answer my question and does not make your comment any less hallucinated and wrong.

p2004a: This does not answer my question and does not make your comment any less hallucinated and wrong.

p2004a: > This is LLM generated comment response, isn't it?

p2004a: Per https://github.com/beyond-all-reason/bar-lobby/blob/master/AI_POLICY.md, many points apply here, but specifically

> Contributors remain responsible for research, accuracy, and clarity.

If all you will do in the PR review is point my comments at LLM, get responses, and copy back that's a waste of everybody's time and very disrespectful.

---

> undici implements a strict RFC-9111 shared cache

1. Is not exactly true, not a concern. What is type of cache is configurable https://github.com/nodejs/undici/blob/cb4c2f1f7231d82bd699aa3a09efb5a235faaa3b/types/cache-interceptor.d.ts#L40
2. We control server, we can ensure caching headers are set appropriately

> because this is running inside an electron process, the global fetch is overriden to pipe requests through net

This is just wrong. `src/main/oauth2/oauth2.ts` runs inside of the node.js process. Node.js process networking does not run via chromium networking, global fetch does not run via chromium networking. That's why https://www.electronjs.org/docs/latest/api/net#netfetchinput-init exists, to allow to do `fetch` via chromium stack, it's not by default, we don't do it anywhere in the lobby.

> Also, it's just another dependency which adds extra work to the table to keep it updated, watch out for any possible supply chain vulnerabilities, etc.

That's an opinion, with a very local-view of the bar lobby in this PR. Using better library across new lobby to handle retries and caching will be beneficial. We for example use `axios` in one place already but I would prefer to use more standard `fetch` with undici and I would want to consider applying it everywhere via `setGlobalDispatcher`.

thvl3: > Also, it's just another dependency which adds extra work to the table to keep it updated, watch out for any possible supply chain vulnerabilities, etc.

This is a bad faith argument against the package. This statement does not help your credibility.

src/main/oauth2/oauth2.ts
p2004a: I'm not convinced we need this complexity and that purging cache is beneficial. Especially on transient network errors.

src/main/oauth2/oauth2.ts
p2004a: > Unless your planning to add a bigger pr that updates many files to use undici at once, then I don't see how it would be more beneficial.

I addressed that already:

> Using better library across new lobby to handle retries and caching will be beneficial. We for example use axios in one place already but I would prefer to use more standard fetch with undici and I would want to consider applying it everywhere via setGlobalDispatcher.

> And it introduces issues and complexity of its own.

That's hand-waving.

> The original code had many issues including race conditions

And by the way your PR description should reflect that more clearly. Right now PR description isn't very useful.

thvl3: > But when you're using npm, the thing that's vulnerable and has had the most supply chain attacks, it's obviously a worry.

What do you think version pinning is?

PrivacyIsARight: > > But when you're using npm, the thing that's vulnerable and has had the most supply chain attacks, it's obviously a worry.
> 
> What do you think version pinning is?

What? Version pinning is nice but it doesn't protect you if something already exists. Anyways, I'm going to keep resolving more edge cases.

p2004a:
That doesn't really matter and we can drop that particular check entirely. What is kind of relevant is verifying `issuer` field: right now even that isn't critical for us as we use only single authorization server in client, but it's a defense in depth just in case that would somehow change in the future.

Issue #607, opened 2026-06-26
Download speed progress


Christopher-Hayes:
## Description

Small thing—the "Downloading game" screen shows the bytes written to disk, not the download speed. Download clients often show the actual network download speed.

<img width="755" height="681" alt="Download screen saying 'Initial Setup, Downloading game' with the speed 105 megabytes per a second." src="https://github.com/user-attachments/assets/7331e864-a042-4e5d-9c54-d3a9182387e8" />

_100 MB/s download is impossible with my connection._

Not a big deal, I think this is fine as a **won't fix.**

Only options for showing the download speed is either an update to [beyond-all-reason/pr-downloader](https://github.com/beyond-all-reason/pr-downloader) or approximating download speed with a rough estimate of GZIP decompression ratio.

solventak:
I confirmed the original issue: the Rapid progress counter we currently surface in Lobby is logical content progress, not raw network transfer bytes. Current state takes the SDP’s uncompressed file sizes, sums the missing unique pool objects, then maps real compressed download progress onto that scale. So using `delta(progress) / delta(time)` without specifying that it is _logical_ MB/s not _network_ MB/s is misleading.

I tried to solve this entirely in Lobby:

- resolve the exact missing Rapid object set;
- `HEAD` each pool object to get compressed `Content-Length`;
- monitor the matching local `.gz` files for payload progress/rate.

That works technically, but it is not viable for the normal path. A game can involve ~17k small objects. Even with 128 concurrent requests, the `HEAD` preflight is roughly a 30+ second extra request pass, and the actual download has to make most of those same object requests anyway.

The new direction is:

- keep current existing logical `current / total` for progress and ETA;
- add a separate native raw network-rate metric in `pr-downloader`, which is decoupled from logical progress pct and ETA;
- show those independently in Lobby.

This avoids blocking downloads on a full metadata preflight, keeps the existing progress bar useful, and stops us pretending the logical counter is network throughput. The numbers will not necessarily move in lockstep, but they are measuring different things.

I have a PR coming shortly for that direction, pending team confirmation.

Christopher-Hayes:
@solventak sounds reasonable. I'm not working on a PR, so feel free to make one.

p2004a:
pr-downloader fetches packages by first pulling the sdp index and resolving files from there. sdp index files do not have compressed size, only uncompressed one, that's WAI. pr-downloader does not know how much compressed data it needs to fetch.

[...]

PR #633, opened 2026-07-14
Restore pool prefetch and network-speed progress for Initial Setup

### LINKED ISSUE #632: Initial Setup skips pool prefetch because it creates an empty pool directory before checking for it
On a fresh Initial Setup, Lobby is intended to download and extract the `data.7z` before using `pr-downloader` to reconcile the remaining game-content delta.

That does not currently happen when the user confirms an assets/install path. The path-initialization flow creates an empty pool directory before the prefetch check runs. Since the prefetch checks only whether that directory exists, it treats the empty directory as an already-seeded pool and skips the bulk archive download.

### Reproduction

1. Start Lobby without an installed engine/game.
2. Use a completely fresh assets/install directory, or choose a directory that does not yet exist.
3. Continue through Initial Setup and confirm the assets path.
4. Enable debug logging.
5. Observe the game-install stage.

### Observed behavior

Lobby logs:

```text
game-content.ts - Preloading pool data
pool-cdn.ts - Pool folder already exists, skipping download
```

This occurs even when the selected install directory did not exist before Initial Setup.

The subsequent game installation uses `pr-downloader` to download the full game object set rather than first downloading the bulk pool archive.

### Expected behavior

For a fresh installation, Lobby should:

1. Download the bulk `data.7z` pool archive.
2. Extract it into the selected assets directory’s pool.
3. Run `pr-downloader` only to reconcile any remaining package delta.

The presence of an empty pool directory should not suppress the prefetch.

### Impact

This regression bypasses the intended bulk initial-install path and makes a first-time install perform many individual Rapid object downloads through `pr-downloader`.

It also complicates download telemetry unnecessarily: the bulk `data.7z` downloader already exposes real transferred and total archive bytes, while  `pr-downloader` progress values are logical content progress and should not be interpreted as raw network transfer bytes.

solventak:
Closes: #632 #607 

Fixes the Initial Setup regression where Lobby creates an empty `<assets>/pool` directory during path initialization, then treats that directory as proof that the bulk pool archive has already been installed.

This restores the intended initial-install sequence:

```text
data.7z pool seed → pr-downloader reconciliation → local SDP validation
```

rather than skipping `data.7z` and downloading the full game object set individually through pr-downloader.

## network-speed telemetry

A side effect of this change is that we go back to reporting real network speed for the pool download since we aren't using pr-downloader the same way.

Restoring this bulk-prefetch path therefore restores meaningful Initial Setup percentage, transferred bytes, network speed, and ETA for the main download.

## Tests

Added coverage for:

- clearing a partial pool while retaining the adjacent `data.7z` archive;
- publishing a `pool-data` start record before a resumed archive transfer's first progress event.

Closes #632 and #607.

_AI disclosure: used for research and writing tests._


Hectate:
So I did some testing of this PR and I'm concerned about how we're holding back the startup time and also not communicating that to the user. I just timed 40 seconds to get to the Login button, while the entire time it just spun and said "downloading game" with no actual context about what was happening. A first-time launch looks pretty reasonable, with things happening and the bar moving, but a second launch or more I'm just waiting around for seemingly nothing to happen (in fact, my download speed is faster than this).

So for this to work, the actual activity needs to be communicated to the user. We also need to find some way to speed it up so every launch doesn't have a minute extra time added to it.

solventak:
You're right there's been a regression here.

solventak:
The slowdown is a full verification step.

The regression is the combination of:

- unconditional full integrity validation on every launch;
- duplicated validation even when `downloadGame()` performed no reconciliation;
- no validation-specific progress reporting;
- misleading “Downloading game” UI while the process is actually verifying local content.

Defect: if the installed package passed validation and `downloadGame()` skipped because the requested version was already installed, validating the same SDP again proves nothing new.

Defect: removing the duplicate still leaves one complete integrity scan, so it may become roughly twice as fast but will not necessarily become “very very quick.”

solventak:
I removed the file validation step as that is new functionality that should be separated into its own PR and tested separately from this fix.

Ready for another review.

Hectate (approved):

PR #634, opened 2026-07-15
Fix map spring name having trailing "undefined"/space when mapinfo.lua version is missing

### LINKED ISSUE #629: Map "paradise_lost_v4" does not retain "Downloaded" status between launches
During matchmaking tests, I noticed that I would have to trigger a download of the `Paradise_Lost_V4` map upon each launch of the client to satisfy the queue's asset requirements.

A quick check shows that this file _is_ located in my assets folder alongside other, recognized maps.

Image

There is no output in the terminal to suggest any issues with the map being recognized upon launch. This will require further research to see if this specific map may have some overlooked property that is affecting the client's ability to parse it upon launch.

solventak:
Closes #629

Maps whose `mapinfo.lua` is missing a `version` field (or has an empty one) were never recognized as "already downloaded" on subsequent launches, because their derived spring name was malformed.

## Problem

Any map with a `version` that is undefined or empty fell into the branch that appends the version to the name, producing `"Paradise Lost V4 undefined"` or `"Paradise Lost V4 "` (trailing space) instead of `"Paradise Lost V4"`. That string is used as the lookup key for "is this map installed", so it never matched the name reported by the matchmaker/queue, and the client re-triggered a download of the map on every launch even though the file was already present.

### Tests

Added `tests/unit/main/ultrasimple-map-parser.spec.ts` covering different cases, including a regression test for the error case.

### AI / LLM usage statement

AI for tests, Codex.

### DIFF
--- a/tests/unit/main/ultrasimple-map-parser.spec.ts
+++ b/tests/unit/main/ultrasimple-map-parser.spec.ts
@@ -50,9 +50,9 @@ describe("UltraSimpleMapParser", () => {
         expect(result.springName).toBe("Some Map 1.2");
     });
 
-    // Regression test: a missing "version" field used to fall through to the 
-    // "append version" branch, producing "<name> undefined"/"<name> " which 
-    // never matched the expected spring name, so the map was re-downloaded 
+    // Regression test: a missing "version" field used to fall through to the
+    // "append version" branch, producing "<name> undefined"/"<name> " which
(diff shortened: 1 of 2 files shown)


Hectate (approved):
Thanks! I removed the issue number since it won't be relevant once closed (and we can always `git blame` if we need to research).

ak-traversal:
Ty!

PR #635, opened 2026-07-15
feat: add sequential map download queue

### LINKED ISSUE #423: Give map downloads a max concurrent download limit

### LINKED ISSUE #627: Option to "Download Selected/All" from Maps view
It would be nice to have a download all map button in the library/maps section like in the image

Image

solventak:
Closes #627 

_Big diff is mostly just tests.  < 400 lines change without that._

Adds a shared FIFO queue for map downloads. The existing Downloads navbar popover now shows the ordered map queue, including the active `Downloading` map and queued maps.

Adds ability to select multiple maps for download, which are eligible (not already downloaded).
Add functionality to download all: `Select All -> Download Selected`

**Can extended eventually to support concurrent downloads, but right now downloads one at a time.**

_AI disclosure: research, investigation, preliminary implementation pass. Codex 5.6 Terra_

### DIFF
--- a/src/main/content/maps/map-download-queue.ts
+++ b/src/main/content/maps/map-download-queue.ts
@@ -1,0 +1,84 @@
+// SPDX-FileCopyrightText: 2026 The BAR Lobby Authors
+//
+// SPDX-License-Identifier: MIT
+
+import { Signal } from "$/jaz-ts-utils/signal";
+
+export type MapDownloadQueueEntry = {
+    springName: string;
+    status: "queued" | "downloading";
+};
+
+type QueuedMapDownload = {
+    springName: string;
+    resolve: () => void;
+    reject: (error: unknown) => void;
+};
+
+/** Owns the serial map-download lifecycle and exposes only observable queue state. */
+export class MapDownloadQueue {
+    public readonly onChanged: Signal<MapDownloadQueueEntry[]> = new Signal();
+
+    private readonly queue: QueuedMapDownload[] = [];
+    private readonly pendingDownloads = new Map<string, Promise<void>>();
+    private activeDownload?: QueuedMapDownload;
+    private isProcessing = false;
+
+    public constructor(private readonly download: (springName: string) => Promise<void>) {}
+
+    public getSnapshot(): MapDownloadQueueEntry[] {
+        return [
+            ...(this.activeDownload ? [{ springName: this.activeDownload.springName, status: "downloading" as const }] : []),
+            ...this.queue.map((download) => ({ springName: download.springName, status: "queued" as const })),
+        ];
+    }
+
+    public enqueue(springName: string): Promise<void> {
+        const pendingDownload = this.pendingDownloads.get(springName);
+        if (pendingDownload) return pendingDownload;
+
+        let resolve!: () => void;
+        let reject!: (error: unknown) => void;
+        const download = new Promise<void>((resolveDownload, rejectDownload) => {
+            resolve = resolveDownload;
+            reject = rejectDownload;
+        });
+
+        this.pendingDownloads.set(springName, download);
+        this.queue.push({ springName, resolve, reject });
+        this.dispatchChanged();
+        void this.process();
+        return download;
+    }
+
+    private async process() {
+        if (this.isProcessing) return;
+
+        this.isProcessing = true;
+        try {
+            while (this.queue.length > 0) {
+                const queuedDownload = this.queue.shift();
+                if (!queuedDownload) continue;
+
+                this.activeDownload = queuedDownload;
+                this.dispatchChanged();
+                try {
+                    await this.download(queuedDownload.springName);
+                    queuedDownload.resolve();
+                } catch (error) {
+                    queuedDownload.reject(error);
+                } finally {
+                    this.pendingDownloads.delete(queuedDownload.springName);
+                    this.activeDownload = undefined;
+                    this.dispatchChanged();
+                }
+            }
+        } finally {
+            this.isProcessing = false;
+        }
+    }
+
+    private dispatchChanged() {
+        this.onChanged.dispatch(this.getSnapshot());
+    }
+}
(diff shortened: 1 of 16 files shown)


src/renderer/components/navbar/Downloads.vue
```diff
@@ -7,9 +7,21 @@ SPDX-License-Identifier: MIT
...
+                            <div class="downloads__status">
+                                {{
+                                    download.status === "downloading"
+                                        ? t("lobby.navbar.downloads.downloading")
+                                        : t("lobby.navbar.downloads.queued")
+                                }}
+                            </div>
+                        </div>
+                    </div>
+                    <div v-for="download in nonMapDownloads" :key="`${download.type}:${download.name}`" class="downloads__download">
```
Hectate, line 24: By splitting this into "maps" and "non maps" downloads, you've accidentally attached the progress bar to only the non-maps downloads. We should have a progress bar for both types, of course.

src/renderer/components/maps/MapListComponent.vue
```diff
@@ -14,13 +14,30 @@ SPDX-License-Identifier: MIT
...
                 <div class="maps">
                     <TransitionGroup name="maps-list">
-                        <MapOverviewCard v-for="map in maps" :key="map.springName" :map="map" @click="mapSelected(map)" />
+                        <div v-for="map in maps" :key="map.springName" class="map-card">
+                            <MapOverviewCard :map="map" @click="mapSelected(map)" />
+                            <div v-if="isMapDownloadEligible(map)" class="map-card__selection" @click.stop>
+                                <Checkbox
+                                    :modelValue="selectedMapNames.has(map.springName)"
+                                    @update:modelValue="toggleMapSelection(map.springName)"
+                                />
```
Hectate, line 38: It appears that these (empty) checkboxes are visibly flying across the MapListComponent to to top left corner as they fade out when a map with a checkbox is hidden. This is most easily visible when doing large numbers of them at once and a cluster all zip up to the corner at once.
They should stay in position with their matching cards and fade out with them, not separately.

src/main/content/maps/map-download-queue.ts
```diff
@@ -0,0 +1,84 @@
...
+    reject: (error: unknown) => void;
+};
+
+/** Owns the serial map-download lifecycle and exposes only observable queue state. */
+export class MapDownloadQueue {
+    public readonly onChanged: Signal<MapDownloadQueueEntry[]> = new Signal();
+
+    private readonly queue: QueuedMapDownload[] = [];
+    private readonly pendingDownloads = new Map<string, Promise<void>>();
+    private activeDownload?: QueuedMapDownload;
```
Hectate, line 24: Do you expect to extend this in the future to permit more than a single download at a time from the queue?

Hectate (changes requested):
You have completely missed the DownloadContentButton component which is used in significant ways in the client already to download one or more items. It does its own form of a hidden queue, so converting it over to use this queue system will be a requirement for this PR to be accepted.

src/main/content/maps/map-download-queue.ts
solventak: Yes.

src/renderer/components/navbar/Downloads.vue
solventak: What other downloads can happen in game beyond map downloads?

solventak: Ah I see we also have game content.

solventak:
I would recommend retiring the current `DownloadContentButton` abstraction, but preserve its reusable presentation in a simpler component.

### Why

The component currently mixes three responsibilities:

1. Deciding whether maps, games, and engines are ready.
2. Managing and awaiting downloads.
3. Rendering the Download/Start/Watch/Launch button.

With a global queue, the first two responsibilities belong in shared application state, not a transient button.

Hectate:
> I would recommend retiring the current `DownloadContentButton` abstraction, but preserve its reusable presentation in a simpler component.
> ### Why
> 
> The component currently mixes three responsibilities:
> 
>     1. Deciding whether maps, games, and engines are ready.
> 
>     2. Managing and awaiting downloads.
> 
>     3. Rendering the Download/Start/Watch/Launch button.
> 
> 
> With a global queue, the first two responsibilities belong in shared application state, not a transient button.

Items 1 and 2 are in service of item 3, so I'm not sure how you consider these "mixed". The button keeps the user informed as to state and progress. Since the button can appear in so many different contexts, it has some flexibility built into the design. 

From my point of view, the only changes needed are to update `beginDownload()` to use the queue system instead, and fix `isDownloading` to track the downloads of interest from the store/queue since it's no longer internal.

If you have a different idea, we will judge it on it's own merits.

burnhamrobertp:
I think a broader problem exists that is too easily missed and actually requires a different approach. I've started on this and should have it finished today, although its a much bigger undertaking than the changes in this PR.

tl;dr concurrency is not enforceable per-type content type (i.e. "download only 4 maps at a time"). Concurrency needs to be enforced more globally, such as between the Server and Client; if not globally at the client level itself. Downloading 15 things at once is bad because it starves any download from completing effectively. Consequently, governing download concurrency based on individual content type will only ever be a half-solution since "Only download 4 maps at a time" doesn't preclude "Only download 2 engines at a time" and "Only download 3 game versions at a time" all simultaneously being true.

burnhamrobertp:
A broader refactor of the content-downloading system was necessary; I think https://github.com/beyond-all-reason/bar-lobby/pull/675 supplanted these changes and this can be closed

Hectate:
> A broader refactor of the content-downloading system was necessary; I think #675 supplanted these changes and this can be closed

Correct, yes the changes represented here are supplanted by other work and now unnecessary.

PR #636, opened 2026-07-15
fix: select all maps across paginated results

solventak:
**READ: #635 MUST be merged first! This PR is stacked**

Fixes **Select all available** in Library → Maps so it selects every eligible map matching the current catalog view, rather than only the maps currently loaded by infinite-scroll pagination.

- Keep the rendered map list paginated.
- Run an additional unlimited catalog query for bulk selection and submission.
- Share the same search, filter, and sort logic between the rendered and full-catalog queries.
- 
_AI disclosure: investigation, initial implementation. Codex Terra 5.6_


burnhamrobertp:
Given that this was dependent upon #635 , I think https://github.com/beyond-all-reason/bar-lobby/pull/675 supplanted these changes and this can be closed

Hectate:
> Given that this was dependent upon #635 , I think #675 supplanted these changes and this can be closed

Correct, yes this will not be merged due to reliance on a prior PR that will not be merged. Feature-wise, this is still a desirable goal.

PR #637, opened 2026-07-15
feat: show active map queue count on Downloads

### LINKED ISSUE #423: Give map downloads a max concurrent download limit

solventak:
**READ: #635 MUST be merged first! This PR is stacked**

Adds a badge to the existing Downloads navbar button showing the current number of map downloads that are queued or actively downloading. The badge is hidden when the map queue is empty, while the button’s existing progress styling and popover behavior remain unchanged.

_AI disclosure: investigation, initial implementation. Codex Terra 5.6_

### DIFF
--- a/src/main/content/maps/map-content.ts
+++ b/src/main/content/maps/map-content.ts
@@ -15,7 +15,6 @@ import chokidar, { FSWatcher } from "chokidar";
 import { UltraSimpleMapParser } from "$/map-parser/ultrasimple-map-parser";
 import { removeFromArray } from "$/jaz-ts-utils/object";
 import { engineContentAPI } from "@main/content/engine/engine-content";
-import { settingsService } from "@main/services/settings.service";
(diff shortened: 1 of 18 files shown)


burnhamrobertp:
Given the dependence upon #635, I think https://github.com/beyond-all-reason/bar-lobby/pull/675 supplanted these changes and this can be closed

Hectate:
> Given the dependence upon #635, I think #675 supplanted these changes and this can be closed

Correct, thanks. This will be closed as the structure relies on a PR that will not be merged. The concept is still open to consideration, but as a UI item it will be deferred until the production team provides their input.

PR #641, opened 2026-07-16
Fix bot faction propagation to engine start scripts

### LINKED ISSUE #640: Propagate bot faction to the engine start script
## Bug

Offline battles retain a bot's selected `faction` in lobby battle state, but launching the battle does not pass that faction to the engine.

`Bot` in `src/main/game/battle/battle-types.ts` has an optional `faction`, and the generated engine start-script `Team` supports `side`. However, `StartScriptConverter.offlineBattleToStartScript()` does not set `Team.side` from a bot's faction.

As a result, any bot-faction selection made in the lobby is lost at launch and the engine falls back to its default faction behavior.

## Expected behavior

For each offline bot participant with a selected faction, its generated start-script team has the corresponding `side` value so the engine launches that bot with the selected faction.

solventak:
## Summary
Propagates a selected offline bot faction into its generated engine team `side` field, so bot-faction choices persist at launch while bots without a selected faction retain engine defaults. Fixes #640.

### DIFF
--- a/src/main/utils/start-script-converter.ts
+++ b/src/main/utils/start-script-converter.ts
@@ -147,6 +147,7 @@ class StartScriptConverter {
                     players.push(player);
                 } else {
                     team.teamleader = teamMember.host;
+                    team.side = teamMember.faction;
                     const bot: Bot = {
                         id,
                         team: team.id,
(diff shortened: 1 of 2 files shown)


Hectate (approved):
I see that you're handling the ability to set a faction in the other PR, so this is good. Thanks!

PR #643, opened 2026-07-16
Fix stale bot-option updates after participant removal

### LINKED ISSUE #642: Bot Configure flow throws when its participant is removed before an option update
## Bug

The offline-battle renderer can throw an uncaught error when a bot-options update is emitted after the corresponding bot is no longer in `battleStore.teams`.

Observed renderer error:

```text
Uncaught Error: Failed to find bot BARbarIAn (1)
```

`LuaOptionsModal` can emit a later `set-options` event from the Configure flow. Meanwhile, battle reconciliation may remove or relocate bots when a map/start-box/team configuration changes.

## Expected behavior

If the configured bot has been removed before an option update arrives:

- do not throw an uncaught renderer error;
- close/dismiss the obsolete configuration surface, or safely ignore the stale update;
- when the participant still exists, update the resolved current participant (`foundBot`), rather than relying on the captured `bot` reference.

## Evidence

Log archive:
https://log.beyondallreason.dev/logs-20260716T1042-vY6rxOmL.zip

solventak:
Prevent the offline battle renderer from throwing when an open bot Configure modal submits options after its bot has been removed during team reconciliation. 

**The update now targets the resolved live participant** if it no longer exists, the stale update is safely ignored and the obsolete modal closes. Closes #642.

### DIFF
--- a/src/renderer/components/battle/BotParticipant.vue
+++ b/src/renderer/components/battle/BotParticipant.vue
@@ -12,7 +12,31 @@ SPDX-License-Identifier: MIT
                 <Icon v-else-if="isScavenger(bot)" :icon="robotAngry" />
                 <Icon v-else :icon="robot" />
             </div>
-            <div>{{ bot.name }}</div>
+            <div class="bot-name">{{ bot.name }}</div>
+            <div v-if="hasInlineControls" class="bot-controls">
+                <label class="bot-control">
+                    <span>{{ t("lobby.components.battle.botParticipant.faction") }}</span>
+                    <Select
+                        data-test="bot-faction"
+                        :modelValue="bot.faction"
+                        :options="factionOptions"
+                        optionLabel="name"
+                        optionValue="value"
+                        @update:model-value="setBotFaction"
+                    />
+                </label>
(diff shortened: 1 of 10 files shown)


src/renderer/components/battle/BotParticipant.vue
```diff
@@ -93,7 +94,13 @@ async function configureBot() {
 }
 
 function setBotOptions(options: Record<string, unknown>) {
-    battleActions.updateBotOptions(props.bot, options);
+    if (!battleActions.updateBotOptions(props.bot, options)) {
```
Hectate, line 97: While I get that it's basically a "success/failure" response, I would prefer that `updateBotOptions` not carry and additional meaning with this return value. If we do it this way, in the future this function could end up in an online lobby and we'd have to just repeat this warning there too.

Instead, we should have itself trigger the notification if `findBot` is not successful. This ensures that any uses of this elsewhere will behave equally not not be subject to (possibly forgotten about) return value behavior.

Hectate (changes requested):

PR #650, opened 2026-07-26
Fix lots of typos

PrivacyIsARight:


Hectate (approved):
Thanks! I'll note that the files changed in `lang/en` are imported translation files and not directly used by `bar-lobby` (except to the degree that they're combined into the final set of translation strings), so those typos are an upstream issue. For now, it won't change anything to fix them locally, but it won't hurt anything either.

PR #651, opened 2026-07-26
Polish markdown, and fix a typo

PrivacyIsARight:


Hectate:
If you've noticed that the test failed on `prettier`, you can see that we use a specific formatting, inclusive of these files. If there's a case to be made to update that, then feel free to share. Otherwise, we'll have to fix the typos only and leave out other formatting changes.

Hectate (approved):

PR #654, opened 2026-07-27
Tidy up leftover files in the logs directory

burnhamrobertp:
A few things in the logs directory don't clean up after themselves properly:

- packLogFiles copies the newest log to `<name>most_recent.log` right next to the real logs, and getSortedLogFiles globs `*.log`, so if that copy ever survives it gets treated as a real log and picks up the suffix again on the next run
- the copy only gets deleted on the happy path, so a failed pack leaves it behind
- the zip removal after upload isn't awaited, and doesn't happen at all if the upload throws

First commit pulls the log file name into its own small module so the logger and the log service aren't each carrying their own idea of the format. That lets the service tell real run logs apart from anything else sitting in the folder. There's a test that fails if the name the logger builds stops matching what the service filters on, which felt better than a comment asking people to keep the two in sync.

Second commit uses that to filter, moves the copy cleanup into a finally, and awaits the zip removal.

Related to #610. I couldn't tell how much of what's in the screenshot there this accounts for, so I'd treat it as tidying rather than a fix for that report.

Developed with Claude Code.

### DIFF
--- a/src/main/services/log.service.ts
+++ b/src/main/services/log.service.ts
@@ -10,7 +10,10 @@ import { packSpecificFiles } from "@main/utils/pack-7z";
 import { randomBytes } from "crypto";
 import axios from "axios";
 import { logger } from "@main/utils/logger";
+import { LOG_FILE_NAME } from "@main/utils/log-file-name";
 import { configService } from "@main/services/config.service";
+
+const PACK_COPY_GLOB = "*most_recent.log";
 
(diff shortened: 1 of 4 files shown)


Hectate:
If you want to check the conflict (it should just be swapping the settings for config from a quick glance) I'll get this added.

Hectate (approved):

PR #656, opened 2026-07-27
Preserve Tachyon failure details across IPC

### LINKED ISSUE #588: Errors thrown from Main to Renderer are stripped of details
As a result of how IPC works, if ``main`` throws an error in response to some ``ipcRenderer.invoke()`` call, the Error arrives in Renderer without custom properties and is technically an entirely different Error object. See [notes in Electron docs.](https://www.electronjs.org/docs/latest/api/ipc-renderer#ipcrendererinvokechannel-args)

As a result of this, the Tachyon failure reasons (among other things) in the Renderer require parsing of the Error string, making validation more fragile and open to mistakes. For example,[ this line](https://github.com/beyond-all-reason/bar-lobby/commit/6c1f0a82a5dea88cc91aeec9c17126a989543582#diff-cb685be91221553dbb89698891e40d678c210a4fe863c89f45eb3dc8d29294c3R164) is used to check the error message to see if it includes a specific Tachyon failure reason, which is not ideal.

Work in #547 presented an approach where requests/responses would be structured differently and ensure that the complete error information arrives at the Renderer. Something to this effect would benefit not only error handling, but also permit better, translation-friendly error notifications in the client.

burnhamrobertp:
Closes #588.

There are two request functions now rather than a fix to the existing one, because throwing can only carry a string. The renderer would have to pick the reason back out of the message text, which is the fragile bit in the first place. Not throwing is the only way to hand over something structured.

Only matchmaking/queue uses the new one, since it was the one doing the string matching. The rest are untouched.

Tests, typecheck and lint pass. I haven't run it against a live server, so I haven't actually watched the version mismatch case happen.

AI disclosure: written with Claude Code. I directed the approach and reviewed the diff.

### DIFF
--- a/src/main/services/tachyon.service.ts
+++ b/src/main/services/tachyon.service.ts
@@ -67,6 +67,10 @@ function registerIpcHandlers(webContents: BarIpcWebContents) {
     ipcMain.handle("tachyon:request", async (_event, command, args) => {
         return await tachyonClient.request(command, args);
     });
+
+    ipcMain.handle("tachyon:requestStructured", async (_event, command, args) => {
+        return await tachyonClient.requestStructured(command, args);
+    });
 }
 
 export const tachyonService = {
--- a/src/main/tachyon/tachyon-client.ts
+++ b/src/main/tachyon/tachyon-client.ts
@@ -112,24 +112,22 @@ export class TachyonClient {
         });
     }
 
-    public async request<C extends GetCommandIds<"user", "server", "request">>(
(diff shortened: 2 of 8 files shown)


Hectate:
Narrow test for this specific matchmaking failure worked. For reference, it is possible introduce this error by simply hard-coding an incorrect version number for the request. For example:
```ts
        const response = await tachyonRequest("matchmaking/queue", {
            // queues: [{ id: playlist.id, version: playlist.version }],
            queues: [{ id: playlist.id, version: "000000" }],
        });
```
I'll do some further testing and review, thanks!

Hectate (approved):

PR #657, opened 2026-07-27
Handle Tachyon disconnects and reconnects

### LINKED ISSUE #652: Client should cleanly handle disconnects and reconnects
Currently, the client does not properly handle disconnections (e.g. Tachyon server crashing, internet connection dropped). The `tachyon-client.ts` has [an event listener for websocket closing](https://github.com/beyond-all-reason/bar-lobby/blob/8299b31202ef5f9bb0c45386043fe38ae72faa86/src/main/tachyon/tachyon-client.ts#L79-L101), but parties, lobbies, etc. do not watch for this status and respond appropriately.

If the connection is dropped, the state should be maintained until reconnection provides fresh updates. The user should be shown a "Attempting to reconnect" notification, view overlay, or similar that indicates reconnection attempts are ongoing. Online-only actions should be disabled during the connection loss. A "Cancel Reconnection Attempts" option should be presented to the user, which will end the reconnection attempts and place the user into a fully Offline mode.

burnhamrobertp:
Closes #652. Keeps party, lobby and matchmaking state when the connection drops and blocks interaction behind a reconnecting overlay, rather than clearing it all and dropping you to offline.

That reversal came out of the discussion on the issue. The server holds your session open across a dropped socket, so wiping the client's copy tells you you are out of a party and a queue you are almost certainly still in, and then puts you back in a few seconds later. Blocking the UI while we retry is the compromise p2004a settled on. Going offline is the only thing that discards any of it now, and the overlay offers that as the way out.

Each store registers its own cleanup against a signal rather than one central place reaching into all of them. Not tidiness: having the tachyon store import the lobby store pulls in the battle store, which reads the game store while that is still being evaluated, and the renderer won't start. The disconnect handler also only reacts to the first close of an outage, since every failed retry opens a socket that immediately closes and fires another, which would otherwise restart the ten second countdown each time.

Your selected queue, the download requirements and your DMs survive going offline on purpose. None of them are session state, and emptying the download list broke the matchmaking page, which reads straight out of it. Nothing navigates you anywhere either, since a two second blip shouldn't lose your place.

Reconciling against the server on reconnect isn't here, that's #664 and the three under it. Until they land, an outage longer than the server's session timeout leaves state on screen that is already gone.


src/renderer/store/me.store.ts
```diff
@@ -129,8 +129,13 @@ window.tachyon.onEvent("friend/removed", async (event) => {
     await unsubscribeFromUsers([event.from]);
 });
 
+// Identity fields survive; they're persisted in db and used while offline.
+function clearOnlineState() {
+    subsManager.clearAllFromList(friendsSymbol);
```
Hectate, line 134: Using `clearAllFromList` (both here and elsewhere in this PR) will result in a downstream `user/unsubscribeUpdates` request sent to the server that we just disconnected from. This results in an error displayed in the client.

That's happening in `tachyon.store.ts:159`, and we could guard against that by simply checking if we're online and declining to send the request if not.

src/renderer/store/tachyon.store.ts
```diff
@@ -44,6 +46,33 @@ async function connect() {
     }
 }
 
+function stopReconnecting() {
+    if (tachyonStore.reconnectInterval) {
+        clearInterval(tachyonStore.reconnectInterval);
+        tachyonStore.reconnectInterval = undefined;
+    }
+}
```
Hectate, line 57: I see here that we turn of reconnection attempts once we successfully reconnect; what if the user wants to completely cancel trying to reconnect? I don't see any way for them to halt further attempts.

Hectate (changes requested):

Hectate:
Side note since the conversation was on Discord; we will need to open issues or PRs to address reconnection functions in `matchmaking.store`, `lobby.store`, `party.store`, and `me.store` (possibly elsewhere, will need review) after this.

MM changes to reference: https://github.com/beyond-all-reason/tachyon/pull/150

burnhamrobertp:
The various small additions made since your first review were identified in the broader authentication cleanup, and we easier / more appropriate to make here rather than there.

burnhamrobertp:
@Hectate I believe that accounts for all the commentary in the issue, that I had missed, and this should be reviewable now

Hectate:
Z-index issue. If the middle one is selected, it's actually entirely blocking the pop-up to the point that it can't be clicked.
image

src/renderer/views/profile/party.vue
Hectate, line 396: Whatever value we end up using for the z-indez in `matchmaking.vue` we need to use here also FYI

src/renderer/views/play/matchmaking.vue
Hectate, line 246: The likely source of the z-index issue.

Hectate (changes requested):
Thanks. I'm good with this once we fix the merge conflicts and z-index issue(s).

src/renderer/views/profile/party.vue
burnhamrobertp: The underlying issue was that the views weren't operating in their own isolated stacking context; the change I made out at the layout layer will fix it unilaterally, everywhere.

Hectate (approved):
Made one small tweak, but all good now. Thanks!

PR #658, opened 2026-07-27
Handle shell open failures instead of crashing the client

### LINKED ISSUE #507: Error invoking remote method shell:openReplaysDir: reply was never sent
I'm getting the `Error: Error invoking remote method shell:openReplaysDir: reply was never sent`  after clicking on the `Open Replay Directory` button and running the command seen in the screenshot.
I assume it is related to me not having a filemanager installed

Image

burnhamrobertp:
Closes #507. Shell failures now come back as data and show a dismissible message, instead of taking the client down with the fatal error modal.

The try/catch around each handler is what actually fixes this, not the tidier return type. Anything thrown in main reaches the renderer as an unhandled rejection, and Error.vue turns those into a modal you can only quit or reload out of.

The result type is new because #588's can't be reused - that one is generated by the tachyon package and keyed on its command ids. This mirrors its wording so there's one way of describing an IPC failure rather than two.

All seven handlers rather than just the replay one from the report, since it's the same bug in each. "Show in folder" also checks the file exists first, because the underlying call reports nothing at all - a missing file manager still isn't detectable there.

Renderer calls all go through `shellApi` rather than `window.shell`, with a lint rule holding that in place. The first version of this PR already missed a call in NewsTile.vue, so leaving it to memory clearly wasn't going to work.

The modal still catches anything else unhandled; that's #509.

Not verified by hand - the reporting environment is Linux without a file manager.

### DIFF
--- a/src/renderer/api/shell.ts
+++ b/src/renderer/api/shell.ts
@@ -11,9 +11,9 @@ const i18n = setupI18n();
 // Failing to open a folder or a link isn't worth interrupting anyone over, but it shouldn't be
 // swallowed either - the OS reason is the only clue a player has about what went wrong.
 //
-// Callers fire this without awaiting, so it has to absorb its own rejections too. Anything escaping
+// Callers fire these without awaiting, so rejections have to be absorbed here too. Anything escaping
 // reaches Error.vue's unhandledrejection listener, which is an unrecoverable modal.
-export async function openAndReport(open: () => Promise<IpcResult>) {
+async function openAndReport(open: () => Promise<IpcResult>) {
     const result = await open().catch((err): IpcResult => {
         console.error("Shell request failed", err);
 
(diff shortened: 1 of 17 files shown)


src/renderer/api/shell.ts
```diff
@@ -0,0 +1,28 @@
...
+import { setupI18n } from "@renderer/i18n";
+
+const i18n = setupI18n();
+
+// Failing to open a folder or a link isn't worth interrupting anyone over, but it shouldn't be
+// swallowed either - the OS reason is the only clue a player has about what went wrong.
+//
+// Callers fire this without awaiting, so it has to absorb its own rejections too. Anything escaping
+// reaches Error.vue's unhandledrejection listener, which is an unrecoverable modal.
+export async function openAndReport(open: () => Promise<IpcResult>) {
```
Hectate: Realistically, instead of having devs remember to wrap this around every `window.shell`, we should just make this API the sole source of those calls. Then, we pull in this API exported from here instead of using `window.shell...`. That will make sure that it's used correctly every time.

src/renderer/assets/languages/en.json
```diff
@@ -1832,6 +1832,9 @@
     },
     "lobby": {
         "api": {
+            "shell": {
```
Hectate, line 1835: Edits should be in `lobby.json`, not here.

src/renderer/components/misc/DebugSidebar.vue
```diff
@@ -95,19 +96,19 @@ async function onRouteSelect(newRoute: string) {
 }
 
 function openSettings() {
-    window.shell.openSettingsFile();
+    openAndReport(() => window.shell.openSettingsFile());
```
Hectate: Nit; this could be written simpler as 
```ts
openAndReport(window.shell.openSettingsFile);
```
and similar across the various calls.

Hectate (changes requested):

src/renderer/components/misc/DebugSidebar.vue
burnhamrobertp: The restructure took these call sites out, they're `shellApi.openSettingsFile()` now, so there's no wrapper left to make point-free. The zero-arg ones do pass the reference straight through inside the api module. `showReplayInFolder` and `openInBrowser` still need an arrow since they take arguments.

Hectate (approved):

PR #661, opened 2026-07-29
Expose user reporting in the UI

### LINKED ISSUE #582: Expose user reporting in the UI
Users should be able to right-click on a user nearly anywhere in the client where a user is shown, and from a context menu (among other options) select "Report User".
This would then pop up a modal with the necessary fields to be filled out.

burnhamrobertp:
Adds a report flow to the client: a right-click entry on battle participants, plus a button on friends, party members and profiles, all opening one shared modal.

Closes #582.

The reasons and fields aren't invented - chobby only links out to the website, so the website's own form was the reference. It stores a type and a sub type per report and caps the description at 255, which is where the two columns of reasons and the length limit come from. Tachyon only carries a single reason string and has nowhere to put a match, so the two halves are joined into `chat/spam` and the replay link goes in the message text. Raised that with the protocol folks in beyond-all-reason/tachyon#148, since the server side of `user/report` isn't written yet and whoever writes it should know we're guessing.

Rather than telling people to go find a replay link themselves, the modal lists the reported player's recent games from the public replay site and puts the link in for them. That lookup runs in main because the renderer isn't allowed to call out. Worth knowing it only finds games they played in, not ones they only watched, so the "I cannot find the match" path still matters.

Ignore, avoid and block are there but disabled, with tooltips saying so. They're on the website page and worth having, but there's no command for any of them yet - same issue as above.

The reasons are one step rather than the website's two, and the modal keeps a fixed size so stepping through doesn't resize it.

One change reaches outside the report feature: `Modal` used to treat any enter keypress as a submit, which meant enter on the back button sent the report from whichever step you were on. It now leaves that to the form, and the send button is marked `type="submit"`. Nothing else listens for the modal's submit event today, so the only knock-on is that a future modal wanting enter-to-submit has to mark its own button.


Hectate:
FYI; My long term goal was to replace every instance of a username with an actual `User` Component (name TBD) that had this behavior tied into it, along with styling options like flag, chev, etc. That way anywhere we display a user (both currently and in future places) we automatically get the entirety of this behavior "for free". Component (or composables like yours here) are great for letting is set the behavior in one place and maintain it in one place.

Not to say that this PR is wrong, but I wanted you to be aware of it since I realize now it's not spelled out anywhere.

burnhamrobertp:
I think that would be useful for driving the same contextual behavior everywhere, although I'm not sure you'll be able to get away from a component entirely (though the implementation of the component will change after a central implementation like `User` exists (as an aside I might suggest `Player` instead of `User` though perhaps the differentiation is a nit)) since I'd expect the report UX would need to be called out distinctly at least on _certain_ places in the UI?

Hectate:
Nit-wise I see it as "All players are users but not all users are players, some may be spectators, and some may be neither."

But yes, outside of user-name context menus, there will be a need for ways to report things. A basic example is "No player names are on my screen to report because I've gone back to the main menu, but I still want to report a player in the last match I played."

burnhamrobertp:
I'm not 100% certain I agree with you? Though I don't disagree with you that it _could_ be useful. Generally I'd think you would want to carefully balance the need to have interacted with people in order to report them against players quickly playing a match and then leaving to make it harder to report them.

Hectate:
The mod team will determine the requirements for considering a report valid, not us.

burnhamrobertp:
> The mod team will determine the requirements for considering a report valid, not us.

The UX by which reporting is accessed is not the same as the content of the report. The former is what I'm referring to. The latter is what I would describe what you're referring to.

That being said, as far as the availability of the ability to report other players; I'm not sure why BAR seems so inclined to buck the trend on what literally every other game does (this is not the first instance I've encountered, of the sentiment)

Ultimately; I have no strong preference; if they really want a hyper-readily available report button 🤷

src/renderer/components/user/ReportUserModal.vue
```diff
@@ -0,0 +1,514 @@
...
+
+const stepTitleKey = computed(() => stepTitleKeys[stage.value]);
+
+const matchTeams = computed(() => Map.groupBy(matchDetails.value?.players ?? [], (player) => player.allyTeamId));
+
+const reportedUserSpectated = computed(() =>
+    matchDetails.value?.spectators.some((spectator) => spectator.userId?.toString() === reportedUser.value?.userId)
+);
+
+const messageSuffix = computed(() => (selectedMatch.value ? `\nReplay: https://bar-rts.com/replays/${selectedMatch.value.id}` : ""));
```
Hectate: Can we use the constant here instead in some fashion? Not only will it help in general, the Config Service has this as a customizable value so it'll need to match whatever the client configures it as.

tests/unit/main/online-replays.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+});
+
+describe("searchOnlineReplaysByPlayer", () => {
+    it("queries the replay API for that player and summarises the matches", async () => {
+        const fetchMock = mockFetch({ ok: true, body: { data: [apiReplay] } });
+
+        const result = await searchOnlineReplaysByPlayer("Naughty", 10);
+
+        const [url] = fetchMock.mock.calls[0];
+        expect(url).toContain("https://api.bar-rts.com/replays?");
```
Hectate: Nit: Test will break _if_ the config service provides a different URL in the future. Not a big deal in a dev environment but worth fixing now by pulling in the constant value?

src/main/content/replays/online-replays.ts
```diff
@@ -0,0 +1,120 @@
+// SPDX-FileCopyrightText: 2026 The BAR Lobby Authors
+//
+// SPDX-License-Identifier: MIT
+
+import { IpcResult } from "@main/typed-ipc";
+import { logger } from "@main/utils/logger";
+
+const log = logger("online-replays.ts");
+
+const ONLINE_REPLAYS_API_URL = "https://api.bar-rts.com/replays";
```
Hectate: This value should probably be stored adjacent to `REPLAY_SERVICE_URL` in `src\main\services\shell.service.ts`.
I can add it to the config service PR later that way and everything else can pull it from there as an export.

lang/en/lobby.json
```diff
@@ -582,6 +582,65 @@
...
+                    },
+                    "reasons": {
+                        "chat": {
+                            "spam": "Spam",
+                            "bullying": "Bullying",
+                            "hate": "Hate speech",
+                            "other": "Other"
+                        },
+                        "actions": {
+                            "noob": "Noob",
```
Hectate, line 629: Let's pull "Noob" out as a reason to report, along with the associated UI element. I don't think it would survive long-term as a valid reason, so there's no reason to add it now.

Hectate (changes requested):
Looks good for the most part. Just a couple adjustments. We may have to revisit after the server implements, but I expect additional features requested related to reporting users anyway.

For future readers; testing of the replay fetch worked by forcing the username to be one that was known to have played recent matches on the live servers. Since Tachyon users are not currently generating replays, you are not expected to locate replays except by ensuring that the Tachyon username exactly matches a live (e.g. Chobby) username.

Hectate:
Images from testing for posterity. Note that the username visible at the top of the modal does _not_ match the username searched for in replay details due to testing methodology.

image
image

lang/en/lobby.json
burnhamrobertp: The ux for this was modeled after what's available right now for reporting players 😂 though I don't disagree

Hectate: Oh boy. I didn't realize.

Ok well then we leave it.

Maybe the mod team keeps it around to avoid people using "Griefing" to report poor play, while "Noob" gets less mod attention.

src/main/content/replays/online-replays.ts
burnhamrobertp: This is on the rendering side, but shell.service.ts is on the electron side; it cannot be imported here.

burnhamrobertp:
@Hectate pulled those URLs out into constants, ready for another look.

src/main/content/replays/online-replays.ts
burnhamrobertp: Went with an exported const at the top of this file instead of moving it into shell.service.ts. The modal needs the replay site URL as well, and pulling shell.service into the renderer would drag electron along with it, which I do not think survives the bundle. Since you are picking both of these up in #597 anyway they should be easy enough to find.

Hectate (approved):

Issue #385, opened 2025-05-25
Unhandled Promise Rejection Warnings upon receiving 400 Bad Request while renewing the auth token


Hectate:
Lobby was idle in the background while I was doing other stuff and I came back to the following error in the terminal output. I believe I had switched my login from "Hectwo" to "Hecthree" on the lobby server in Firefox, but not in bar-lobby itself. I'm unsure if that was relevant.

```ts
[22:13:49.723] ERROR: main - 2025-05-25T01:39:05.398Z - oauth2-utils - Failed to renew token, wiping: 400 Bad Request: {"error_description":"invalid request","error":"invalid_request"}
(node:16068) UnhandledPromiseRejectionWarning: Error: Failed to renew token, wiping: 400 Bad Request
    at Timeout.renewAccessToken (D:\git\bar-lobby\src\main\oauth2\oauth2.ts:150:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
(node:16068) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
[22:13:49.724] ERROR: main - 2025-05-25T01:39:05.398Z - oauth2-utils - 400 Bad request: https://lobby-server-dev.beyondallreason.dev/oauth/token?grant_type=refresh_token&client_id=generic_lobby&scope=tachyon.lobby&refresh_token=BFNL4NK0LQQDIHKLFB9B1KE4FQIHIA9KMNAAHDBQDKFT3404C68G
```

Slightly further down, I also had this.
```ts
[22:28:49.313] ERROR: main - 2025-05-25T01:39:05.398Z - oauth2-utils - No refresh token available
(node:16068) UnhandledPromiseRejectionWarning: Error: No refresh token available
    at Timeout.renewAccessToken (D:\git\bar-lobby\src\main\oauth2\oauth2.ts:131:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
[...]

PR #662, opened 2026-07-30
Give the main process one owner for session state

burnhamrobertp:
Auth state was spread across a few files and the renderer kept its own copy of whether you were logged in, so a refresh failing in the background left the UI still saying you were online. One place tracks it now and the renderer reads from that. A temporary server hiccup no longer signs you out and sends you back through the browser, and sign-in details no longer reach the log files, which is how #385 turned up in the first place.

Logout means logout now. It destroys the stored account rather than quietly keeping the refresh token, which is what "Change account" existed to work around, so that's gone too. Signing in on launch is still its own setting and logging out leaves it alone.

Being signed in without being connected is a normal thing to be, so going online and going offline are things you can do on purpose, from the server status. Going offline asks first because it drops you out of anything you're in. Going online doesn't, since there's nothing to lose by it.

Your name and id sit next to the credentials now rather than in the renderer's browser-side storage. That storage is a cache as far as the browser is concerned, so it can be thrown out when disk gets tight, it's one click to clear from devtools, and a change to the database layout can drop it, while the credentials it describes are in a plain file that survives all three. When it went you were left signed in with no name to show.

Most of the size is collecting auth logic that was scattered across main, the renderer and a timer in the oauth code into one place.

Rebased onto #657 and still draft until that lands, so the diff includes it.

### DIFF
--- a/src/renderer/store/me.store.ts
+++ b/src/renderer/store/me.store.ts
@@ -89,6 +89,11 @@ async function serverChanged() {
     await logout();
 }
 
+// Main persists the identity off this same event and that is what gets read back
+// on startup, but db.users is still where anything looking up a user by id goes,
+// the profile view included, so our own record has to be in there too.
+// TODO tidy this up: the row is a snapshot of the whole reactive object, written
+// in two steps that aren't in one transaction, and nothing reads isMe any more.
 window.tachyon.onEvent("user/self", async (event) => {
     console.debug(`Received user/self event: ${JSON.stringify(event)}`);
     if (event && event.user) {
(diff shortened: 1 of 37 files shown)


src/renderer/store/me.store.ts
```diff
@@ -249,29 +262,44 @@ export async function initMeStore() {
...
 
+    await syncAuthState();
+    await restorePreviousSession();
+
     me.isInitialized = true;
 }
+
+// Multiplayer is behind devMode, so outside it there is no session to restore.
+async function restorePreviousSession() {
+    if (!settingsStore.devMode) return;
```
Hectate: I feel like this is going to bite us later when we forget about it when removing the dev-mode-only flag from multiplayer features. Let's at least put the return inside a code block that also includes logging something like "restorePreviousSession returned early because devMode is not enabled".

Hectate:
Apologies. I was testing and I accidentally pushed the local merge commits that I did not intend to.

Hectate (approved):
Eventually, we will start unlocking multiplayer features that are currently tied to Dev Mode. Once that happens, we'll need to make the following adjustments
1. Get rid of the `devMode` check in `me.store`
2. Send the user back to the main menu if they log out while in a multiplayer screen
3. If the user is offline and they navigate to an online screen, offer to log them in.

But this is good for now. Thanks!

Issue #491, opened 2025-10-06
Download game needs to check for free-space available on destination volume


WybrenKoelmans:
If the user doesn't not have enough disk space on the destination volume, bar-lobby should not attempt to install the game. Instead it should warn the user to clear up more disk-space and exit.

Currently the game will soft-lock and leave the user with ~5MiB of diskspace left which isn't always easy to recover from.

It might even be good to ask if the user wants to install the game and a size estimate.

Hectate:
Now that we have the ability to define the asset path, the plan should be to use Node's `fs.statfs()` to do a quick calculation on the targeted drive. This should be done during the initial setup, and any time asset files are being moved. We can warn the user if the available space is below a certain number, or even prevent entire actions based on an additional lower limit.

Issue #197, opened 2023-04-10
Content APIs should remove old versions


Jazcash:
Currently, there is no way to remove downloaded content other than deleting it from the filesystem. However, the map and game APIs don't handle this well because they store references to which content is installed in the db cache, and I don't believe they validate those entries against the filesystem to make sure they still exist.

First we need to make sure all content APIs stay synced with the actual files in the filesystem. Deleting a file from the cache should also delete it from the filesystem and vice-versa.

As for cleaning up old content, I suggest we store lastLaunched date for every version of engine/game/maps, which gets updated whenever the game API launched that content. Then we can set some configurable expiration time (e.g. 1 year), and periodically delete all content that hasn't been launched within that time (not replays ofc!). Maybe we don't want to do that for maps, but at the least we should delete old versions of maps when newer versions of them are installed, but should also consider maps downloaded to watch old replays.

Jazcash:
Done for old engine versions, but can't do same for game versions until https://github.com/beyond-all-reason/pr-downloader/issues/21 is implemented

burnhamrobertp:
I've got a mechanism for this working on a branch, but deliberately left switched off, because what should actually be kept feels like a decision to make out loud rather than bury in a default.

What it does when something asks it to sweep - installed content is kept if:

- something currently claims it
- it's pinned
- it was last used within the last 90 days
- it's a local `.sdd` game
- removing it would leave no engine installed at all

Content with no recorded last-used date gets stamped and kept rather than removed, so the first sweep on an existing install can't wipe someone's library. Pinning works but isn't exposed anywhere yet, so there's no way for a user to say "keep this" today.

Two things I don't think I should be deciding on my own.

Which things get to claim content. The obvious candidates are the versions a replay needs, whatever the client treats as its default engine and game, and anything a running game is using. I've written those as examples but left them commented out, since each one is really a claim about what a user expects to still be there later.

And when it runs. On startup, on a timer, or only when someone asks in a settings screen. Running it automatically is the point of the issue, but it also means content disappearing without anyone asking for that, so it seems worth choosing deliberately.

Retention is 90 days and not a user setting, on the grounds that it isn't a number anyone can reason about and getting it wrong deletes their stuff.

PR #675, opened 2026-08-04
Own engine, game and map content behind one API

burnhamrobertp:
Engine, game and map downloads each had their own API, their own IPC channels and their own progress events, and callers reached into whichever one they needed. One owner now sits in front of all three, with `ensure` and `remove` taking a list of refs. It is a single PR because the old per-type path cannot be deleted until the renderer is already reading the new change stream.

How much runs at once cannot be decided per content type - four maps plus two engines plus a game is not a limit anyone chose - so the cap sits with the queue, at the 2-4 asked for in #423. Only one pr-downloader runs at a time: it rewrites rapid's repo index on every invocation, maps included, and two of them racing on it kills one. Concurrency for its content comes from handing a single invocation the whole set instead, which is what upstream suggests and what it parallelises internally, each asset still costing a slot so the limit means the same thing. It abandons the whole invocation when one asset cannot be resolved, so a failed batch is retried one at a time rather than letting one bad name take the rest down.

That leaves progress covering the set rather than each asset, so rows downloading together show the same figures; splitting them needs per-asset sizes pr-downloader does not report.

The sweep for #197 is here and deliberately disconnected: it works, nothing calls it, and the claim sources that would decide what is worth keeping are commented out, because that policy has not been agreed. The free-space refusal for #491 is wired live. TypeScript cannot scope an export to a directory, so an eslint rule is what keeps anything outside the content layer from reaching a provider directly.

Ran first-run setup against an empty assets directory: engine, pool archive, game and both default maps landed, the two maps going out as a single pr-downloader invocation.


Hectate:
I've gotten through 85 of the file changes, so mainly just the entirely new files to look at now. Not a formal review but I noted that somehow changes to `lang/en/lobby.json` file are missing from this PR even though the keys have been added to the languages. Make sure we have corresponding changes committed there, otherwise we will lose these keys the next time we run `generate-i18n-assets`.

Hectate:
When downloading multiple assets, for example from the matchmaking view, the progress inside the DownloadContentButton does not show the correct percentage complete. For example, with downloading the entire queue's maps, the percentage would start at left (zero) and then move small amount rightward before resetting back to zero. The reset happens when a batch completes and a new set of 4 maps from the queue are started. I suspect it's just a math error on the overall percentage displayed.

image

Hectate (approved):

PR #676, opened 2026-08-05
Refactor Tachyon server-to-client request handlers

Hectate:
Refactors Tachyon request handlers (server-to-client) for cleaner, type-safe code.
* Structured in a separate file where new requests should be added in the future
* Unhandled requests will log the problem and respond with failure using `command_unimplemented` or `internal_error`
* `matchmaking/checkAssets` is newly implemented for the client

### AI / LLM usage statement:
GitHub Copilot was used to generate the code and tests for this PR. This was done in small, directed batches to ensure minimal LLM bloat and close adherence to desired implementation.

### DIFF
--- a/src/main/tachyon/tachyon-client.ts
+++ b/src/main/tachyon/tachyon-client.ts
@@ -23,8 +23,6 @@ export type TachyonClientRequestHandlers = {
     [CommandId in ServerToUserRequestCommandId]: (data: GetCommandData<ServerToUserRequest<CommandId>>) => Promise<UserToServerResponseBody<CommandId>>;
 };
 
-export type PartialTachyonClientRequestHandlers = Partial<TachyonClientRequestHandlers>;
-
 class InternalError extends Error {
     constructor(message: string) {
         super(message);
@@ -48,10 +46,10 @@ export class TachyonClient {
     public onSocketClose: Signal<void> = new Signal();
     public onEvent: Signal<TachyonEvent> = new Signal();
 
-    private requestHandlers: PartialTachyonClientRequestHandlers;
+    private requestHandlers: TachyonClientRequestHandlers;
     private responseHandlers: Map<string, (response: TachyonResponse | { status: "socket_closed" }) => void> = new Map();
 
-    constructor(requestHandlers: PartialTachyonClientRequestHandlers) {
+    constructor(requestHandlers: TachyonClientRequestHandlers) {
         this.requestHandlers = requestHandlers;
     }
 
--- a/src/main/tachyon/tachyon-client.ts
+++ b/src/main/tachyon/tachyon-client.ts
@@ -14,11 +14,16 @@ import { MessageEvent, WebSocket } from "ws";
 
 const log = logger("tachyon-client");
 
+type ServerToUserRequestCommandId = GetCommandIds<"server", "user", "request">;
+type ServerToUserRequest<C extends ServerToUserRequestCommandId = ServerToUserRequestCommandId> = GetCommands<"server", "user", "request", C>;
+type UserToServerResponseBody<C extends ServerToUserRequestCommandId = ServerToUserRequestCommandId> = Omit<GetCommands<"user", "server", "response", C>, "type" | "commandId" | "messageId">;
+type AnyUserToServerResponseBody = UserToServerResponseBody<ServerToUserRequestCommandId>;
+
 export type TachyonClientRequestHandlers = {
-    [CommandId in GetCommandIds<"server", "user", "request">]: (
-        data: GetCommandData<GetCommands<"server", "user", "request", CommandId>>
-    ) => Promise<Omit<GetCommands<"user", "server", "response", CommandId>, "type" | "commandId" | "messageId">>;
+    [CommandId in ServerToUserRequestCommandId]: (data: GetCommandData<ServerToUserRequest<CommandId>>) => Promise<UserToServerResponseBody<CommandId>>;
 };
+
+export type PartialTachyonClientRequestHandlers = Partial<TachyonClientRequestHandlers>;
 
 export class TachyonClient {
     public socket?: WebSocket;
@@ -27,10 +32,10 @@ export class TachyonClient {
     public onSocketClose: Signal<void> = new Signal();
     public onEvent: Signal<TachyonEvent> = new Signal();
 
-    private requestHandlers: TachyonClientRequestHandlers;
+    private requestHandlers: PartialTachyonClientRequestHandlers;
     private responseHandlers: Map<string, (response: TachyonResponse | { status: "socket_closed" }) => void> = new Map();
 
-    constructor(requestHandlers: TachyonClientRequestHandlers) {
+    constructor(requestHandlers: PartialTachyonClientRequestHandlers) {
         this.requestHandlers = requestHandlers;
     }
 
@@ -178,7 +183,7 @@ export class TachyonClient {
             throw new Error(`Message does not match expected command structure`);
         }
         if (obj.type === "request") {
-            this.handleRequest(obj);
+            this.handleRequest(obj as ServerToUserRequest);
         } else if (obj.type === "response") {
             this.handleResponse(obj);
         } else if (obj.type === "event") {
@@ -189,16 +194,25 @@ export class TachyonClient {
     }
 
     private async handleRequest(command: TachyonRequest) {
-        const handler = this.requestHandlers[command.commandId as keyof typeof this.requestHandlers] as unknown as (
-            data?: unknown
-        ) => Promise<Omit<TachyonResponse, "type" | "commandId" | "messageId">>;
+        const commandId = command.commandId as ServerToUserRequestCommandId;
+        const handler = this.requestHandlers[commandId] as ((data?: unknown) => Promise<AnyUserToServerResponseBody>) | undefined;
+        const requestData = "data" in command ? command.data : undefined;
+
+        let handlerResponse: AnyUserToServerResponseBody;
         if (!handler) {
-            throw new Error(`No response handler found for: ${command.commandId}`);
+            log.warn(`No response handler found for: ${commandId}`);
+            handlerResponse = {
+                status: "failed",
+                reason: "command_unimplemented",
+                details: `No response handler found for: ${commandId}`,
+            } as AnyUserToServerResponseBody;
+        } else {
+            handlerResponse = await handler(requestData);
         }
-        const handlerResponse = await handler("data" in command ? command.data : undefined);
+
         const response = {
             type: "response",
-            commandId: command.commandId,
+            commandId,
             messageId: command.messageId,
             ...handlerResponse,
         } as TachyonResponse;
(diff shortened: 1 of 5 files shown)


src/main/tachyon/tachyon-client.ts
```diff
@@ -189,16 +194,25 @@ export class TachyonClient {
...
     private async handleRequest(command: TachyonRequest) {
-        const handler = this.requestHandlers[command.commandId as keyof typeof this.requestHandlers] as unknown as (
-            data?: unknown
-        ) => Promise<Omit<TachyonResponse, "type" | "commandId" | "messageId">>;
+        const commandId = command.commandId as ServerToUserRequestCommandId;
+        const handler = this.requestHandlers[commandId] as ((data?: unknown) => Promise<AnyUserToServerResponseBody>) | undefined;
+        const requestData = "data" in command ? command.data : undefined;
+
+        let handlerResponse: AnyUserToServerResponseBody;
         if (!handler) {
```
burnhamrobertp, line 226: Is it a problem that this only covers commands that exist in the protocol build we ship? `isCommand` just checks the three fields are present, so an unknown commandId gets here, we build the `command_unimplemented` body, and then `validateCommand` below throws `No validator found with id: ...`. `handleRequest` isn't awaited in `handleMessage`, so the try/catch there wouldn't catch it - looks like an unhandled rejection with nothing sent back, which seems like precisely the class of issue we'd expect this to prevent.

src/main/tachyon/tachyon-client.ts
```diff
@@ -14,23 +14,28 @@ import { MessageEvent, WebSocket } from "ws";
...
+type AnyUserToServerResponseBody = UserToServerResponseBody<ServerToUserRequestCommandId>;
+
 export type TachyonClientRequestHandlers = {
-    [CommandId in GetCommandIds<"server", "user", "request">]: (
-        data: GetCommandData<GetCommands<"server", "user", "request", CommandId>>
-    ) => Promise<Omit<GetCommands<"user", "server", "response", CommandId>, "type" | "commandId" | "messageId">>;
+    [CommandId in ServerToUserRequestCommandId]: (data: GetCommandData<ServerToUserRequest<CommandId>>) => Promise<UserToServerResponseBody<CommandId>>;
 };
 
+export type PartialTachyonClientRequestHandlers = Partial<TachyonClientRequestHandlers>;
```
burnhamrobertp: Going partial gives up the build error that flagged checkAssets in the first place - I compiled the old non-partial type against 1.23.0 and tsc points straight at the missing handler, whereas now a new server-to-user request just compiles and we only hear about it at runtime.

src/main/tachyon/tachyon.handlers.ts
```diff
@@ -0,0 +1,80 @@
...
+import { engineContentAPI } from "@main/content/engine/engine-content";
+
+const log = logger("tachyon-handlers");
+
+type ServerToUserRequestId = GetCommandIds<"server", "user", "request">;
+type RequestData<C extends ServerToUserRequestId> = GetCommandData<GetCommands<"server", "user", "request", C>>;
+type ResponseBody<C extends ServerToUserRequestId> = Omit<GetCommands<"user", "server", "response", C>, "type" | "commandId" | "messageId">;
+type Exact<Expected, Actual extends Expected> = Expected & Record<Exclude<keyof Actual, keyof Expected>, never>;
+
+const createTypedTachyonRequestHandler =
```
burnhamrobertp, line 21: I don't think `Exact` does anything - `R` only appears inside `keyof Actual`, so TS falls back to `ResponseBody<C>` and it collapses to `Exact<X, X>`. I put a `bogusExcessProperty: 123` in the battle/start return and `npm run typecheck:node` was clean, and that would have gone out over the wire. Between that and the `as unknown as Pick<...>`, does this beat a plain literal with `satisfies`?

src/main/tachyon/tachyon.handlers.ts
```diff
@@ -0,0 +1,80 @@
...
+    } satisfies PartialTachyonClientRequestHandlers;
+}
+
+function createMatchmakingHandlers() {
+    return {
+        ...defineTachyonRequestHandler(
+            "matchmaking/checkAssets",
+            createTypedTachyonRequestHandler<"matchmaking/checkAssets">()(async (data: MatchmakingCheckAssetsRequestData) => {
+                log.info(`Received matchmaking check assets request: ${JSON.stringify(data)}`);
+                const itemsRequired =
```
burnhamrobertp, line 72: The docs say a `missing` reply from any party member fails the leader's queue request. Nothing here starts a download or tells the renderer, so wouldn't the member be a silent blocker? Also `setRequiredAssetsArrays` treats a map with no db row as fine while this calls it missing, so their own screen could disagree.

burnhamrobertp (changes requested):
Nice cleanup - the separate handlers file reads much better. A few things I poked at below.

src/main/tachyon/tachyon.handlers.ts
Hectate: Not "leader" but the member that sends the `matchmaking/queue` request will see a `failed` response after a `missing` response or if any member times out of their response window. We can use the same error handling as done with `version_mismatch` to handle a `party_missing_asset` response. E.G.
```ts
if (isTachyonErrorForCommand(error, "matchmaking/queue") && error.reason === "party_missing_asset") {
```
and then notify the person that attempted to queue about the failure response.

Since the protocol won't let us know _who_ is not ready (whether one or many), we could also have a notificationAPI pop up to advise the user that a queue request was refused and for which queueId. It would still be the responsibility of the party to communicate about readiness. I expect that they would have done that before clicking a button (assuming a reasonably well-acquainted party, anyway), while `checkAssets` is a stopgap against failure to load into a match for asset reasons.

I'll add the appropriate notifications.

Auto-downloading for the user is not an option; that will be reserved for direct user input only.

src/main/tachyon/tachyon-client.ts
Hectate: I guess the question here is if we want a build error when updating a Tachyon version that includes a change to the requests. With this we could implement new requests whenever time permitted, and not end with a requirement to implement even a basic "unimplemented" response into a version bump.

I'm not against doing so, there is value to both ways. It's certainly not difficult to add a new request as "unimplemented" with this new structure.

I'm also not opposed to a Partial approach either. Given that we're building the protocol in parallel with the client and server, the team involved is usually well aware of new, unimplemented features.

Do you have a strong preference?

src/main/tachyon/tachyon-client.ts
Hectate: Any time it's a _known_ command from the protocol, we should use `command_unimplemented`. On the other hand, anything that arrives that we don't recognize at all should be an `invalid_request`.

I'll look at that closer.

src/main/tachyon/tachyon.handlers.ts
Hectate: This kind of Typescript shenanigans is exactly why I wanted more eyes on it. 👀 

The goal was strict type-safety, so having bogus properties possible should be an automatic no-go.

src/main/tachyon/tachyon-client.ts
burnhamrobertp: > if we want a build error when updating a Tachyon version that includes a change to the requests 

Specifically, a change in the tachyon library, right? I.e. I pump the tachyon client to 1.23 but don't implement the corresponding changes in the lobby that were made to the tachyon library?

If that's the case, then yes, I would think that should be a build error (and stubbing functionality in the client should be highly discouraged / disallowed). This ensures that the contracts between client and server can be reliably enforced at build time (at least, in theory, in practice its not ever this tight).

src/main/tachyon/tachyon-client.ts
Hectate: I think this is covered now with the latest commit.

src/main/tachyon/tachyon-client.ts
Hectate: Yes, you have it right.

I think that's a solid case for it. Certainly, a breaking change is going to require work, and we have the ability to determine if there will be issues at the point of connection based on version numbers on the server and client.
Obviously, right now things are very much in flux but that will change when we hit a public release point in "production".

src/renderer/store/matchmaking.store.ts
```diff
@@ -196,8 +192,12 @@ async function sendQueueRequest() {
...
     } catch (error) {
         if (isTachyonErrorForCommand(error, "matchmaking/queue") && error.reason === "version_mismatch") {
-            notificationsApi.alert({ text: "Queue version changed; refreshing list.", severity: "info" });
-            await sendListRequest();
+            if (error.reason === "version_mismatch") {
+                notificationsApi.alert({ text: "Queue version changed; refreshing list.", severity: "info" });
+                await sendListRequest();
+            } else if (error.reason === "party_assets_missing") {
+                notificationsApi.alert({ text: "Party queue rejected, required assets are missing for the queue.", severity: "error" });
+            }
```
burnhamrobertp: This is dead code due to the `&& error.reason === "version_mismatch"` check on 194

src/main/tachyon/tachyon-client.ts
```diff
@@ -189,21 +218,57 @@ export class TachyonClient {
...
         if (!handler) {
-            throw new Error(`No response handler found for: ${command.commandId}`);
+            log.warn(`No response handler found for: ${commandId}`);
+            handlerResponse = {
+                status: "failed",
+                reason: "command_unimplemented",
+                details: `No response handler found for: ${commandId}`,
+            } as AnyUserToServerResponseBody;
+        } else {
+            handlerResponse = await handler(requestData);
```
burnhamrobertp: This is still outside the try/catch, meaning a failure inside the handler execution produces no response to the server and leaves an unhandled exception.

burnhamrobertp (changes requested):

src/renderer/store/matchmaking.store.ts
Hectate: This is what happens when I'm crunching code instead of going to bed like I should have 😪

Hectate:
OK, I think that hit everything.

src/renderer/store/matchmaking.store.ts
```diff
@@ -195,9 +191,13 @@ async function sendQueueRequest() {
...
-        if (isTachyonErrorForCommand(error, "matchmaking/queue") && error.reason === "version_mismatch") {
-            notificationsApi.alert({ text: "Queue version changed; refreshing list.", severity: "info" });
-            await sendListRequest();
+        if (isTachyonErrorForCommand(error, "matchmaking/queue")) {
+            if (error.reason === "version_mismatch") {
+                notificationsApi.alert({ text: "Queue version changed; refreshing list.", severity: "info" });
+                await sendListRequest();
+            } else if (error.reason === "party_missing_asset") {
+                notificationsApi.alert({ text: "Party queue rejected, required assets are missing for the queue.", severity: "error" });
+            }
```
burnhamrobertp, line 203: Tachyon errors besides `version_mismatch` and `party_missing_asset` get eaten silently. I'm not sure what other errors there might be or how severe they are...but probably _something_ is necessary here. Or else restructure this code so that it more cleanly reaches the else below, at line 202, if that's intended to be the generic handling code?

burnhamrobertp (changes requested):

PR #680, opened 2026-08-07
Make switching the active server take effect

burnhamrobertp:
Picking a different active server only wrote the setting - nothing watched it, so the connection stayed on the old one while the modal claimed the change had immediate effect.

Switching ends the session rather than moving the socket over, since that setting picks the authorization server too and the stored tokens came from the server being left.

Turned up while testing #657, there's no issue for it.

### DIFF
--- a/src/renderer/store/me.store.ts
+++ b/src/renderer/store/me.store.ts
@@ -4,9 +4,10 @@
 
 import { Me } from "@main/model/user";
 import { db } from "@renderer/store/db";
-import { reactive, toRaw } from "vue";
+import { reactive, toRaw, watch } from "vue";
 import { tachyonStore } from "@renderer/store/tachyon.store";
 import { PrivateUser } from "tachyon-protocol/types";
+import { settingsStore } from "@renderer/store/settings.store";
 import { subsManager } from "@renderer/store/users.store";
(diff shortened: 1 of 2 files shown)


Hectate:
Switching servers while logged in caused a crash in the client after about 10 seconds.,
```ps
[11:40:45.945] ERROR: main - 2026-08-07T15:40:28.825Z - tachyon-client - No response received for request system/disconnect
[11:40:45.945] INFO: main - 2026-08-07T15:40:28.825Z - tachyon-service - Disconnected from Tachyon server
[11:40:45.945] INFO: main - 2026-08-07T15:40:28.825Z - tachyon-client - Disconnected: 1000
[11:40:45.946] ERROR: main - 2026-08-07T15:40:28.825Z - tachyon-client - Error sending disconnect command: Error: No response received for request system/disconnect, socket closed.
[11:40:55.948] ERROR: renderer - 2026-08-07T15:40:55.947Z - Error.vue - Error: Not authenticated
    at connect (http://localhost:5173/src/renderer/store/tachyon.store.ts:16:34)
```

Hectate (approved):

Hectate (approved):

PR #686, opened 2026-08-12
Party size matchmaking validation

Hectate:
Adds party size awareness to the client for enabling/disabling specific queues. This will prevent requests that would fail because the party was too large for the match type. Parties will be permitted to join any queue with a team size equal to or larger than their current member count.

Note that this needed to happen in both the party view as well as the main matchmaking view, because the server sees any matchmaking request as a party request if the request comes from a user in a party.


src/renderer/views/play/matchmaking.vue
```diff
@@ -53,6 +53,9 @@ SPDX-License-Identifier: MIT
                         >This string should never be visible.</DownloadContentButton
                     >
                 </div>
+                <button v-else-if="partyTooLarge" class="quick-play-button invalid" disabled>
```
burnhamrobertp, line 63: this sits above all the status branches. a member who didn't start the queue still has their own selectedQueue, so if it's a 1v1 they'd be stuck here through Searching -> MatchFound and never get the ready button. same in party.vue.

src/renderer/composables/usePartySizeMatchmaking.ts
```diff
@@ -0,0 +1,45 @@
...
+            /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
+            const { activeParty, queue, members, invited } = newData;
+
+            if (!activeParty || !queue) {
+                partyTooLarge.value = false;
+                return;
+            }
+
+            const party = partyStore.parties.get(activeParty);
+            const playlist = matchmakingStore.playlists.find((p) => p.id === queue);
```
burnhamrobertp: playlists is only read here, not in the source getter above - so this wouldn't re-run when the queue list arrives, right? party view's "Request Queues" button lands right on that.

src/renderer/composables/usePartySizeMatchmaking.ts
```diff
@@ -0,0 +1,45 @@
...
+
+            if (!party || !playlist) {
+                partyTooLarge.value = false;
+                return;
+            }
+
+            partyTooLarge.value = party.members.length > playlist.teamSize;
```
burnhamrobertp, line 37: any reason not to put this in sendQueueRequest next to the downloads check? covers both views and gets checked at click time.

src/renderer/composables/usePartySizeMatchmaking.ts
```diff
@@ -0,0 +1,45 @@
+// SPDX-FileCopyrightText: 2025 The BAR Lobby Authors
+//
+// SPDX-License-Identifier: MIT
+
+import { matchmakingStore } from "@renderer/store/matchmaking.store";
+import { partyStore } from "@renderer/store/party.store";
+import { ref, watch } from "vue";
+
+const partyTooLarge = ref(false);
```
burnhamrobertp: module-level ref but the watch is per-call - both views are under KeepAlive, so after visiting both you'd have two watchers on the one ref. computed instead?

src/renderer/views/profile/party.vue
```diff
@@ -86,6 +86,9 @@ SPDX-License-Identifier: MIT
                                                 >This string should never be visible.</DownloadContentButton
                                             >
                                         </div>
+                                        <button v-else-if="partyTooLarge" class="quick-play-button invalid" disabled>
+                                            {{ t("lobby.views.party.partyTooLarge") }}
```
burnhamrobertp, line 97: every other button in this block uses lobby.multiplayer.ranked.buttons.* - reuse partyTooLarge from there rather than a second key across 9 locale files?

src/renderer/composables/usePartySizeMatchmaking.ts
```diff
@@ -0,0 +1,45 @@
...
+        () => ({
+            activeParty: partyStore.activeParty,
+            queue: matchmakingStore.selectedQueue,
+            // Spreading map entries forces Vue to evaluate the Map's contents
+            members: [...(partyStore.parties.get(partyStore.activeParty ?? "")?.members ?? [])],
+            invited: [...(partyStore.parties.get(partyStore.activeParty ?? "")?.invited ?? [])],
+        }),
+        (newData) => {
+            /* eslint-disable-next-line @typescript-eslint/no-unused-vars */
+            const { activeParty, queue, members, invited } = newData;
```
burnhamrobertp: members/invited never get used - drop them and the eslint-disable?

src/renderer/views/profile/party.vue
```diff
@@ -86,6 +86,9 @@ SPDX-License-Identifier: MIT
                                                 >This string should never be visible.</DownloadContentButton
                                             >
                                         </div>
+                                        <button v-else-if="partyTooLarge" class="quick-play-button invalid" disabled>
```
burnhamrobertp, line 96: the downloads branch above wins over this, so you'd download maps for a queue you can't enter.

src/renderer/views/play/matchmaking.vue
```diff
@@ -287,6 +292,10 @@ function onInfoClick() {
         box-shadow 0.3s ease;
 }
 
+.quick-play-button.invalid {
```
burnhamrobertp, line 308: same rule pasted in party.vue, and both files already have a .disabled class for the not-clickable look.

src/renderer/composables/usePartySizeMatchmaking.ts
```diff
@@ -0,0 +1,45 @@
...
+            const playlist = matchmakingStore.playlists.find((p) => p.id === queue);
+
+            if (!party || !playlist) {
+                partyTooLarge.value = false;
+                return;
+            }
+
+            partyTooLarge.value = party.members.length > playlist.teamSize;
+        },
+        { immediate: true, deep: true }
```
burnhamrobertp: deep looks redundant - the getter returns a fresh object each run so it fires anyway.

src/renderer/views/profile/party.vue
```diff
@@ -290,8 +294,8 @@ const downloadsAreRequiredForSelected = computed(() => {
 
 const downloading = ref(false);
 
-// Switching the active queue during a download cannot be allowed because it messes with state.
 function queueSelected(queue: string) {
+    // Switching the active queue during a download cannot be allowed because it messes with state.
```
burnhamrobertp, line 314: unrelated to the PR? matchmaking.vue still has this one above the function.

src/renderer/composables/usePartySizeMatchmaking.ts
```diff
@@ -0,0 +1,45 @@
...
+//
+// SPDX-License-Identifier: MIT
+
+import { matchmakingStore } from "@renderer/store/matchmaking.store";
+import { partyStore } from "@renderer/store/party.store";
+import { ref, watch } from "vue";
+
+const partyTooLarge = ref(false);
+
+export function usePartySizeMatchmaking() {
```
burnhamrobertp, line 9: worth a spec alongside tests/unit/renderer/store-going-offline.spec.ts?

burnhamrobertp (changes requested):
Rule itself checks out against the server - teiserver rejects when party members > team_size, counting members only and not pending invites, so the comparison and the boundary both line up.

Client side has a couple of snags though. The one at matchmaking.vue:56 is the one I'd want sorted first - looks like a member who didn't start the queue would lose their ready button. And the composable doesn't seem to watch playlists, so the flag would go stale the moment the queue list loads.

src/renderer/views/profile/party.vue
Hectate: Necessary in this setup. Since the button space is shared among all these functions, we don't want to deny the user the option to download the assets. The only way around this would be to flag the queue in some other fashion, which may ultimately be the way the UI team decides to go..

src/renderer/views/play/matchmaking.vue
Hectate: Good call. We will likely need to switch the selectedQueue for the user when they join matchmaking to the joined queue.

src/renderer/composables/usePartySizeMatchmaking.ts
Hectate: I believe you may be right about that. I'll review it.

src/renderer/composables/usePartySizeMatchmaking.ts
Hectate: During my testing I had developed a sense that these were required to be in here for the watcher to work correctly. That said, after finding the content bug that I mentioned in Discord, there could be something else going on. I'll confirm.

src/renderer/composables/usePartySizeMatchmaking.ts
Hectate: I'm not sure I follow what you mean. `queue` is merely an ID (should rename) so we still have to fetch the actual queue details with it; which is what becomes `playlist`.

src/renderer/composables/usePartySizeMatchmaking.ts
Hectate: It's really a matter of the approach. If we were to do the check at the point of the click, we'd be withholding that "status" from the user, absent any other indicators.
From my perspective, if the user has a queue selected that their party won't fit into, the click shouldn't be an option at all (thus this setup that converts the button into a invalid/disabled state).

lang/en/lobby.json
```diff
@@ -97,7 +97,8 @@
                 "declineInvite": "Decline",
                 "kickMember": "Kick Member",
                 "cancelInvite": "Cancel Invite",
-                "noInvites": "No current invitations."
+                "noInvites": "No current invitations.",
```
burnhamrobertp: Also is this trailing `.` consistent with the other potential localizations that might show to players? It seems like the only one in this group that has punctuation

src/renderer/views/play/matchmaking.vue
Hectate: There was a reason for using invalid instead of just disabled, but I can't recall it now. Will check.

src/renderer/composables/usePartySizeMatchmaking.ts
Hectate: Disregard the above reply. You're saying if the playlists change we won't be recalculating whether or not the party would fit in the new queues.
This _should_ trigger if that happens though, since `selectedQueue` will also change. I think for safety, however, I'll add that too.

lang/en/lobby.json
Hectate: This is technically an incomplete sentence while the others are tooltips or buttons. It just happens to be the only one that is meant to be a sentence, and thus the period. That said, it can be revised.

src/renderer/views/profile/party.vue
Hectate: I'd separated them just in case we eventually want to change the text between the two screens. In theory, parties will most likely queue up from the party screen directly. However, individual users may find themselves in parties for non-matchmaking purposes and then get confused why they can't queue up on the "individual matchmaking" view.

It's something to discuss with the UI team on how we want to convey this information, like adding a "you are in a party" identifier to the matchmaking view.

src/renderer/composables/usePartySizeMatchmaking.ts
burnhamrobertp: selectedQueue only gets reassigned when the current pick isn't in the new list - `if (playlists.length > 0 && !hasSelectedQueue)`. so on a plain refresh where 1v1 is still there it stays put and nothing re-fires. think that dep is doing the actual work rather than being a backstop, so probably worth keeping even if it looks redundant later.

src/renderer/views/play/matchmaking.vue
burnhamrobertp: think it's that `.disabled` is `opacity: 0.1` in both files, so the Party Too Large text would come out near invisible. so invalid looks right and that half of my comment was off - the duplicated hex across the two files is the only bit I'd still poke at.

Hectate: Yeah, that was it.

On the topic of CSS, there's a need to do a sweep and unify all of it - but it's not worth the effort until we have the final UI requirements to work against.

PR #705, opened 2026-08-20
Make interface scale a bounded user setting

### LINKED ISSUE #660: Text size follows the window height, so it gets big on tall screens
The new client has always felt to me like looking at an app that's stuck at a high zoom level - everything is bigger than it feels like it should be, a bit like being in an accessibility theme without asking for one. Went looking for why.

The client scales everything by how tall the window is. In `src/main/main-window.ts` the zoom is set to window height divided by 1080, and it's redone every time the window is resized. So the sizes we write in the stylesheets only come out at that size when the window's inner height is exactly 1080 - anything taller and it all gets multiplied. Most screens people play on are taller than that, which would explain the permanent zoomed-in feel.

Headings get the same treatment on top of already being big. In `_text.scss` an h4 is 30px, so on a taller screen it comes out much larger again, which is what made me look.

Some things worth knowing before anyone changes this:

- It only looks at height, so dragging the window shorter shrinks the text even if it's just as wide.
- A maximised 1080p window is actually a bit under 1080 once the frame is taken off, so even the intended case is slightly smaller than what's in the stylesheets.
- None of the sizes are relative to each other, they're all fixed pixel values, so there's no one number to turn.
- `_popper.scss` forces 16px with an important on it, so tooltips wouldn't follow a change made elsewhere.

Not sure whether 1080 is the wrong baseline or whether the headings are just too big for it. Either way it's every view, so it probably wants screenshots at a couple of sizes before picking.

burnhamrobertp:
Fixed #660 by making interface scale a user setting bounded by a supported viewport range, instead of being derived from the window height against a fixed 1080 target.

This deviates from what I understand was the original intent of the new lobby's design constraints, but does so in a way that I have confidence will produce a better result. It wouldn't require more design work, simply different design work, and avoids a lot of the disadvantages associated with the fixed-height-target approach. It matches OS scale level, at least by default, and makes building and maintaining the new lobby more consistent with all other web development practices (a benefit we arguably need, given that collaborators in this repo are most likely to come from this background).


burnhamrobertp:
This is technically finished, but potentially contentious in the approach it takes visually, to the client

Hectate:
Regardless, worth a look. Before we go heavy into UI implementation I'd be best to have this concern sorted out properly one way or another.

burnhamrobertp:
Has some issues both at specific resolutions / scales and with UX involving double-clicking the top header bar and changing the "display" option, which should move the client to that display, but doesn't.

Hectate (approved):

PR #708, opened 2026-08-22
feat(devlog): link dev blog entries to their source

### LINKED ISSUE #702: Dev blog entries are not linked
Currently, the development blog does not offer a way for the user to navigate to the blog entry. Entries that are longer than a few sentences are cut off as well. Entries have no linking back to the source. As a result, it is not possible to read the entire entry in the client, nor to navigate to it. A user would have to know to look for those entries (e.g. on the server blog screen) to read it.

While the exact disposition of the dev blog and overall look and feel of the News view are subject to change, we should be able to easily provide a clickable link to each blog found and pushed to the list. Work expected:
1. Turn the headline or entire preview into a link that can be clicked.
2. If the text runs off the preview, add a "Read more..." indicator
3. Add some visual differentiation between each entry so the headline is not the only separator and the text is not displayed directly on the background image
3. Add the server blog URL to the whitelisted allowed URLs if necessary

Image for current reference

Image

MFA-G:
Closes #702.

### AI / LLM usage statement

This PR was written with AI assistance (Claude, agentic coding). The disclosure was missing from the original description — that was my mistake, and the reviews here are fair.

To be straight with you about verification: the change is covered by unit tests, typecheck, lint and format, but it has **not** been run in the client by a human, so I cannot make the "I tested this in the app" statement the AI policy asks for. @Hectate — that is the honest answer to your question, and I would rather say so than claim a screenshot I do not have. If that is disqualifying under the policy, please close it; the duplicate-key problem it surfaced is worth an issue on its own either way.

### Problem

Dev blog entries in the News view are plain text: no way to navigate to an entry, entries longer than the preview are cut off with no indication, and no link back to the source.

### What the review found, and what changed

**The missing translation string (@Hectate).** `readMore` was added under a *second* `lobby.components.misc.devlogEntry` key. The generator keeps the last of a duplicated key, so the new string was dropped and the entry rendered a raw translation path. The string now lives at `lobby.buttons.readMore` — a generic affordance does not belong to one component, and it cannot collide with the existing `devlogEntry` section there. The duplicate section is gone.

The reason this passed CI *and* passed my own test is the same reason: the test mocked `@renderer/i18n` so `t()` returned its own key, which is true whether or not the string exists. That mock is gone; the test uses the real messages from `tests/setup.renderer.mts` and asserts against the value read from `en.json`, so a dropped string now fails the suite.

**Invalid button content (@burnhamrobertp).** The children are `<span>`s set to `display: block` instead of `<div>`s.

**"Read more" on entries that fit.** It now appears only when the preview actually overflows, measured on the clamped element with a resize observer, rather than whenever a link exists.

**Wrong prop type.** `entry` is `NewsFeedEntry`, not `NewsFeedData`. You were right that it only type-checked because `FeedData` carries a `link` of its own. `DevlogFeed` always passes an entry, so the `undefined` handling (and its test) went with it.

**`split("|")[1]`.** Keeps everything after the leading `author |` prefix, so a body containing a pipe is no longer cut short.

**Redefined defaults and physical properties.** The redeclared button resets are gone (the CSS reset already unsets them) and the remaining box properties are logical (`inline-size`, `margin-block-end`, `border-inline-start`).

**Duplicated `#ffcc00`.** Now `--accent-color` on `:root`, used by both `DevlogEntry` and `NewsTile`. I stopped short of extracting a shared component: `NewsTile` is a hover-reveal image tile and this is a text row in a scroll list, so beyond the accent colour and the read-more idea they have little in common, and folding them together would be a larger change than this issue calls for. The keyboard reachability of news tiles is a real problem but a separate one — happy to open an issue for it.

**Stale test comment.** Removed.

`config.ts` is unchanged from the first push: `https://www.beyondallreason.info/microblogs` is added to `allowedUrlLinks`, without which `shellService.openInBrowser` rejects the URL with `url_not_allowed`.

### Validation

- `npm test` → 47 files, 377 tests passed
- `npm run typecheck` → node and web clean
- `npm run lint`, `npm run format`, `npm run generate-i18n-assets:check` → clean
- Not verified by running the client (see the disclosure above).

### DIFF
--- a/lang/en/lobby.json
+++ b/lang/en/lobby.json
@@ -32,7 +32,8 @@
             }
         },
         "buttons": {
-            "quickPlay": "Quick play"
+            "quickPlay": "Quick play",
+            "readMore": "Read more..."
         },
         "composables": {
             "useTerrainIcon": {
@@ -392,9 +393,6 @@
                 },
                 "devlogFeed": {
                     "latestChanges": "Latest changes"
-                },
-                "devlogEntry": {
-                    "readMore": "Read more..."
                 },
                 "preloader": {
                     "loading": "Loading",
--- a/lang/en/lobby.json
+++ b/lang/en/lobby.json
@@ -393,6 +393,9 @@
                 "devlogFeed": {
                     "latestChanges": "Latest changes"
                 },
+                "devlogEntry": {
+                    "readMore": "Read more..."
+                },
                 "preloader": {
                     "loading": "Loading",
                     "loadingFonts": "Loading fonts",
--- a/src/renderer/components/misc/DevlogEntry.vue
+++ b/src/renderer/components/misc/DevlogEntry.vue
@@ -5,7 +5,13 @@ SPDX-License-Identifier: MIT
 -->
 
 <template>
-    <div>
+    <component
+        :is="entry?.link ? 'button' : 'div'"
+        class="dev-entry"
+        :class="{ clickable: !!entry?.link }"
+        :type="entry?.link ? 'button' : undefined"
+        @click="openEntry"
+    >
         <div class="dev-title">
             {{ title }}
         </div>
@@ -13,19 +19,52 @@ SPDX-License-Identifier: MIT
             {{ formatDistanceToNow(entry.published, { addSuffix: true }) }}
         </div>
         <div class="dev-desc">{{ description }}</div>
-    </div>
+        <div v-if="entry?.link" class="dev-more">{{ t("lobby.components.misc.devlogEntry.readMore") }}</div>
+    </component>
 </template>
 <script lang="ts" setup>
 import { NewsFeedData } from "@main/services/news.service";
 import { formatDistanceToNow } from "date-fns";
 import { computed } from "vue";
+import { useTypedI18n } from "@renderer/i18n";
+import { shellApi } from "@renderer/api/shell";
+
+const { t } = useTypedI18n();
 
 const { entry } = defineProps<{ entry: NewsFeedData | undefined }>();
 
 const title = computed(() => entry?.title?.replace(" ⇀ Microblog ★ Beyond All Reason RTS", ""));
 const description = computed(() => entry?.description?.split("|")[1]?.trim());
+
+const openEntry = () => {
+    if (entry?.link) shellApi.openInBrowser(entry.link);
+};
 </script>
 <style lang="css" scoped>
+.dev-entry {
+    display: block;
+    width: 100%;
+    text-align: left;
+    margin-bottom: 15px;
+    padding: 10px;
+    border: none;
+    border-left: 2px solid rgba(255, 255, 255, 0.15);
+    background: rgba(0, 0, 0, 0.35);
+    color: inherit;
+    font: inherit;
+}
+
+.dev-entry.clickable {
+    cursor: pointer;
+    transition: 0.1s all;
+}
+
+.dev-entry.clickable:hover,
+.dev-entry.clickable:focus-visible {
+    background: rgba(0, 0, 0, 0.6);
+    border-left-color: #ffcc00;
+}
+
 .dev-title {
     font-size: 1.2em;
     font-weight: semibold;
@@ -43,8 +82,17 @@ const description = computed(() => entry?.description?.split("|")[1]?.trim());
     color: rgba(255, 255, 255, 0.8);
     filter: drop-shadow(3px 3px 5px rgba(0, 0, 0, 0.8));
     font-size: 1em;
-    margin-bottom: 15px;
-    overflow-x: hidden;
-    text-overflow: ellipsis;
+    overflow: hidden;
+    display: -webkit-box;
+    -webkit-box-orient: vertical;
+    -webkit-line-clamp: 3;
+    line-clamp: 3;
+}
+
+.dev-more {
+    margin-top: 5px;
+    font-size: 0.8em;
+    font-weight: 600;
+    color: #ffcc00;
 }
 </style>
--- a/src/renderer/components/misc/DevlogEntry.vue
+++ b/src/renderer/components/misc/DevlogEntry.vue
@@ -5,83 +5,84 @@ SPDX-License-Identifier: MIT
 -->
 
 <template>
-    <component
-        :is="entry?.link ? 'button' : 'div'"
-        class="dev-entry"
-        :class="{ clickable: !!entry?.link }"
-        :type="entry?.link ? 'button' : undefined"
-        @click="openEntry"
-    >
-        <div class="dev-title">
-            {{ title }}
-        </div>
-        <div v-if="entry?.published" class="dev-date">
+    <button class="dev-entry" type="button" @click="openEntry">
+        <span class="dev-title">{{ title }}</span>
+        <span v-if="entry.published" class="dev-date">
             {{ formatDistanceToNow(entry.published, { addSuffix: true }) }}
-        </div>
-        <div class="dev-desc">{{ description }}</div>
-        <div v-if="entry?.link" class="dev-more">{{ t("lobby.components.misc.devlogEntry.readMore") }}</div>
-    </component>
+        </span>
+        <span ref="descriptionEl" class="dev-desc">{{ description }}</span>
+        <span v-if="isTruncated" class="dev-more">{{ t("lobby.buttons.readMore") }}</span>
+    </button>
 </template>
 <script lang="ts" setup>
-import { NewsFeedData } from "@main/services/news.service";
+import { NewsFeedEntry } from "@main/services/news.service";
 import { formatDistanceToNow } from "date-fns";
-import { computed } from "vue";
+import { computed, onMounted, ref, useTemplateRef } from "vue";
+import { useResizeObserver } from "@vueuse/core";
 import { useTypedI18n } from "@renderer/i18n";
 import { shellApi } from "@renderer/api/shell";
 
 const { t } = useTypedI18n();
 
-const { entry } = defineProps<{ entry: NewsFeedData | undefined }>();
+const { entry } = defineProps<{ entry: NewsFeedEntry }>();
 
-const title = computed(() => entry?.title?.replace(" ⇀ Microblog ★ Beyond All Reason RTS", ""));
-const description = computed(() => entry?.description?.split("|")[1]?.trim());
+const title = computed(() => entry.title?.replace(" ⇀ Microblog ★ Beyond All Reason RTS", ""));
+// The feed prefixes the body with "<author> | ", and bodies contain pipes of their own.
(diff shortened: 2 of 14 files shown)


lang/en/lobby.json
```diff
@@ -393,6 +393,9 @@
                 "devlogFeed": {
                     "latestChanges": "Latest changes"
                 },
+                "devlogEntry": {
```
burnhamrobertp: Read more can have other users outside of just the dev blog entries. Should be under a more generic key

src/renderer/components/misc/DevlogEntry.vue
```diff
@@ -5,27 +5,66 @@ SPDX-License-Identifier: MIT
...
 
 const title = computed(() => entry?.title?.replace(" ⇀ Microblog ★ Beyond All Reason RTS", ""));
 const description = computed(() => entry?.description?.split("|")[1]?.trim());
+
+const openEntry = () => {
+    if (entry?.link) shellApi.openInBrowser(entry.link);
+};
 </script>
 <style lang="css" scoped>
+.dev-entry {
```
burnhamrobertp, line 49: Many of these look like default styles. Do we need to be redefining default styles?

src/renderer/components/misc/DevlogEntry.vue
```diff
@@ -5,27 +5,66 @@ SPDX-License-Identifier: MIT
...
+    display: block;
+    width: 100%;
+    text-align: left;
+    margin-bottom: 15px;
+    padding: 10px;
+    border: none;
+    border-left: 2px solid rgba(255, 255, 255, 0.15);
```
burnhamrobertp: Logical properties need to be used throughout these styles (and everywhere; physical properties should almost never see any use nowadays)

https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Logical_properties_and_values

lang/en/lobby.json
```diff
@@ -393,6 +393,9 @@
...
+                    "readMore": "Read more..."
+                },
```
Hectate: This key in the JSON is a duplicate entry of line 432 and is resulting in a missing translation string for `readMore` since it is lost by the generation script.

I realize that the checks passed, however that simply means that we have overlooked a case where duplicate keys are inserted, and will need to check for that. In this case, it appears that "last key wins" results in dropping the new string entirely.

Hectate (changes requested):
Thank you for this contribution. Unfortunately, there is one glaring flaw. The new translation string was missing. I have noted the reason in my review.

It was clearly obvious in my testing, and would not survived even a cursory glance by a human author. The Contribution Guidelines, the Pull Request Template, and the AI Policy all indicate that a human must test their PR before submitting it. As a result, I'm concerned that you have either willfully ignored those instructions, or generated what looked like an easy PR via AI and not properly reviewed the result.

I'm willing to accept this PR if you can rectify the issue found by review, state that you have personally tested all of your changes, and disclose if you have or have not used AI assistance and to what extent if so.

lang/en/lobby.json
burnhamrobertp: Also you've duplicated the devlogEntry section, so even if we were leaving this here, it should be added to the existing devlogEntry path

src/renderer/components/misc/DevlogEntry.vue
```diff
@@ -5,27 +5,66 @@ SPDX-License-Identifier: MIT
 -->
 
 <template>
-    <div>
+    <component
```
burnhamrobertp: Making this a button puts a stack of divs inside it, which isn't valid button content and tends to get flattened into one long accessible name - spans set to display block would do the same job.

src/renderer/components/misc/DevlogEntry.vue
```diff
@@ -5,27 +5,66 @@ SPDX-License-Identifier: MIT
...
+    >
         <div class="dev-title">
             {{ title }}
         </div>
         <div v-if="entry?.published" class="dev-date">
             {{ formatDistanceToNow(entry.published, { addSuffix: true }) }}
         </div>
         <div class="dev-desc">{{ description }}</div>
-    </div>
+        <div v-if="entry?.link" class="dev-more">{{ t("lobby.components.misc.devlogEntry.readMore") }}</div>
```
burnhamrobertp: This shows whenever there's a link rather than when the text is actually clamped, so a short entry gets a read more that doesn't lead anywhere new.

src/renderer/components/misc/DevlogEntry.vue
```diff
@@ -5,27 +5,66 @@ SPDX-License-Identifier: MIT
...
 <script lang="ts" setup>
 import { NewsFeedData } from "@main/services/news.service";
 import { formatDistanceToNow } from "date-fns";
 import { computed } from "vue";
+import { useTypedI18n } from "@renderer/i18n";
+import { shellApi } from "@renderer/api/shell";
+
+const { t } = useTypedI18n();
 
 const { entry } = defineProps<{ entry: NewsFeedData | undefined }>();
```
burnhamrobertp: This is typed as the whole feed rather than a single entry - DevlogFeed passes an entry and NewsTile has the right type for the same thing, so entry.link looks like it only type checks because FeedData happens to have a link of its own.

src/renderer/components/misc/DevlogEntry.vue
```diff
@@ -5,27 +5,66 @@ SPDX-License-Identifier: MIT
...
 
 const title = computed(() => entry?.title?.replace(" ⇀ Microblog ★ Beyond All Reason RTS", ""));
 const description = computed(() => entry?.description?.split("|")[1]?.trim());
```
burnhamrobertp: split(|)[1] only keeps the second segment, so a description with another pipe anywhere in it still gets cut short, which is the thing this PR is trying to fix.

src/renderer/components/misc/DevlogEntry.vue
```diff
@@ -5,27 +5,66 @@ SPDX-License-Identifier: MIT
...
+
+const openEntry = () => {
+    if (entry?.link) shellApi.openInBrowser(entry.link);
```
burnhamrobertp: NewsTile already does exactly this, along with the same read more affordance and the same hardcoded #ffcc00 - any reason not to pull the shared parts out rather than keep two copies, especially while news tiles are still not keyboard reachable?

tests/unit/renderer/devlog-entry.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+// SPDX-License-Identifier: MIT
+
+import type { NewsFeedData } from "@main/services/news.service";
+import DevlogEntry from "@renderer/components/misc/DevlogEntry.vue";
+import { mount } from "@vue/test-utils";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const openInBrowser = vi.hoisted(() => vi.fn());
+
+vi.mock("@renderer/i18n", () => ({ useTypedI18n: () => ({ t: (key: string) => key }) }));
```
burnhamrobertp: Mocking the i18n module so t() returns its own key means this passes whether or not the string exists, which looks like how the missing translation got through - the renderer test setup already hands components the real messages.

tests/unit/renderer/devlog-entry.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+function mountEntry(entryProp: NewsFeedData | undefined) {
+    return mount(DevlogEntry, { props: { entry: entryProp } });
+}
+
+describe("DevlogEntry", () => {
+    beforeEach(() => {
+        openInBrowser.mockReset();
+    });
+
+    // The entry was plain text, so there was no way to read the rest of a truncated post.
```
burnhamrobertp: This comment describes what the code used to be rather than anything in the file, so it stops meaning much to anyone reading the test without the PR open.

burnhamrobertp (changes requested):
There are a number of minor corrections, but given the size of the work involved and the use of AI for the work, its far too many corrections for such a small amount of work, and on top of that a broken localization that should've been obvious for anyone that had actually run the code. I'd expect anyone looking to contribute to bereviewing their own changes for correctness, looking through the changes themselves for correctness, simplifications, testing, etc.

In addition, your PR body does not adhere to the AI Policy. Please review it and ensure its written by you and not just the output of the coding agent.

MFA-G:
Thank you both for the detailed review — @Hectate, @burnhamrobertp. Pushed as 96040cd, and I rewrote the PR body.

Answering the direct question first, because it matters more than the code: **yes, this was AI-assisted** (Claude, agentic coding), and I did not disclose it. That was a violation of the policy and I am not going to dress it up. On testing: it is covered by unit tests, typecheck, lint and format, but I have **not** run it in the client, so I cannot make the statement the policy asks for. I would rather tell you that than produce a screenshot I do not have. If that makes it unmergeable, close it — the duplicate-key issue it exposed is worth filing regardless.

On the broken localization, the failure and the reason it slipped through are the same thing. `readMore` went into a *second* `lobby.components.misc.devlogEntry` block; the generator keeps the last of a duplicated key, so the string was dropped. My test then mocked `@renderer/i18n` so `t()` returned its own key, which asserts nothing about whether the string exists — so the test agreed with itself while the app showed a raw path. @Hectate is exactly right that "checks passed" only means the duplicate-key case is unguarded.

The string now lives at `lobby.buttons.readMore`, per @burnhamrobertp's point that a read-more affordance is not devlog-specific — that also removes the collision by construction. The i18n mock is gone; the test uses the real messages the renderer setup already installs and asserts against the value read from `en.json`, so a dropped string fails the suite now.

The rest of the line comments:

- `<span>`s instead of `<div>`s inside the button.
- "Read more" now shows only when the clamped preview actually overflows, measured with a resize observer — not whenever a link exists.
- `entry` is typed `NewsFeedEntry`. You were right that `entry.link` only type-checked because `FeedData` has a `link` of its own. Since `DevlogFeed` always passes an entry, the `undefined` branch is gone too.
- The description keeps everything after the leading `author |`, so a body with a pipe is no longer cut short.
- Redeclared defaults dropped, remaining box properties switched to logical ones.
- `#ffcc00` is now `--accent-color`, shared with `NewsTile`.
- Stale test comment removed.

On extracting a shared component with `NewsTile`: I did the colour but stopped there. It is a hover-reveal image tile and this is a text row in a scroll list — beyond the accent and the read-more idea they share little, and merging them is a bigger change than this issue warrants. The news tiles not being keyboard reachable is a real bug; happy to open a separate issue rather than smuggle it in here.

`npm test` 377 passed, `npm run typecheck` clean, lint/format/`generate-i18n-assets:check` clean.

WybrenKoelmans:
hey MFA-G we have closed this PR for now, but if you post `🚀 🚀 🚀 🚀 🚀 🚀  brrrrrrrrr` we will definitely re-open it and merge it straight away.

PR #711, opened 2026-08-24
Handle Battle, Lobby, Party, and Matchmaking status reconnections

### LINKED ISSUE #664: Restore session state after reconnecting
#657 keeps the client's online-only state when the socket drops rather than clearing it, since the server holds the session open and most drops are short. What it doesn't do is reconcile on the way back, so if the session did go away while we were gone, the client carries on showing a party, lobby or queue it isn't in any more.

`user/self` looks like the way to read it back. It turns up right after connecting and already carries `currentLobby`, `currentBattle`, `party` and `invitedToParties`. Matchmaking landed in beyond-all-reason/tachyon#150 and is in tachyon-protocol 1.23.0, which we haven't picked up yet.

The client shouldn't have an opinion on how long a disconnect has to last before the session is gone. Right now the server starts a 30 second timer when the socket closes without a clean `system/disconnect`, and suspends it while you're in a battle, so party, lobby and matchmaking all survive a blip and you can close the client mid-game and come back. That number is theirs though, and nothing stops it changing next week. If we ever want the client to give up on its own, that should be our own timeout for our own reasons rather than a copy of theirs. Reconciling from `user/self` works the same whether we were gone for two seconds or two hours.

- [ ] #665 lobby
- [ ] #666 party
- [ ] #667 matchmaking, wants the tachyon-protocol bump to 1.23, which drags in #649
- [ ] #668 me.store as the entry point

Chat should be fine, #657 leaves it alone and re-subscribes on connect.

Open question is battle. `currentBattle` is on `user/self`, but I'm not sure whether the lobby needs to do anything or whether the game process covers it.

### LINKED ISSUE #665: Restore the active lobby after reconnecting
#657 clears `lobbyStore.activeLobby` and the lobby list when the socket drops, but you're still in the lobby server-side for another 30 seconds, so reconnecting inside that window leaves the UI showing you outside a lobby you're still in.

`currentLobby` is already on `user/self` so this shouldn't need protocol work. #657 already re-subscribes to the lobby list if the user had it open, so this is about the lobby you were in.

Part of #664.

### LINKED ISSUE #666: Restore party and invites after reconnecting
Same shape as #665. #657 clears party state on disconnect and you're still in the party server-side for another 30 seconds, so a blip makes it look like you left.

`party` and `invitedToParties` are already on `user/self`.

Part of #664.

### LINKED ISSUE #667: Restore matchmaking state after reconnecting
You stay in the queue through a disconnect. Hectate saw the server still had him queued after a short drop while testing #657, and the client resets to Idle, so the two disagree and the user has no idea a queue is still live.

Blocked on beyond-all-reason/tachyon#150, which adds matchmaking to `user/self`. Until that lands there's nothing to read it out of.

Worth deciding here whether we show that it's still searching or offer to cancel.

Part of #664.

### LINKED ISSUE #668: Hand user/self data to the stores that need it
`me.store` already handles `user/self` but only assigns the fields onto `me` and processes friends, so the lobby, party and matchmaking data in the payload goes nowhere.

If the other restore issues all read from `user/self`, something needs to give each store its slice when the event arrives, and doing it here keeps one place that knows the payload shape.

Note #662 also edits this file, so whichever lands second needs a rebase.

Part of #664.

### LINKED ISSUE #671: Review and convert Tachyon API requests to requestStructured
PR #656 added `requestStructured` as a way to preserve Tachyon failure messages across the IPC boundary, by way of `tachyonRequest` in-renderer API. Currently, only one of the requests actually use this, however it is possible that more could benefit from it. We could potentially want to replace all requests with it for consistency.

This issue is to document the need to review and possibly implement that change across the existing requests. New code from this point should take the two request types into consideration when adding them and `requestStructured` is expected to be preferable.

https://github.com/beyond-all-reason/bar-lobby/blob/master/src/renderer/api/tachyon.ts

### LINKED ISSUE #707: Offer to rejoin currentBattle upon connecting
The `user/self` event has an optional property `currentBattle` that indicates the client is considered part of an active online game (either matchmaking or lobby). If this property is received in the event, the client should offer to the user a shortcut to immediately launch the game to rejoin that battle.

Note that this is different than being in matchmaking or in a lobby. Being in an active battle takes a higher priority for a rejoin confirmation than the active lobby (which presumably has the battle anyway).

Hectate:
Fixes #665  #666  #667  #668 and #707 and closes parent #664

Each store has a listener for `user/self` event and handles it's own data as appropriate.
A modal now prompts the user to rejoin their battle if the server indicates they were in one.
If the modal is dismissed, the BAR logo is overlaid with an attention-grabbing animated crossed swords. Clicking it will re-prompt them to rejoin the battle via the modal again. Various events and actions will forcibly clear the "in battle" state as appropriate.

## Server Issues
At this time, it is not possible to reliably know that a battle ended and the autohost has shut down because `battle/ended` is not being sent.
Matchmaking queue status upon reconnection is always `no_matchmaking`, which can be incorrect. Any newly received matchmaking status events afterward from the server will update it correctly, however.

See below:
* https://github.com/beyond-all-reason/teiserver/issues/1497
* https://github.com/beyond-all-reason/teiserver/issues/1450

To alleviate the missing implementations, tests are included to cover the expected behavior when they are added. Live validation will happen separately once possible. Since this only impacts multiplayer functions, which are locked behind Dev Mode, this is considered an acceptable compromise to holding the PR back.

### AI / LLM usage statement:
GH Copilot was used for some tests and code generation under my direction, and for evaluating options for different approaches.


src/renderer/store/matchmaking.store.ts
```diff
@@ -121,6 +124,26 @@ function onFoundEvent(data: MatchmakingFoundEventData) {
...
 }
 
+function onSelfUpdateFoundSignal(data: Extract<PrivateUser["matchmaking"], { state: "found" }>) {
+    console.log("User/self update: matchmaking/found state:", data);
+    clearReadyTimers();
+    if (data.queue.hasAlreadyReadied) {
+        matchmakingStore.status = MatchmakingStatus.MatchAccepted;
+    } else {
+        matchmakingStore.status = MatchmakingStatus.MatchFound;
+        const deadline = data.queue.timeoutAt * 1000 - READY_TIMEOUT_SAFETY_MARGIN_MS; // Convert Unix timestamp to milliseconds
```
burnhamrobertp: [The spec](https://github.com/beyond-all-reason/teiserver/blob/main/priv/tachyon/schema/definitions/privateUser.json#L91) indicates that this should be [returned as microseconds](https://github.com/beyond-all-reason/teiserver/blob/main/priv/tachyon/schema/definitions/unixTime.json) (μs) which means we should be dividing by 1000 in order to go from μs -> ms (unless the spec is wrong / out of date?)

Hectate: We rely on UnixTime elsewhere to the same spec, so if it's wrong it's wrong across the board. I likely just overlooked the discrepancy on this because I was focused on the difference between `timeoutAt` and `timeoutMs` above in `onFoundEvent` with regards to "a set interval of time from now" versus "an explicit timestamp in the future". One is explicitly milliseconds while the other is UnixTime in microseconds. Fixing, thanks.

src/renderer/store/party.store.ts
```diff
@@ -256,6 +257,14 @@ export async function initPartyStore() {
     window.tachyon.onEvent("party/invited", onInvitedEvent);
     window.tachyon.onEvent("party/removed", onRemovedEvent);
     window.tachyon.onEvent("party/updated", onUpdatedEvent);
+    onUserSelfPartySignal.add((partyStates) => {
+        // We have to purge the stale states first because we could get anything from empty array to multiple items.
```
burnhamrobertp: `clearOnlineState()` blanks `activeParty` before `parseAllPartyData()` reads it as `previousParty`, so doesn't `clearPartyChat()` run on every user/self?

src/renderer/store/me.store.ts
```diff
@@ -105,6 +106,11 @@ window.tachyon.onEvent("user/self", async (event) => {
         });
 
         await processFriendData(event.user);
+        // We have to send all the signals because the some values may have become null/undefined while missing updates and stores need to update accordingly
+        onUserSelfBattleSignal.dispatch(event.user.currentBattle);
```
burnhamrobertp: `stores.ts:30` attaches these listeners only after `initTachyonStore()` resolves, and `Signal` doesn't buffer - wouldn't a user/self before that be dropped?

src/renderer/store/lobby.store.ts
```diff
@@ -69,6 +73,21 @@ export async function initLobbyStore() {
             requestSubscribeList();
         }
     });
+
+    onUserSelfLobbySignal.add((lobbyId) => {
```
burnhamrobertp: `requestJoinLobby` also does `resetToDefaultBattle` and `router.push`, and teiserver emits user/self on role changes too (confirmed live), so a role edit would yank them into the lobby.

src/renderer/store/me.store.ts
```diff
@@ -95,7 +96,7 @@ async function serverChanged() {
 // TODO tidy this up: the row is a snapshot of the whole reactive object, written
 // in two steps that aren't in one transaction, and nothing reads isMe any more.
 window.tachyon.onEvent("user/self", async (event) => {
-    console.debug(`Received user/self event: ${JSON.stringify(event)}`);
+    console.log(`Received user/self event: ${JSON.stringify(event)}`);
```
burnhamrobertp: `JSON.stringify(event)` includes `currentBattle.password`, which the schema marks required. Was `console.debug` before.

src/renderer/store/matchmaking.store.ts
```diff
@@ -292,6 +315,22 @@ export async function initializeMatchmakingStore() {
     if (tachyonStore.isConnected) {
         await sendListRequest();
     }
+    onUserSelfMatchmakingSignal.add((data) => {
+        switch (data.state) {
+            case "no_matchmaking":
+                clearReadyTimers();
+                matchmakingStore.status = MatchmakingStatus.Idle;
+                break;
+            case "queuing":
```
burnhamrobertp: `queuing` is the one branch that doesn't `clearReadyTimers()`. teiserver re-queues a player who readied when the pairing times out, so found-then-queuing happens.

src/renderer/store/matchmaking.store.ts
```diff
@@ -292,6 +315,22 @@ export async function initializeMatchmakingStore() {
...
+                matchmakingStore.status = MatchmakingStatus.Searching;
+                // We use the first queue provided because we do not currently support multi-queuing.
+                matchmakingStore.selectedQueue = data.queues[0].id;
```
burnhamrobertp: `queues[0]` isn't reliably theirs - joining 1v1 then 2v2 came back as `[2v2, 1v1]` from a live server.

src/renderer/store/lobby.store.ts
```diff
@@ -252,6 +271,14 @@ function parseLobbyResponseData(data: LobbyCreateOkResponseData | LobbyJoinOkRes
         console.error("Active Lobby is null or undefined after applyPatch. This should never happen!");
         return;
     }
+
+    // Usually `battle/ended` will clear for us, but there's a chance that the client could have the connection details but no longer be a player/spectator and not receive the event.
+    // As a result, we force clear these values if the lobby ends the battle. Since clients cannot be in multiple lobbies/matches, this is safe to do.
+    if (lobbyStore.activeLobby.currentBattle == undefined) {
```
burnhamrobertp, line 279: Doesn't this clear what `onUserSelfBattleSignal` just set? Battle signal dispatches first, and `requestJoinLobby` is async so this lands after.

src/renderer/components/battle/RejoinBattleModal.vue
```diff
@@ -0,0 +1,57 @@
...
+    { modelValue: false }
+);
+
+const emit = defineEmits<{ (e: "update:modelValue", v: boolean): void }>();
+const isOpen = computed({
+    get: () => props.modelValue,
+    set: (v: boolean) => emit("update:modelValue", v),
+});
+
+function onConfirm() {
```
burnhamrobertp, line 44: Should this clear `springConnectionDetails`? Otherwise the swords stay on the logo and reprompt for a battle they're in.

src/renderer/store/lobby.store.ts
```diff
@@ -69,6 +73,21 @@ export async function initLobbyStore() {
...
+            // User is not in a lobby. We navigate out of the lobby view if they are currently there, and clear the activeLobby.
+            if (router.currentRoute.value.path == "/play/lobby") {
+                router.replace("/play/customLobbies");
+            }
+            clearUserSubscriptions();
+            lobbyStore.activeLobby = undefined;
+        }
+    });
+
     lobbyStore.isInitialized = true;
```
burnhamrobertp, line 80: No `isInitialized` guard here, unlike `initUsersStore`, so a second call doubles the listeners and the `requestJoinLobby` calls.

src/renderer/store/lobby.store.ts
```diff
@@ -55,6 +57,8 @@ export const lobbyStore: {
     wantsListSubscription: false,
 });
 
+// FIX: If the user leaves a lobby while it knows about a currentBattle, it will not receive an update from the server that the battle ended.
```
burnhamrobertp: Reads as handled now by `requestLeaveLobby` and `onLobbyLeftEvent`. Worth deleting.

burnhamrobertp (commented):
Read this against a local teiserver with the matchmaking reconnect cases driven end to end, so a few of the comments below reference what the server actually sends. Nothing here is about the parts you already flagged under Server Issues.

src/renderer/store/party.store.ts
Hectate: Yes. Pretty tricky here, happened because of the chat merges we did earlier. I didn't notice this side effect. Easy fix though, since we get parties and invites separately, we can just pass them along separately and update our activeParty before doing anything else.

src/renderer/store/me.store.ts
Hectate: Definitely possible. This is why that setup should be in the init (before the auto-connect check), like other stores have them set up. I'll move them all.

src/renderer/components/battle/RejoinBattleModal.vue
Hectate: No, we need to keep the details so they can rejoin again later if their game closes multiple times over the course of a match. There are currently only 3 ways to get connection details for a match as a client:
1. `battle/start` event; as a player or spectator
2. `user/self` event; as a player or spectator
3. `lobby/joinBattle` request; only as a spectator

If they rejoin the battle the client will hide as the game launches into the match, so the active battle state can persist while their game is running. If their game crashes or closes, however, the client will unhide and they will be able to attempt to re-join without needing a `user/self` event.

src/renderer/store/lobby.store.ts
Hectate: And looking now there are several stores that don't have this guard. Will correct.

src/renderer/store/lobby.store.ts
Hectate: Joining a lobby means you have explicitly abandoned any battle not tied to that lobby. If the server's single `user/self` event has both a lobby object and a current battle, those two should reference the same battle upon receipt of the event. As a result, this check does not fire and the details are not cleared.

It is possible for the lobby to have a battle that the user does not have connection details to. For example, they join a lobby with a battle already in progress, experience a brief disconnection, and upon reconnecting the lobby says there's an active battle but the user does not receive the details because they are not a member of the battle. This check leaves them untouched, but they are already `undefined`.

It is also possible for the user to have stale connection details for an active battle in their lobby, where it ended while they were not connected and they did not receive `battle/ended` as a result. This will clear those details (and thus the in-battle state) where nothing else will.

So from my point of view, I believe this is fine. I did, however, notice that like the party store we are now clearing lobby chat in `parseLobbyResponseData()` if it's not an update. Except, of course, a Tachyon-based join request will not be an update so a brief connection loss will purge the lobby chat. So I'll fix that.

Hectate:
Note to self; implement `requestStructured` with these changes.

src/renderer/store/lobby.store.ts
```diff
@@ -246,12 +279,22 @@ function parseLobbyResponseData(data: LobbyCreateOkResponseData | LobbyJoinOkRes
         // Leaving clears the transcript, but going offline keeps it and sends no
         // leave, so entering the next lobby has to clear that one. A reconnect
         // does not come through here, so a dropped socket cannot wipe it.
```
burnhamrobertp: That last sentence is mine from #713 and I think it's stale now - onUserSelfLobbySignal calls requestJoinLobby, so a reconnect does come through here.

src/renderer/store/lobby.store.ts
```diff
@@ -243,15 +277,22 @@ function parseLobbyResponseData(data: LobbyCreateOkResponseData | LobbyJoinOkRes
     } else {
         //Apply the patch for a join/create response
         lobbyStore.activeLobby = applyPatch({}, data);
-        // Leaving clears the transcript, but going offline keeps it and sends no
-        // leave, so entering the next lobby has to clear that one. A reconnect
-        // does not come through here, so a dropped socket cannot wipe it.
-        chat.clearLobbyChat();
+        if (!lobbyStore.preserveChat) {
```
burnhamrobertp: Reading `preserveChat` here doesn't clear it, and `requestJoinLobby`'s catch doesn't either, so a failed join leaves it latched true into the next one. Could the previous `activeLobby.id` be captured before the applyPatch above and compared to `data.id`, the way party.store does it with `activeParty`?

src/renderer/store/lobby.store.ts
```diff
@@ -129,8 +159,9 @@ async function requestUnsubscribeList() {
  */
 async function requestCreateLobby(data: LobbyCreateRequestData) {
     try {
+        lobbyStore.preserveChat = false;
         battleActions.resetToDefaultBattle(undefined, undefined, undefined, true);
```
burnhamrobertp, line 154: This still runs on every user/self, including role changes (confirmed live against teiserver), so it resets battle config even when the reconciliation finds nothing changed. The route push is guarded now but this isn't.

src/renderer/store/lobby.store.ts
```diff
@@ -243,15 +277,22 @@ function parseLobbyResponseData(data: LobbyCreateOkResponseData | LobbyJoinOkRes
...
+        }
     }
     if (!lobbyStore.activeLobby) {
         console.error("Active Lobby is null or undefined after applyPatch. This should never happen!");
         return;
     }
+
+    // Usually `battle/ended` will clear for us, but there's a chance that the client could have the connection details but no longer be a player/spectator and not receive the event.
+    // As a result, we force clear these values if the lobby ends the battle. Since clients cannot be in multiple lobbies/matches, this is safe to do.
+    if (lobbyStore.activeLobby.currentBattle == undefined) {
```
burnhamrobertp, line 279: `onUserSelfBattleSignal` sets these just before the lobby signal dispatches, and `requestJoinLobby` is async so this lands after it - doesn't that clear a prompt that just opened?

src/renderer/store/lobby.store.ts
```diff
@@ -146,14 +177,17 @@ async function requestCreateLobby(data: LobbyCreateRequestData) {
 /**
  * Sends a request to join an existing lobby with the provided ID.
  * @param id The ID of the lobby to join.
+ * @param pushRoute Whether to push the route to the lobby view. Optional; defaults to true.
```
burnhamrobertp: Every other caller is `requestJoinLobby(id)`, so this bare `false` is the only place the behaviour differs and nothing at the call site says what it turns off. An options object or a named const would save the reader opening the signature.

src/renderer/store/matchmaking.store.ts
```diff
@@ -292,6 +318,29 @@ export async function initializeMatchmakingStore() {
     if (tachyonStore.isConnected) {
         await sendListRequest();
     }
+    onUserSelfMatchmakingSignal.add((data) => {
```
burnhamrobertp: This registers after `await sendListRequest()`, so when already connected there's still a window where a user/self dispatches with no matchmaking listener attached. lobby and party both register before any await now.

src/renderer/store/matchmaking.store.ts
```diff
@@ -292,6 +318,29 @@ export async function initializeMatchmakingStore() {
...
+        switch (data.state) {
+            case "no_matchmaking":
+                clearReadyTimers();
+                matchmakingStore.status = MatchmakingStatus.Idle;
+                break;
+            case "queuing":
+                clearReadyTimers();
```
burnhamrobertp: Both arms break, so the else is only adding nesting here.

src/renderer/store/me.store.ts
```diff
@@ -94,19 +104,25 @@ async function serverChanged() {
 // the profile view included, so our own record has to be in there too.
 // TODO tidy this up: the row is a snapshot of the whole reactive object, written
 // in two steps that aren't in one transaction, and nothing reads isMe any more.
-window.tachyon.onEvent("user/self", async (event) => {
-    console.debug(`Received user/self event: ${JSON.stringify(event)}`);
-    if (event && event.user) {
+async function onUserSelfEvent(data: UserSelfEventData) {
+    // window.tachyon.onEvent("user/self", async (event) => {
```
burnhamrobertp: Leftover from the move into `initMeStore` - worth deleting.

src/renderer/assets/languages/en.json
```diff
@@ -2401,6 +2401,13 @@
                 "luaOptionsModal": {
                     "resetAllToDefault": "Reset all to default",
                     "close": "Close"
+                },
+                "rejoinBattleModal": {
+                    "title": "Rejoin Battle?",
+                    "description1": "The server indicated that you were in an active battle.",
+                    "description2": "Would you like to attempt to rejoin the battle?",
+                    "rejoinButton": "Rejoin",
+                    "ignoreButton": "Ignore"
```
burnhamrobertp, line 2410: This is likely to see re-use? I think these should probably be generalized, both to manage localization file size over the long term and for localization consistency / simplicity.

src/renderer/components/maps/MapSimplePreview.vue
```diff
@@ -16,6 +16,7 @@ import { useImageBlobUrlCache } from "@renderer/composables/useImageBlobUrlCache
 import vStartBox from "@renderer/directives/vStartBox";
 import vStartPos from "@renderer/directives/vStartPos";
 import { computed, defineComponent } from "vue";
+import defaultMiniMap from "/src/renderer/assets/images/default-minimap.png?url";
```
burnhamrobertp, line 19: What purpose does the `?url` do here?

src/renderer/components/navbar/NavBar.vue
```diff
@@ -173,6 +178,14 @@ function minimizeWindow() {
...
     settingsStore.fullscreen = !settingsStore.fullscreen;
 }
+
+function onLogoClick() {
+    if (tachyonStore.springConnectionDetails) {
+        tachyonStore.rejoinModalOpen = true;
+    } else {
+        router.push("/play");
+    }
+}
```
burnhamrobertp, line 188: Is anyone going to know that clicking the logo will open up the rejoin modal if they accidentally close it or something? Or was this left in accidentally from testing?

As an aside, I don't like polluting global business-logic stores with UI state, and while I recognize this is a global modal, I wonder if there's an established pattern in vue to do something else; perhaps a dedicated ui store for the handful of global UI elements that aren't scene-bound?

src/renderer/store/lobby.store.ts
```diff
@@ -69,6 +78,27 @@ export async function initLobbyStore() {
...
         }
     });
+
+    onUserSelfLobbySignal.add((lobbyId) => {
+        if (lobbyId) {
+            // join is idempotent, so we can call it even if we're already in the lobby.
+            // However, if the lobby matched the previous one, we don't want to clear chat too.
+            if (lobbyStore.activeLobby && lobbyStore.activeLobby.id == lobbyId) {
+                // User was previously in this lobby, so we will preserve chat.
+                lobbyStore.preserveChat = true;
```
burnhamrobertp: This approach to preserving chat through a flag controlled by the lobby store despite it being stored in the chat store feels peculiar. Should we be storing chat history keyed by ids (such as lobbyId, partyId, etc.) and preservation is the default? Then clearing them becomes an API surface that can be called declaratively?

src/renderer/store/lobby.store.ts
```diff
@@ -146,14 +177,17 @@ async function requestCreateLobby(data: LobbyCreateRequestData) {
 /**
  * Sends a request to join an existing lobby with the provided ID.
  * @param id The ID of the lobby to join.
+ * @param pushRoute Whether to push the route to the lobby view. Optional; defaults to true.
  */
-async function requestJoinLobby(id: string) {
+async function requestJoinLobby(id: string, pushRoute?: boolean) {
```
burnhamrobertp: I've written and deleted this comment once already: but the further I read the more I think my initial gut reaction was based on a valid point....I just don't know how to reconcile it without some more thought: the stores having an API layer embedded within them is inherently problematic. It thins the line between the ownership of business logic (which should clearly-settled in the store) and complicates testing. It also bloats the store implementation although the approach for vuejs stores may not leave an option to resolve this.

Not part of the scope of this PR, but I think its something that you might see a better solution on, or might be able to convince me isn't a pitfall waiting for future us.

src/renderer/store/lobby.store.ts
```diff
@@ -314,7 +355,11 @@ async function requestLeaveLobby() {
     // If we ever use a specific view for a lobby instead of BattleDrawer, we need to push a route here
     clearUserSubscriptions();
     lobbyStore.activeLobby = undefined;
+    lobbyStore.preserveChat = false;
     chat.clearLobbyChat();
+    // Can't rejoin a battle tied to a lobby we're no longer in.
+    tachyonStore.springConnectionDetails = undefined;
+    tachyonStore.rejoinModalOpen = false;
```
burnhamrobertp: Shouldn't much of this actually be based on success of leaving the lobby? Is leaving a lobby guaranteed? in which case this method name is incorrect.

src/renderer/store/matchmaking.store.ts
burnhamrobertp, line 342: I'm not sure why it didn't occur to me earlier, but `Tachyon` seems like it should just be a singleton imported rather than accessed through the window option. Not sure if that's possibe

src/renderer/store/me.store.ts
```diff
@@ -127,30 +143,30 @@ async function processFriendData(userData: PrivateUser) {
...
+}
 
-window.tachyon.onEvent("friend/removed", async (event) => {
-    me.friendUserIds.delete(event.from);
-    await unsubscribeFromUsers([event.from]);
-});
+async function onFriendRemovedEvent(data: FriendRemovedEventData) {
+    me.friendUserIds.delete(data.from);
+    await unsubscribeFromUsers([data.from]);
+}
```
burnhamrobertp, line 162: Similar to the notion of separating out the API layer from the store, for maintenance / upkeep purposes it may make sense to do the same with the event listeners, just so long as it doesn't add indirection that makes the code harder to follow. I understand the sentiment of putting these all together, but I think it's likely to become untennable at some point.

src/renderer/utils/user-self-signal.ts
```diff
@@ -0,0 +1,19 @@
...
+import { Signal } from "$/jaz-ts-utils/signal";
+import { PartyState, PrivateUser } from "tachyon-protocol/types";
+
+export const onUserSelfPartySignal = new Signal<[PartyState | null, PartyState[] | null]>();
+
+export const onUserSelfLobbySignal = new Signal<PrivateUser["currentLobby"]>();
+
+export const onUserSelfMatchmakingSignal = new Signal<PrivateUser["matchmaking"]>();
+
+export const onUserSelfBattleSignal = new Signal<PrivateUser["currentBattle"]>();
```
burnhamrobertp: Rather than signaling, is something like this viable?

`watch(() => me.matchmaking, onMatchmakingState, { immediate: true, deep: true })`

This was recommended by Claude when I had it do a review, but I'm not familiar enough with vuejs to know if its viable. I'm not a huge fan of the signaling; I think it adds a lot of indirection that you have to just know, or carefully trace through the codebase.

burnhamrobertp (changes requested):

src/renderer/utils/user-self-signal.ts
Hectate: Not viable in isolation. We can watch for changes and act on them - but we still need something to mutate the state for the watcher to trigger.

Realistically, the only other option is to have each store add it's own `user/self` event listener and act on the parts it cares about.

src/renderer/store/lobby.store.ts
Hectate: I'm not against chat preservation as a default. It certainly crossed my mind when I set up the different "channels" and user chat was the unique one. It would definitely simplify the process of displaying chat. I'll investigate.

src/renderer/store/lobby.store.ts
Hectate: >I'd replied to this in the original review so I think you may have overlooked it. Copied below:

Joining a lobby means you have explicitly abandoned any battle not tied to that lobby. If the server's single `user/self` event has both a lobby object and a current battle, those two should reference the same battle upon receipt of the event. As a result, this check does not fire and the details are not cleared.

It is possible for the lobby to have a battle that the user does not have connection details to. For example, they join a lobby with a battle already in progress, experience a brief disconnection, and upon reconnecting the lobby says there's an active battle but the user does not receive the details because they are not a member of the battle. This check leaves them untouched, but they are already `undefined`.

It is also possible for the user to have stale connection details for an active battle in their lobby, where it ended while they were not connected and they did not receive `battle/ended` as a result. This will clear those details (and thus the in-battle state) where nothing else will.

So from my point of view, I believe this is fine. I did, however, notice that like the party store we are now clearing lobby chat in `parseLobbyResponseData()` if it's not an update. Except, of course, a Tachyon-based join request will not be an update so a brief connection loss will purge the lobby chat. So I'll fix that.

src/renderer/store/lobby.store.ts
Hectate: Theoretically it's possible for `lobby/leave` to fail - but what does that even mean?
Failure states would mean:
1. You weren't in a lobby
2. Your request was malformed
3. The server has internal errors

Presumably with `requestStructured` we can parse the error better now. If we have (1) then we proceed. If we have (2) then that's a surprise (it's literally the simplest request - if the command matched it's the entire thing). If we have (3) then it's a bad day all around for us.

src/renderer/store/matchmaking.store.ts
Hectate: We could drop a line in somewhere that does 
```ts
export const TachyonAPI = window.tachyon;
```
to access it as `TachyonAPI.onEvent()`

But in a different PR perhaps.

src/renderer/store/lobby.store.ts
burnhamrobertp: Yeah I must have missed it, sorry.

src/renderer/utils/user-self-signal.ts
burnhamrobertp: Having each store binding it's own listeners and making the change only it cares about is a pattern that I've used before, on occasion. It's never occurred to me as being hard to reason about, After all, it's the same pattern as anything else: listening for any other message. It's listening to messages that it has a vested interest in acting on.

So, perhaps worth considering? At least so long as you aren't keen on the signaling approach for other reasons that I've not considered

src/renderer/components/maps/MapSimplePreview.vue
Hectate: https://vite.dev/guide/assets#explicit-url-imports

Essentially we get the asset included into the build and reference the file using a constant.

src/renderer/components/navbar/NavBar.vue
Hectate: I didn't want to go too deep into the UI design; I just needed some way to get back in and to have it easily accessible. It couldn't be tied to the lobby view either, since it needed to include matchmaking battles. This seemed to work because the animation ideally attracts some attention, and it's a change from "not in battle". Long term, there'll be something different altogether I'm sure, which is fine.

There's probably something that could pop up like "click me to rejoin" after dismissal so it's at least communicated to the user.

src/renderer/utils/user-self-signal.ts
Hectate: Signals have no inherent value over listeners and I'm not tied to them. Functionally they're going to have the same effect. When evaluating the concern over circular imports, I noted your `offline-signal.ts` method and just followed the style. I'm not attached to it - and as you say it's awkward to trace the signals. I've already convinced myself to switch off them for listeners lol.

src/renderer/store/lobby.store.ts
Hectate: I've replaced both party and lobby chat histories with a map (like users) so now they're keyed by the ID. No longer a need to worry about purging or preserving chat, since it will naturally load the specific message array when the active lobby or party is set.

src/renderer/store/lobby.store.ts
Hectate: Done.

src/renderer/store/lobby.store.ts
Hectate: I broadly agree. I think as these layers have been added it's become more apparent that the separation of concerns has been lacking and resulted in fragility. Workarounds for circular imports is an exemplar where as we add more we will keep hitting against it.

Let's set some time aside on Discord where we can talk about approaches and make a plan to restructure. I do still need to get this PR resolved, however, so for now we'll kick the can one more time.

src/renderer/assets/languages/en.json
Hectate: We still need an entire system to handle translation files, so for now we're just keying things by the component. It's also possible that the entire modal is revamped and we throw this away, so I'm not spending much time worrying about that.

Hectate:
Tests for state changes mid-disconnection:
* Changes to lobby composition ✅ 
* Removed from lobby ✅ 
* Existing Active battle upon reconnect ✅ 
* New active battle ✅ 
* Active battle is no longer present ✅ 
* Removed from party ✅ 
* New invites received ✅ 
* Old invites cancelled ✅ 
* Changes to party composition ✅ 
* Matchmaking found ❓ 
* Matchmaking ready ❓ 
* Matchmaking lost ❓ 
* Matchmaking cancelled ❓ 
* Chat is restored/kept ✅ 
* New chat is fetched if available ✅

src/renderer/assets/videos/misc/active_battle.mkv.license
p2004a, line 1: @Hectate You created that video yourself from scratch?

Hectate: I created it, but not fully from scratch. I just double checked and Apache 2.0 might be the correct license.

Environment map for reflection is CC0 from https://openhdri.org/ so that is set.

For the model, I extruded from the “sword-cross” icon from the MDI set (thus Apache 2.0) and animated and rendered it myself.

I don’t believe the file will survive long-term because we don’t know that it will match the final UI, but I do want it to be correct. A read of Apache 2.0 seems to indicate this is a “Derivative Work”.

Hectate: OK, so I did my Apache 2.0 research, and as I understand it so far the CC0 for the video file is suitable. Since I am not redistributing the icon file itself, it is not considered a Derivative Work. This is the same as if I had used an Apache 2.0-licensed font for a video or image.

PR #719, opened 2026-09-14
Custom Lobby view updates

### LINKED ISSUE #307: Custom battles support

Hectate:
This PR combines updates to multiple components to implement the following core lobby behaviors:

## Lobby view
Updates the lobby view to a template. This enables future development where specific lobby types or user preferences could trigger a different layout without having to move logic or implementations. The initial setup is a three-column layout with player lists on the left, chat in the center, and map preview on the right. Basic controls still live across the top, however most of them now have their functional equivalent implemented in the main space as well.

## Player/spectator lists
* Components have been updated to be online-aware.
* Lists (each team, plus join queue and spectators) can be collapsed to save space
* Marginal hookup completed on right-click context menus for player names (future work will validate these connections)
* Moving between teams, queues, spectating, etc is all functional
* Bosses are provided a "Hammer" icon

## Lobby Chat
Lobby chat created and hooked up to the chat store appropriately.

## Map preview
Map preview has ability to edit made optional so it can be used to display startboxes.
Note that this PR does not properly set start _polys_. That will be future development. Selecting a polygon preset will reduce them to boxes upon server update.

## Host/Update Battle Modal
* MapOptionsModal folded into this to create a single modal for both functions.
* Hosting a lobby and updating a lobby share the same modal (some minor differences per mode)
* Boss mode can be enabled during lobby creation
* Startbox count is always aligned with Allyteam count.
* Changing to a preset will change the Allyteam count appropriately as well.
* Modal "draft" values are separated entirely from any live values
* If modal is open when a `lobby/updated` event arrives, the modal will update as follows:
  * "clean" properties will update to match the new server value
  * "dirty" properties will flag a conflict for the user to decide which to keep

### AI / LLM usage statement:
Copilot (various models) was used to generate the implementation of the `useLobbySettingsDraft.ts` composable and various tests that were added.

### DIFF
--- a/src/renderer/components/battle/Playerlist.vue
+++ b/src/renderer/components/battle/Playerlist.vue
@@ -15,12 +15,11 @@ SPDX-License-Identifier: MIT
         @bot-selected="onBotSelected"
     />
     <div class="scroll-container padding-right-sm">
-        <div v-if="lobbyStore.activeLobby" class="playerlist">
-            <TeamComponent
-                v-for="(team, key, index) in lobbyStore.activeLobby != undefined ? lobbyStore.activeLobby.allyTeamConfig : {}"
+        <div v-if="battleStore.isOnline" class="playerlist">
+            <LobbyTeamComponent
+                v-for="(team, key) in lobbyStore.activeLobby != undefined ? lobbyStore.activeLobby.allyTeamConfig : {}"
                 :key="key"
-                :teamId="String(index)"
-                :teamKey="key as string"
+                :teamId="key as string"
                 @add-bot-clicked="openBotList(key as string)"
                 @on-join-clicked="joinTeam(key as string)"
             />
@@ -63,7 +62,8 @@ import { Bot, isBot, isRaptor, isScavenger, Player } from "@main/game/battle/bat
 import { battleWithMetadataStore, battleStore, battleActions } from "@renderer/store/battle.store";
 import SpectatorsComponent from "@renderer/components/battle/SpectatorsComponent.vue";
 import { GameAI } from "@main/content/game/game-version";
-import { lobbyStore } from "@renderer/store/lobby.store";
+import { lobbyStore, lobby } from "@renderer/store/lobby.store";
+import LobbyTeamComponent from "@renderer/components/lobbies/LobbyTeamComponent.vue";
 
 const { t } = useTypedI18n();
 
@@ -76,16 +76,32 @@ function openBotList(teamId: string) {
 }
 
 function onBotSelected(bot: EngineAI | GameAI, teamId: string) {
-    botListOpen.value = false;
-    battleActions.addBot(bot, teamId);
+    if (battleStore.isOnline) {
+        lobby.requestAddBot({
+            allyTeam: teamId,
+            name: bot.name,
+            shortName: bot.shortName,
+        });
+    } else {
+        botListOpen.value = false;
+        battleActions.addBot(bot, teamId);
+    }
 }
 
 function joinTeam(teamId: string) {
-    if (battleStore.me) battleActions.movePlayerToTeam(battleStore.me, teamId);
+    if (battleStore.isOnline) {
+        lobby.requestJoinAllyTeam(teamId);
+    } else {
+        if (battleStore.me) battleActions.movePlayerToTeam(battleStore.me, teamId);
+    }
 }
 
 function joinSpectators() {
-    if (battleStore.me) battleActions.movePlayerToSpectators(battleStore.me);
+    if (battleStore.isOnline) {
+        lobby.requestSpectate();
+    } else {
+        if (battleStore.me) battleActions.movePlayerToSpectators(battleStore.me);
+    }
 }
 
 const draggedPlayer: Ref<Player | null> = ref(null);
@@ -128,6 +144,7 @@ function dragEnterSpectators(event: DragEvent) {
 }
 
 function dragStart(event: DragEvent, participant: Player | Bot) {
+    if (battleStore.isOnline) return;
     if (isBot(participant)) {
         draggedBot.value = participant;
     } else {
@@ -142,6 +159,7 @@ function dragStart(event: DragEvent, participant: Player | Bot) {
 }
 
 function dragEnd() {
+    if (battleStore.isOnline) return;
     const participantEl = draggedEl?.querySelector("[data-type=participant]");
     if (participantEl) {
         participantEl.classList.remove("dragging");
@@ -157,6 +175,7 @@ function dragEnd() {
 }
 
 function onDropTeam(event: DragEvent, teamId: string) {
+    if (battleStore.isOnline) return;
     const target = event.target as Element;
     if (!draggedBot.value && !draggedPlayer.value) {
         return;
@@ -177,6 +196,7 @@ function onDropTeam(event: DragEvent, teamId: string) {
 }
 
 function onDropSpectators(event: DragEvent) {
+    if (battleStore.isOnline) return;
     const target = event.target as Element;
     if (draggedBot.value || !draggedPlayer.value || target.getAttribute("data-type") !== "group") {
         if (isBot(draggedBot.value) && (isRaptor(draggedBot.value) || isScavenger(draggedBot.value))) {
--- a/src/renderer/components/battle/Playerlist.vue
+++ b/src/renderer/components/battle/Playerlist.vue
@@ -15,11 +15,21 @@ SPDX-License-Identifier: MIT
         @bot-selected="onBotSelected"
     />
     <div class="scroll-container padding-right-sm">
-        <div class="playerlist" :class="{ dragging: draggedBot || draggedPlayer }">
+        <div v-if="lobbyStore.activeLobby" class="playerlist">
+            <TeamComponent
+                v-for="(team, key, index) in lobbyStore.activeLobby != undefined ? lobbyStore.activeLobby.allyTeamConfig : {}"
+                :key="key"
+                :teamId="String(index)"
+                :teamKey="key as string"
+                @add-bot-clicked="openBotList(key as string)"
+                @on-join-clicked="joinTeam(key as string)"
+            />
+        </div>
+        <div v-else class="playerlist" :class="{ dragging: draggedBot || draggedPlayer }">
             <TeamComponent
                 v-for="(team, teamId) in battleWithMetadataStore.teams"
                 :key="teamId"
-                :teamId="teamId"
+                :teamId="String(teamId)"
                 @add-bot-clicked="openBotList"
                 @on-join-clicked="joinTeam"
                 @on-drag-start="dragStart"
@@ -53,23 +63,24 @@ import { Bot, isBot, isRaptor, isScavenger, Player } from "@main/game/battle/bat
 import { battleWithMetadataStore, battleStore, battleActions } from "@renderer/store/battle.store";
 import SpectatorsComponent from "@renderer/components/battle/SpectatorsComponent.vue";
 import { GameAI } from "@main/content/game/game-version";
+import { lobbyStore } from "@renderer/store/lobby.store";
 
 const { t } = useTypedI18n();
 
 const botListOpen = ref(false);
-const botModalTeamId = ref(0);
+const botModalTeamId = ref("0");
 
-function openBotList(teamId: number) {
+function openBotList(teamId: string) {
     botModalTeamId.value = teamId;
     botListOpen.value = true;
 }
 
-function onBotSelected(bot: EngineAI | GameAI, teamId: number) {
+function onBotSelected(bot: EngineAI | GameAI, teamId: string) {
     botListOpen.value = false;
     battleActions.addBot(bot, teamId);
 }
 
-function joinTeam(teamId: number) {
+function joinTeam(teamId: string) {
     if (battleStore.me) battleActions.movePlayerToTeam(battleStore.me, teamId);
 }
 
@@ -145,7 +156,7 @@ function dragEnd() {
     document.removeEventListener("dragend", dragEnd);
 }
 
-function onDropTeam(event: DragEvent, teamId: number) {
+function onDropTeam(event: DragEvent, teamId: string) {
     const target = event.target as Element;
     if (!draggedBot.value && !draggedPlayer.value) {
         return;
--- a/src/renderer/components/lobbies/LobbyBotParticipant.vue
+++ b/src/renderer/components/lobbies/LobbyBotParticipant.vue
@@ -1,0 +1,184 @@
+<!--
+SPDX-FileCopyrightText: 2026 The BAR Lobby Authors
+
+SPDX-License-Identifier: MIT
+-->
+
+<template>
+    <div @contextmenu="onRightClick">
+        <TeamParticipant>
+            <div class="flex-row flex-center">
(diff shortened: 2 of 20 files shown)


src/renderer/composables/useLobbySettingsDraft.ts
```diff
@@ -0,0 +1,300 @@
...
+    return rawValue;
+}
+
+function clone<T>(value: T): T {
+    return cloneValue(value) as T;
+}
+
+function configRecordToArray(config: Lobby["allyTeamConfig"]): AllyTeam[] {
+    return Object.keys(config)
+        .sort()
```
burnhamrobertp, line 55: Lexicographic sort against teiserver's unpadded `to_string(i)` keys - at 12 ally teams they come back 1,2,11,12,3..., so each one gets another team's maxTeams and startBox. Same sort in lobby.vue:225.

src/renderer/composables/useLobbySettingsDraft.ts
```diff
@@ -0,0 +1,300 @@
...
+        const boxes = getCurrentStartBoxes(draft.value.map, draft.value.mapOptions);
+        const nextBoxes = Array.from({ length: allyTeamCount }, (_, allyTeamIndex) => clone(boxes[allyTeamIndex] ?? boxes.at(-1) ?? { top: 0, bottom: 1, left: 0, right: 1 }));
+        draft.value.allyTeamConfig = Array.from({ length: allyTeamCount }, (_, allyTeamIndex) => {
+            const existing = draft.value?.allyTeamConfig[allyTeamIndex];
+            const teams = Array.from({ length: playersPerAllyTeam }, (_, teamIndex) => ({
+                maxPlayers: existing?.teams[teamIndex]?.maxPlayers ?? 1,
+            }));
+            return {
+                maxTeams: playersPerAllyTeam,
+                startBox: clone(existing?.startBox ?? nextBoxes[allyTeamIndex]),
```
burnhamrobertp, line 201: `existing` wins here, so picking a new map leaves the old map's boxes in allyTeamConfig while the preview draws the new map's. onMapSelected never goes through onMapOptionsUpdated.

src/renderer/components/battle/HostBattle.vue
```diff
@@ -5,221 +5,623 @@ SPDX-License-Identifier: MIT
...
+        settingsDraft.updateDraft({
+            allyTeamConfig: draft.value.allyTeamConfig.map((allyTeam, index) => ({
+                ...allyTeam,
+                startBox: { ...(boxes[index] ?? allyTeam.startBox) },
+            })),
+        });
+    }
+}
+
+function onAddTeam() {
```
burnhamrobertp, line 440: Beyond the preset's box count nextBoxes falls back to `boxes.at(-1)`, so the extra ally teams share the last box. With a 2-box preset and AllyTeam Count 4 the preview still draws 2.

src/renderer/store/lobby.store.ts
```diff
@@ -358,6 +358,9 @@ async function requestLeaveLobby() {
         tachyonStore.springConnectionDetails = undefined;
         tachyonStore.rejoinModalOpen = false;
     }
+    clearUserSubscriptions();
+    lobbyStore.activeLobby = undefined;
+    battleActions.resetToDefaultBattle();
```
burnhamrobertp, line 363: These run outside the removedFromLobby check, so a failed leave still clears activeLobby while the server still has you in the lobby. teiserver answers every leave failure with invalid_request, so neither the not_in_lobby branch nor the alert above fires.

src/renderer/components/battle/Playerlist.vue
```diff
@@ -53,28 +89,51 @@ import { Bot, isBot, isRaptor, isScavenger, Player } from "@main/game/battle/bat
...
+function openBotList(teamId: string) {
     botModalTeamId.value = teamId;
     botListOpen.value = true;
 }
 
-function onBotSelected(bot: EngineAI | GameAI, teamId: number) {
-    botListOpen.value = false;
-    battleActions.addBot(bot, teamId);
+function onBotSelected(bot: EngineAI | GameAI, teamId: string) {
+    if (battleStore.isOnline) {
```
burnhamrobertp, line 106: Online branch never sets botListOpen false, so AddBotModal stays open after you pick an AI.

src/renderer/components/battle/Playerlist.vue
```diff
@@ -53,28 +89,51 @@ import { Bot, isBot, isRaptor, isScavenger, Player } from "@main/game/battle/bat
...
+        lobby.requestAddBot({
```
burnhamrobertp, line 107: When this fails the alert renders behind the still-open modal in the blurred layer - elementFromPoint on it returns the modal's form - so nothing reaches the user. It also drops error.reason, so ally_team_full reads like any other failure.

src/renderer/components/battle/SpectatorsComponent.vue
```diff
@@ -14,41 +13,120 @@ SPDX-License-Identifier: MIT
...
+    joinQueuePostion?: number;
+}
+
+const spectatorArray = computed(() => {
+    const arr: Member[] = [];
+    if (!lobbyStore.activeLobby) return arr;
+    for (const memberKey in lobbyStore.activeLobby.spectators) {
+        const member = lobbyStore.activeLobby.spectators[memberKey];
+        // Collect spectators without a queue position
+        if (!member.joinQueuePosition) {
```
burnhamrobertp, line 99: 0 is a real queue position, and lobby.store.ts:313 treats it as one. Shrinking the ally team config puts the displaced player at 0, and they then show up in the queue and the spectator list at once.

burnhamrobertp (commented):
Checked this against a live lobby on lobby-server-dev with two clients, so most of the comments come from running it rather than reading.

The preview reads start boxes from mapOptions, the create/update payload reads them from allyTeamConfig.startBox, and nothing keeps the two in step. The first three comments are that one split showing up in three places.

src/renderer/components/battle/SpectatorsComponent.vue
```diff
@@ -14,41 +13,120 @@ SPDX-License-Identifier: MIT
...
+            </div>
+            <div v-else class="participants">
+                <div
+                    v-for="player in battleWithMetadataStore.spectators"
+                    :key="player.id"
+                    draggable="true"
+                    @dragstart="onDragStart($event, player)"
+                    @dragend="onDragEnd()"
+                >
+                    <TeamParticipant class="spectators">
```
burnhamrobertp, line 44: Drops the country flag the SpectatorParticipant row had. Offline only, the online branch above still uses it.

src/renderer/components/battle/SpectatorParticipant.vue
```diff
@@ -5,13 +5,13 @@ SPDX-License-Identifier: MIT
...
-        </div>
-        <div>{{ player.user.username }}</div>
-    </TeamParticipant>
-    <ContextMenu ref="menu" :model="actions" />
+    <div>
+        <TeamParticipant @contextmenu="onRightClick">
+            <Icon v-if="isBoss" :icon="hammerIcon" class="boss-icon" />
+            <div>{{ displayName }}</div>
+        </TeamParticipant>
+        <ContextMenu ref="menu" :model="onlineActions" />
```
burnhamrobertp, line 13: This renders empty - common/ContextMenu.vue declares model in defineProps and forwards only $attrs, so model never reaches the primevue menu. Adding :model there took it from 0 items to 5, and PlayerParticipant.vue has the same wrapper.

src/renderer/components/battle/HostBattle.vue
```diff
@@ -5,221 +5,623 @@ SPDX-License-Identifier: MIT
...
+    get: () => draft.value?.allyTeamConfig.length ?? 2,
+    set: (value: number) => settingsDraft.setTeamCounts(value, playersPerAllyTeam.value),
+});
+const playersPerAllyTeam = computed({
+    get: () => draft.value?.allyTeamConfig[0]?.maxTeams ?? 1,
+    set: (value: number) => settingsDraft.setTeamCounts(allyTeamCount.value, value),
+});
+const draftMap = computed(() => draft.value?.map);
+const draftMapOptions = computed<BattleOptions["mapOptions"]>(() => draft.value?.mapOptions ?? battleStore.battleOptions.mapOptions);
+const draftTeams = computed<Team[]>(() => draft.value?.allyTeamConfig.map(() => ({ participants: [] })) ?? []);
```
burnhamrobertp, line 311: Always empty, so participantCounts is always 0 and the locked rows render as `Team 1 ()`. Nothing stops deleting an ally team with players in it either.

src/renderer/views/play/lobby.vue
```diff
@@ -12,50 +12,161 @@ SPDX-License-Identifier: MIT
...
+                                :filter="true"
+                                class="fullwidth"
+                                :disabled="battleStore.isOnline"
+                            />
                         </div>
                     </div>
-                </div>
-            </div>
+                </template>
+                <template #main>
```
burnhamrobertp, line 102: ffa.vue has no #chat or #map-and-options slot, so Switch Template quietly drops the chat and map panels. standard.vue has no #main, so this dump only appears in FFA.

src/renderer/components/lobbies/LobbyChat.vue
```diff
@@ -0,0 +1,125 @@
+<!--
```
burnhamrobertp, line 1: Identical to PartyChat.vue apart from the party/lobby names, unused `newMessage` ref included. Reads like one ChatPanel taking the messages and a target.

src/renderer/components/battle/HostBattle.vue
```diff
@@ -5,221 +5,623 @@ SPDX-License-Identifier: MIT
...
-                        <div class="flex-row gap-lg flex-center-items">
-                            <div class="flex-row flex-center-items gap-sm">
-                                <Icon :icon="personIcon" />{{ map?.playerCountMin }} - {{ map?.playerCountMax }}
-                            </div>
-                            <div class="flex-row flex-center-items gap-sm">
-                                <Icon :icon="gridIcon" />{{ map?.mapWidth }} x {{ map?.mapHeight }}
+                <div class="host-layout">
+                    <div class="host-left">
+                        <div class="options">
+                            <MapBattlePreview :map="draftMap" :map-options="draftMapOptions">
```
burnhamrobertp, line 26: Same slot and updateBox as EditableMapBattlePreview.vue, which this PR also adds. MapOptionsModal uses it, this doesn't.

src/renderer/components/lobbies/LobbyBotParticipant.vue
```diff
@@ -0,0 +1,184 @@
...
+        command: openConfigureModal,
+    },
+    {
+        label: t("lobby.components.battle.botParticipant.kick"),
+        command: kickBot,
+    },
+];
+
+// Fill it with options as known to the server.
+const convertedOptions = computed(() => {
```
burnhamrobertp, line 82: configureBot() writes botOptions.value from inside a computed, and botOptions is bound to :sections on the same modal.

src/renderer/components/battle/SpectatorsComponent.vue
Hectate: I dropped the country flag and changed it from a component to just a username because in offline views it's literally just the local player's name, if any.

src/renderer/components/battle/HostBattle.vue
Hectate: The result of deleting an AllyTeam with players in it is server-defined behavior, so we can't block on that in the client.

src/renderer/views/play/lobby.vue
Hectate: Yes, currently the FFA is essentially a stub and not for use at this time. I did the lobby details dump (from the existing lobby view) and left the Switch Template button in so it would be easy for us to inspect the active lobby properties.
Once we remove the dev-mode-only limitation from Custom Lobbies we will clean up the options at the same time.

src/renderer/components/battle/HostBattle.vue
Hectate: Either I fixed this with other changes or I misunderstand you. To be sure; if you change the AllyTeam count, you do get new startboxes. They are automatically copied from the last startbox - and as a result they overlap exactly. Users would still be expected to move them to their final position.
I'll add some small jitter to the position so they're not perfectly overlapping. Size will remain the same with the expectation that players will want equal-area startboxes for fairness.

src/renderer/components/battle/SpectatorsComponent.vue
Hectate: Fun catch!

src/renderer/components/battle/HostBattle.vue
burnhamrobertp: The first issue still applies though? Also I don't recall, does the client code handle the server rejecting an attempt at team deletion?

Hectate: I only replied to that part. Still have to dig into the other issue this evening.

For the client, the request and the result are independent Tachyon messages. The client will make the request, but the server’s success/failure response will have no bearing on the lobby object. That will be dictated by the arrival of a lobby/updated event. This approach maintains server authority over the result.

burnhamrobertp: I was meaning in terms of UX though. i.e. showing a loading state until the server approves / makes the deletion, or returns an error indicating that failure occurred? This may require that we not have the "request" and its corresponding response completely decoupled from one another (which I thought was already a capability we had?

If its not....that's somewhat problematic because not being able to tie request -> response means that you'll never be able to close the loop for UX purposes between a player's actions in the client and the success (or failure) of that action.

src/renderer/components/battle/Playerlist.vue
Hectate: Modal will get closed upon sending the request, so the notification will be visible, but I'll also add a specific error for that response.

src/renderer/components/battle/HostBattle.vue
Hectate: Just to be clear, the request/response are connected. I used "result" to specifically mean "the change to the lobby that arrives in a separate `lobby/updated` event." Notably, it's actually possible that the `lobby/updated` event arrives _before_ the success response for the initiating `lobby/update` request.

It could be possible to show a "pending request" until we get success/failure, but I don't see it as problematic the way they are now.
* In a Boss lobby, it'll just generally happen because a lobby boss said to do it. (this PR does not address blocking non-boss clients from sending updates, but it's an easy fix later).
* In a Non-boss lobby, it'll go to a vote, and the players will cast those and wait for the result. Seeing the vote end means they either expect or do not expect a change anyway.

src/renderer/components/lobbies/LobbyChat.vue
Hectate: I don't disagree, and was thinking about that when I copied/converted it over. My only remaining reason was that I thought there was a chance we would be asked to insert system/server messages into the lobby chat in the future, so we may want them to be unique in some fashion. Thinking about it now, there's ways to do that without this repetition.

Edit: and I just remembered; we used to have this disparity between User chat history (stored by ID) and Lobby/Party (originally _not_ stored by ID). That's where PartyChat came from and I carried that forward to LobbyChat. But yeah, that's irrelevant now because they are the same structure.

PR #721, opened 2026-09-17
Build out the report user flow

burnhamrobertp:
Builds out the report user flow to match KayZee's mockup: two reason cards instead of the sub type grid, a step for picking the chat lines being complained about, image attachment, and a success step.

Since you asked for the rough edges, none of this can actually be submitted yet. Teiserver has no handler for user/report, so the command comes back unimplemented and the success step is unreachable until that lands. The chat and image selections are deliberately not sent either: the report request only carries user ids, a reason and a string, and there is no durable id for a chat message to point at even if it could (beyond-all-reason/tachyon#155, beyond-all-reason/teiserver#1520). Picking a message is stubbed on object identity in the client, so swapping it for a real id later is a small change. Sub types went away because the mockup drops the distinction and nothing reads them today.

Chat only shows what the session happens to be holding, which is usually nothing when the report is opened from a profile or a replay, since the server replays around 30 seconds of history to a reconnecting client. Reporting from a replay also needs the player to have been seen online before, because replay files carry names but no user ids.

Right clicking a name goes on the shared chat panel from #719 rather than on either caller, so it covers lobby and party chat together.

Images are picked through a dialog owned by the window instead of an input element, which stops Windows surfacing the taskbar over a borderless window. The Browse button in settings still does that.


Hectate:
The "Which Message" screen should have collapsible sections for each (Lobby/Party/DM). Also if you have reported from a specific message (rather than just Report User in general, like from Friends), ideally the window would auto-collapse the other entries _and_ scroll that particular entry into view, if needed. For consistency, the chevron icons (like the team/spectator components) are preferred.

image

Since names are not carried on DM chat, there is no way to report directly from those DMs. The name does not have an option to report, and if the offending user is not a Friend or pending Friend, there is no direct path to report them from the friend's view _or_ the DM history. A user could spam random IDs, or unfriend someone after a message, and it would be difficult to report them by the recipient.

This is probably the moment where we decide to merge the DM chat into ChatPanel too, with a setting to have the messages in DM-style or Group-style. At the same time, the _entire_ message (not just the username) could be the context menu triggerable space, which solves the lack of reportability from DMs.

src/main/services/paths.service.ts
```diff
@@ -95,6 +95,21 @@ function registerIpcHandlers(webContents: BarIpcWebContents) {
         return canceled ? null : filePaths[0];
     });
 
+    // Parented to the window so the picker stays owned by the app, rather than the taskbar
+    // surfacing over a borderless window.
```
Hectate: The taskbar became visible to me, which seems to contradict this comment?

Hectate (changes requested):
One small comment in addition to the UX tweaks.

src/main/services/paths.service.ts
burnhamrobertp: The comment is only partially correct. I couldn't find a way to prevent the taskbar from showing without implementing a custom in app file picker. That's a separate thing I had meant to bring up with you, if you expected file picking to be a recurring UX. We might as well build it.

Either way, I can update the comment

Hectate: Leaving resolved because I'm ambivalent to the taskbar showing or not. My preference is to _not_ incur the work and tech debt of implementing a file picker when the OS and Electron have a multi-platform solution already in place. I recall fighting with the in-engine file picker Godot Engine provides, and I much prefer permitting the OS to handle it.

